PR #81, opened 2021-02-25
Pre release fix flash attacks

Joeysantoro:
It is possible to execute a flash loan and profit via the following trade:

1. Flash borrow ETH
2. Buy on spot market
3. Buy on bonding curve and execute bonding curve allocation
4. Sell FEI from 2 and 3
5. Profit

The end result is the spot price of FEI/ETH is lower than at the beginning, essentially arbing by spiking the price of FEI since it uses the spot price when depositing. The solution has 2 parts:
1. deposit at oracle price not spot price
2. Restrict allocate() and other potentially risky flows to EOA only

Spreadsheet detailing attack https://docs.google.com/spreadsheets/d/1PU51ptA8_xxCEJjWqMbobmNZqX5VBToAAwHm3mmDpk8/edit#gid=0

### DIFF
--- a/contracts/bondingcurve/BondingCurve.sol
+++ b/contracts/bondingcurve/BondingCurve.sol
@@ -79,7 +79,7 @@ abstract contract BondingCurve is IBondingCurve, OracleRef, PCVSplitter, Timed {
     }
 
     /// @notice batch allocate held PCV
-    function allocate() external override postGenesis whenNotPaused {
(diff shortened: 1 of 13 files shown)


ytrezq:
@Joeysantoro which is the time period used by the Orcale as a result? How does the fix, prevent a big whale using his own funds for doing this?

PR #95, opened 2021-04-21
FIP-2

Joeysantoro:
EDIT: added the TribeDripper

Adds three contracts:
* EthReserveStabilizer which allows for exchanging FEI for ETH at a usd conversion rate
* EthPCVDripper which can send in the ETH to the reserve stabilizer in batches over time
* TribeDripper sends TRIBE to the FeiRewardsDistributor to smoothen the front-loading of rewards

The FeiRewardsDistributor calculates the amount to release every week by taking the total balance of TRIBE and subtracting unreleased future rewards.
When a new TRIBE airdrop happens, it frontloads all the rewards for the weeks that have already passed into the first week.
For FIP-2 this would lead to 6x rewards in week 1 (1x base rewards + 100% boost x 5 drips) instead of 2x.

The TribeDripper drips in 47m, 31m, and 22m TRIBE each over the first 3 weeks to smoothen the distribution.

Week 1: 1x base rewards + 47% boost x 5 drips = ~3.35x
Week 2: 1.47x base rewards + 31% x 6 drips = ~3.33x
Week 3: 1.78x base rewards + 22% x 7 drips = ~3.32x
Week 4+: 2x base rewards

These would be voted on per FIP-2: 
https://tribe.fei.money/t/fip-2-fei-redemption-tribe-staking/3033

### DIFF
--- a/contracts/pcv/EthPCVDripper.sol
+++ b/contracts/pcv/EthPCVDripper.sol
@@ -52,11 +52,17 @@ contract EthPCVDripper is CoreRef, Timed {
        afterTime
        whenNotPaused
    {
+       require(isTargetBalanceLow(), "EthPCVDripper: target balance too high");
+
        // reset timer
        _initTimed();
 
        // drip
        payable(target).sendValue(amountToDrip);
        emit Dripped(amountToDrip);
+   }
+
+   function isTargetBalanceLow() public view returns(bool) {
+       return target.balance < amountToDrip;
    }
 }
--- a/contracts/pcv/EthPCVDripper.sol
+++ b/contracts/pcv/EthPCVDripper.sol
@@ -16,7 +16,7 @@ contract EthPCVDripper is CoreRef, Timed {
    uint256 public amountToDrip;
 
    event Dripped(uint256 amount);
-   event Withdrawl(address indexed to, uint256 amount);
+   event Withdrawal(address indexed to, uint256 amount);
 
    /// @notice Uniswap PCV Deposit constructor
    /// @param _core Fei Core for reference
@@ -43,7 +43,7 @@ contract EthPCVDripper is CoreRef, Timed {
        onlyPCVController
    {
        payable(to).sendValue(amount);
-       emit Withdrawl(to, amount);
+       emit Withdrawal(to, amount);
    }
  
    /// @notice drip ETH to target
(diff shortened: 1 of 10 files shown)


GNSPS (approved):

contracts/pcv/EthPCVDripper.sol
```diff
@@ -0,0 +1,62 @@
...
+   using Address for address payable;
+ 
+   /// @notice target address to drip to
+   address public target;
+
+   /// @notice amount to drip after each window
+   uint256 public amountToDrip;
+
+   event Dripped(uint256 amount);
+   event Withdrawl(address indexed to, uint256 amount);
```
eswak, line 19: small typo here in the event name

contracts/pcv/IReserveStabilizer.sol
```diff
@@ -0,0 +1,35 @@
...
+
+    event Withdrawal(
+        address indexed _caller,
+        address indexed _to,
+        uint256 _amount
+    );
+
+    // ----------- State changing api -----------
+
+    function exchangeFeiForEth(uint256 feiAmount) external returns (uint256);
```
eswak, line 20: in the interface you probably want more generic names (a `DaiReserverStabilizer` with an `exchangeFeiForEth` function would be weird)

contracts/pcv/EthPCVDripper.sol
```diff
@@ -0,0 +1,62 @@
...
+   function drip()
+       external
+       afterTime
+       whenNotPaused
+   {
+       // reset timer
+       _initTimed();
+
+       // drip
+       payable(target).sendValue(amountToDrip);
```
Corsaire, line 59: Would be good to only make the transfer if the target’s balance is lower than some threshold.

contracts/pcv/EthPCVDripper.sol
Joeysantoro: fixed

contracts/pcv/IReserveStabilizer.sol
Joeysantoro: renamed it to exchangeFei

contracts/pcv/EthPCVDripper.sol
ruihop: nic

PR #104, opened 2021-07-10
Cleanup Events

Joeysantoro:
Fix discrepancies in events to fix OZ issue M01


Joeysantoro:
Being reviewed by OZ in another repo

PR #108, opened 2021-07-10
Move WithdrawERC20 to helper PCVDeposit class

Joeysantoro:
Moves the withdrawERC20 method to a helper child class called PCVDeposit.sol.

Also resolves OZ audit issue L06 by removing declaration namespace collision in ReserveStabilizer.

There was an additional collision with `balance` in the PCVSwapperUniswap this PR also corrects


eswak:
👍

PR #119, opened 2021-08-03
Add Mint Cap to Bonding Curve

Joeysantoro:
Adds a cap on the amount of FEI minted by a bonding curve to mitigate the damage from infinite minting attacks and oracle malfunctions

### DIFF
--- a/contracts/bondingcurve/BondingCurve.sol
+++ b/contracts/bondingcurve/BondingCurve.sol
@@ -32,6 +32,9 @@ contract BondingCurve is IBondingCurve, OracleRef, PCVSplitter, Timed, Incentivi
     /// @notice the discount applied on top of peg before at Scale
     uint256 public override discount;
 
+    /// @notice the cap on how much FEI can be minted by the bonding curve
+    uint256 public override mintCap;
+
     uint256 public constant BASIS_POINTS_GRANULARITY = 10_000;
 
     /// @notice constructor
@@ -64,6 +67,7 @@ contract BondingCurve is IBondingCurve, OracleRef, PCVSplitter, Timed, Incentivi
         Timed(_duration)
         Incentivized(_incentive)
     {
+        _setMintCap(_scale); // default mint cap is scale
         _setScale(_scale);
         token = _token;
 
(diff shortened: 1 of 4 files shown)


test/bondingcurve/BondingCurve.test.js
```diff
@@ -270,7 +272,16 @@ describe('BondingCurve', function () {
           });
         });
       });
-        
+      describe('Exceeding cap', function() {
```
sebastiandelgado, line 275: Can't tell if there is a test in this file that buys under the mint cap to check that its still possible to buy even with the mint cap on

sebastiandelgado (approved):

contracts/bondingcurve/BondingCurve.sol
```diff
@@ -32,6 +32,9 @@ contract BondingCurve is IBondingCurve, OracleRef, PCVSplitter, Timed, Incentivi
     /// @notice the discount applied on top of peg before at Scale
     uint256 public override discount;
 
+    /// @notice the cap on how much FEI can be minted by the bonding curve
+    uint256 public override mintCap;
```
sol-sr, line 36: When we upgrade the contracts, are we doing it with a proxy/storage pattern or some other method?
If we are using a proxy/storage pattern, the variable should be added after BASIS_POINTS_GRANULARITY.

test/bondingcurve/BondingCurve.test.js
sol-sr: To add to this point, I only see purchasing happening in the before each hook and we don't verify the fei they received. Recommend adding a test purchasing from the bonding curve that actually succeeds. After the success, we should verify FEI balances and that the bonding curve or receiving contract's eth balance incremented.

sol-sr (changes requested):

ubaistp (approved):

contracts/bondingcurve/BondingCurve.sol
Joeysantoro: We don't use proxy/storage its just drop-and-replace

test/bondingcurve/BondingCurve.test.js
Joeysantoro: There are tests for the FEI being received higher up in the file

eswak (approved):

eswak:
keep this PR simple, but maybe an idea for later : you could add these `mintCap`, `setMintCap`, `availableToMint` functions, and `MintCapUpdate` event, in a generic extension "PCVMinterBase" (or something like it), that contracts could inherit.

The minted/burnt amount can be tracked by an override of `_burnFeiHeld` and `_mintFei` from `CoreRef` in this base contract.

When I was working on the Curve deposit with reweights, I wanted to have this kind of tracking of minted / burnt fei, and implement a kind of `availableToMint` function too. Maybe that's a need that is pretty generic and applies to multiple PCV contracts that have Minter/Burner role ?

thomas-waite (commented):
Looks good to me!

PR #126, opened 2021-08-12
SnapshotDelegator

Joeysantoro:
To vote on snapshot, you need a private key, therefore contracts are excluded.

In order to use the INDEX from the potential OTC deal, we'd need to use the snapshot delegate registry to appoint a proxy EOA. https://docs.snapshot.org/guides/delegation

This PR introduces a snapshot delegator PCV deposit with the following features:
- delegation
- Optimistic approval admin support for re-delegation
- Guardian support for undelegation

It also updates the indexOTC DAO proposal with a transfer of INDEX from the DAO to the snapshotDelegator

### DIFF
--- a/contracts/pcv/snapshot/SnapshotDelegatorPCVDeposit.sol
+++ b/contracts/pcv/snapshot/SnapshotDelegatorPCVDeposit.sol
@@ -51,7 +51,6 @@ contract SnapshotDelegatorPCVDeposit is PCVDeposit {
         external
         override
         onlyPCVController
-        whenNotPaused
     {
         _withdrawERC20(address(token), to, amountUnderlying);
     }
--- a/contracts/pcv/snapshot/SnapshotDelegatorPCVDeposit.sol
+++ b/contracts/pcv/snapshot/SnapshotDelegatorPCVDeposit.sol
@@ -1,0 +1,89 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.0;
+
+import "../PCVDeposit.sol";
+
+interface DelegateRegistry {
+    function setDelegate(bytes32 id, address delegate) external;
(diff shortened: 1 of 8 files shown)


contracts/pcv/snapshot/SnapshotDelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,89 @@
...
+    ) CoreRef(_core) {
+        token = _token;
+        spaceId = _spaceId;
+        _delegate(_initialDelegate);
+    }
+
+    /// @notice withdraw tokens from the PCV allocation
+    /// @param amountUnderlying of tokens withdrawn
+    /// @param to the address to send PCV to
+    function withdraw(address to, uint256 amountUnderlying)
```
GNSPS, line 50: I think it's awkward to have a function that behaves exactly the same way as `PCVDeposit::withdrawERC20` but with just more relaxed pre-conditions.

GNSPS (commented):
I also don't get why this is important for the threat model:

`To vote on snapshot, you need a private key, therefore contracts are excluded.`

And I kind of disagree with the "optimistic" nomenclature here:

`Optimistic approval admin support for re-delegation`

Because in my mind "optimistic" approval kind of requires an _automatic_ fraud-proof-like system that can be initiated by any actor, as opposed to having to be revoked through a DAO vote (i.e., a subjective, more prolonged interaction).

Overall the contract LGTM. These are mostly nitpicks.

GNSPS (approved):

contracts/pcv/snapshot/SnapshotDelegatorPCVDeposit.sol
Joeysantoro: Removed the pause condition

PR #127, opened 2021-08-12
Weth PCV Deposit

Joeysantoro:
Refactors some WETH-related code to allow for more modular PCV components where WETH is needed

This will be useful for an Aave ETH PCV deposit because Aave uses WETH


eswak (approved):
You have weird space/tab mixup & non-consistent indentation (sometimes 4, sometimes 2 spaces) in `WethPCVDeposit.sol`.

But nothing critical. Good idea to factor these functions in a base contract 🙂

PR #137, opened 2021-08-20
FIP-10b - Curve deposit

eswak:
This proposal adds a new `StableSwapOperatorV1` contract to the protocol.

It is a `PCVDeposit` used to manage liquidity in the Curve metapool [3crv, FEI]. No reweights, no FEI minting, and all state-changing functions are access-controlled. The idea is to reduce the attack surface in this first version, and keep it as simple as possible.

The assumption is that 50M DAI have been raised through the bonding curve deployed in [FIP-10a](https://www.withtally.com/governance/fei/proposal/13), and that the protocol owns 50M DAI deposited in Compound.

Short term actions :
- Discuss with the community whether all of the 50M DAI should be moved to the metapool
- Possibly include the opening of a new DAI bonding curve in this same proposal, to get more DAI ?
- In-depth review & audit of the `StableSwapOperatorV1` @Joeysantoro


contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+    // ------------------ Private properties -----------------------------------
+
+    /// some fixed variables to interact with the pool
+    uint8 private _feiIndex; // index of FEI in the pool (0 or 1)
+    uint8 private _3crvIndex; // index of 3crv in the pool (0 or 1)
+    address private _3crv; // address of the 3crv token
+    address private _3pool; // address of the 3pool
+    address private _dai; // address of the DAI token
+    address private _usdc; // address of the USDC token
+    address private _usdt; // address of the USDT token
```
Joeysantoro, line 50: can make immutable to save gas. Immutable variables are compiled inline to save a SLOAD call

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+        uint256 _usdtBalance = IERC20(_usdt).balanceOf(address(this));
+        uint256 _3crvBalanceBefore = IERC20(_3crv).balanceOf(address(this));
+        if (_daiBalance != 0 || _usdcBalance != 0 || _usdtBalance != 0) {
+            uint256[3] memory _add3poolLiquidityAmounts;
+            _add3poolLiquidityAmounts[0] = _daiBalance;
+            _add3poolLiquidityAmounts[1] = _usdcBalance;
+            _add3poolLiquidityAmounts[2] = _usdtBalance;
+            IERC20(_dai).approve(_3pool, _daiBalance);
+            IERC20(_usdc).approve(_3pool, _usdcBalance);
+            SafeERC20.safeApprove(IERC20(_usdt), _3pool, _usdtBalance);
```
Joeysantoro, line 98: Why not use safeApprove for all 3?

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+            I3Pool(_3pool).add_liquidity(
+              _add3poolLiquidityAmounts,
+              0 // minimum 0 3crv lp tokens out (defensively checked below)
```
Joeysantoro, line 101: Why not use this embedded slippage param?

Joeysantoro: We should reduce the logic on our side where possible

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+        _3pool = _curve3pool;
+        _dai = IStableSwap(_3pool).coins(0);
+        _usdc = IStableSwap(_3pool).coins(1);
+        _usdt = IStableSwap(_3pool).coins(2);
+    }
+
+    /// @notice deposit DAI, USDC, USDT, 3crv, and FEI into the pool.
+    /// Note: the FEI has to be minted & deposited on this contract in a previous
+    ///       tx, as this contract does not use the Minter role.
+    function deposit() public override onlyPCVController whenNotPaused {
```
Joeysantoro, line 82: why onlyPCVController? Normally we allow any address to deposit

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+        uint256 _3crvAmountAfter = _3crvAmount + _3crvBalanceAfter;
+
+        // get the usd value of 3crv in the pool
+        uint256 _3crvUsdValue = _3crvAmountAfter * _3crvVirtualPrice / 1e18;
+
+        // compute the number of FEI to deposit
+        uint256 _feiToDeposit = 0;
+        if (_3crvUsdValue > _feiAmount) {
+            _feiToDeposit = _3crvUsdValue - _feiAmount;
+        }
```
Joeysantoro, line 127: why not deposit any FEI in this case? Might make more sense to just deposit proportionally

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+        if (_3crvBalanceAfter > 0 || _feiToDeposit > 0) {
+            // build parameters
+            uint256[2] memory _addLiquidityAmounts;
+            _addLiquidityAmounts[_feiIndex] = _feiToDeposit;
+            _addLiquidityAmounts[_3crvIndex] = _3crvBalanceAfter;
+
+            // approvals
+            IERC20(address(_fei)).approve(pool, _feiToDeposit);
+            // 3crv needs to reset allowance to 0, see CurveTokenV1.vy#L117 :
+            // assert _value == 0 or self.allowances[msg.sender][_spender] == 0
```
Joeysantoro, line 139: weird. Great comment

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+            IERC20(_3crv).approve(pool, 0);
+            IERC20(_3crv).approve(pool, _3crvBalanceAfter);
+
+            // do deposit
+            uint256 _balanceBefore = IERC20(pool).balanceOf(address(this));
+            IStableSwap(pool).add_liquidity(_addLiquidityAmounts, 0);
+            uint256 _balanceAfter = IERC20(pool).balanceOf(address(this));
+
+            // emit event
```
Joeysantoro, line 148: comment is in wrong spot here

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+            uint256 _daiOut = IStableSwap(_3pool).calc_withdraw_one_coin(
+              _3crvOut, // LP tokens just deposited
+              0 // 3pool coin 0 = DAI
+            );
+            emit Deposit(msg.sender, _daiOut);
+        }
+
+        // burn remaining FEI
+        uint256 _feiBalance = _fei.balanceOf(address(this));
+        _fei.burn(_feiBalance);
```
Joeysantoro, line 160: _burnFeiHeld() is a helper method in CoreRef. Not sure we want to do this though because we may want to keep the FEI we have since this contract isn't a minter

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+            // approvals
+            IERC20(address(_fei)).approve(pool, _feiToDeposit);
+            // 3crv needs to reset allowance to 0, see CurveTokenV1.vy#L117 :
+            // assert _value == 0 or self.allowances[msg.sender][_spender] == 0
+            IERC20(_3crv).approve(pool, 0);
+            IERC20(_3crv).approve(pool, _3crvBalanceAfter);
+
+            // do deposit
+            uint256 _balanceBefore = IERC20(pool).balanceOf(address(this));
+            IStableSwap(pool).add_liquidity(_addLiquidityAmounts, 0);
```
Joeysantoro, line 145: No slippage protection here?

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+        uint256 _feiBalance = _fei.balanceOf(address(this));
+        _fei.burn(_feiBalance);
+    }
+
+    /// @notice withdraw DAI from the deposit. This will remove liquidity in the
+    /// current pool's proportion, burn the FEI part of it, and then use the 3crv
+    /// to withdraw liquidity of the 3pool in DAI.
+    /// note: because of slippage & fees on withdraw, the amount of DAI out is
+    ///        not exactly the expected amountUnderlying provided as input.
+    /// @param amountUnderlying of tokens withdrawn (0 = all)
```
Joeysantoro, line 168: I don't like the 0=all logic. If anything it should be withdraw min(balance, amountUnderlying). Either way we have a ratioPCVController that can do the total withdraw so I'd rather remove this logic

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        // remove 3pool liquidity in DAI (3crv -> DAI)
+        uint256 _3crvBalance = IERC20(_3crv).balanceOf(address(this));
+        uint256 _daiBalanceBefore = IERC20(_dai).balanceOf(address(this));
+        IERC20(_3pool).approve(_3pool, _3crvBalance);
+        IStableSwap(_3pool).remove_liquidity_one_coin(
+          _3crvBalance,
+          0, // 0 = DAI token index
+          0 // minimum 0 DAI out (defensively checked below)
```
Joeysantoro, line 203: same comment as above, if we can use the embedded slippage check that is preferrable

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+        uint256 _daiBalanceAfter = IERC20(_dai).balanceOf(address(this));
+        uint256 _minDaiOut = amountUnderlying * (BASIS_POINTS_GRANULARITY - depositMaxSlippageBasisPoints) / BASIS_POINTS_GRANULARITY;
+        uint256 _daiBalanceWithdrawn = _daiBalanceAfter - _daiBalanceBefore;
+        require(_daiBalanceWithdrawn >= _minDaiOut, "StableSwwapOperatorV1: 3pool withdraw slippage too high");
+
+        // compute amount to withdraw
+        uint256 _amount = amountUnderlying;
+        // @dev: there is a small slippage on remove, so the amount of DAI balance
+        //       will be slightly smaller than the amountUnderlying if the contract
+        //       held no DAI before withdraw.
```
Joeysantoro, line 216: Is it possible to approximate/overestimate the withdraw amount to make the output exact?

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -0,0 +1,264 @@
...
+            SafeERC20.safeTransfer(IERC20(_dai), to, _amount);
+        }
+
+        // emit event
+        emit Withdrawal(msg.sender, to, _amount);
+
+        // burn FEI remaining
+        IFei _fei = fei();
+        uint256 _feiBalance = _fei.balanceOf(address(this));
+        _fei.burn(_feiBalance);
```
Joeysantoro, line 232: There is a _burnFeiHeld() helper method in CoreRef. Still unsure if we want to burn when this contract isn't a minter

Joeysantoro (commented):
First pass only on the main solidity

contracts/pcv/curve/StableSwapOperatorV1.sol
eswak: done 👍 

Had to introduce a temporary variable for _3crvIndex because immutable variables cannot be read during contract creation

contracts/pcv/curve/StableSwapOperatorV1.sol
eswak: Only USDT required it because it doesn't have a standard return, but I'll use it for all 3, for uniformity

contracts/pcv/curve/StableSwapOperatorV1.sol
eswak: I don't think it's needed, because we deposit in proportions that makes the pool balanced, unlike when we deposit/withdraw 100% DAI (instead of a basket of DAI/USDC/USDT), where slippage can happen ?

contracts/pcv/curve/StableSwapOperatorV1.sol
eswak: I'll remove the withdraw 0 logic

contracts/pcv/curve/StableSwapOperatorV1.sol
eswak: it used to be able to mint FEI so I wanted stricter access control, but I guess we can remove it now

contracts/pcv/curve/StableSwapOperatorV1.sol
eswak: I'll remove the FEI burn logics in `deposit()` and `withdraw()`. Since the contract can't mint new FEI, it makes sense to keep it now.

eswak: btw, I didn't use _burnFeiHeld() to avoid a second call to fei() to optimize gas

contracts/pcv/curve/StableSwapOperatorV1.sol
eswak: I'll remove the FEI burn logics in `deposit()` and `withdraw()`. Since the contract can't mint new FEI, it makes sense to keep it now.

contracts/pcv/curve/StableSwapOperatorV1.sol
Joeysantoro: Doesn't ever hurt

contracts/pcv/curve/StableSwapOperatorV1.sol
eswak: I added an overshoot of 0.015% to account for fees in `withdraw_liquidity_one_coin()`

contracts/pcv/curve/StableSwapOperatorV1.sol
eswak: Added a slippage check there

contracts/pcv/curve/StableSwapOperatorV1.sol
```diff
@@ -190,6 +190,13 @@ contract StableSwapOperatorV1 is PCVDeposit {
         uint256 _lpTotalSupply = IERC20(pool).totalSupply();
         uint256 _lpToWithdraw = _lpTotalSupply * _amount3crvToWithdraw / IStableSwap(pool).balances(_3crvIndex);
 
+        // overshoot withdraw by the Curve withdraw fee, so that the discrepancy
+        // between amountUnderlying and actual amount out is minimized, when we
+        // will call withdraw_liquidity_one_coin() in the 3pool.
+        // the fee is 0.015% (see SwapTemplateMeta.vy#L916) :
+        // fee: uint256 = self.fee * N_COINS / (4 * (N_COINS - 1))
+        _lpToWithdraw = _lpToWithdraw * 100015 / 100000;
```
Joeysantoro, line 198: What if the _lpToWithdraw is equal to the total before applying this transformation? It should be capped at the total balance of the pcv deposit.

deploy/fip_10b.js
```diff
@@ -0,0 +1,29 @@
...
+  if (
+    !coreAddress
+  ) {
+    throw new Error('An environment variable contract address is not set');
+  }
+
+  const curveMetapoolDeposit = await StableSwapOperatorV1.new(
+    coreAddress,
+    '0x06cb22615ba53e60d67bf6c341a0fd5e718e1655', // FEI-3crv metapool
+    '0xbebc44782c7db0a1a60cb6fe97d0b483032ff1c7', // Curve 3pool
```
Joeysantoro, line 17: These should both be added to mainnetAddresses.json in which case they can be pulled in through the `addresses` parameter in this function

deploy/fip_10b.js
```diff
@@ -0,0 +1,29 @@
...
+    '50', // 0.5% max slippage
```
Joeysantoro, line 18: Is .5% too high? What does Curve UI use

contracts/pcv/curve/StableSwapOperatorV1.sol
eswak: I initially added a check on that, but it has a side effect that I don't like : if the deposit has 1000 DAI, and we try to withdraw 2000, the call does not revert, it just withdraws 1000. What do you think : should we revert, or withdraw the max available ?

deploy/fip_10b.js
eswak: image
🤷‍♂️ 

I guess we can keep 0.5%

PR #140, opened 2021-08-26
Add Proxy Contracts, Wire Them Up in TribalChief Tests

sol-sr:
Adds a TransparentUpgradeableProxy in front of the TribalChief. Updates both TribalChief and CoreRef to initialize-able. Ignores the Pauseable class as the constructor only sets a default false state variable back to false.


contracts/staking/TribalChief.sol
```diff
@@ -94,12 +95,17 @@ contract TribalChief is CoreRef, ReentrancyGuard {
...
 
+
+    /// @notice these variables cannot be initialized with a constant value 
+    /// because of the way that upgradeable contracts work, their initial write
+    // to storage doesn't happen as the logic contract doesn't run the constructor
+
     /// the amount of tribe distributed per block
-    uint256 private tribalChiefTribePerBlock = 1e20;
+    uint256 private tribalChiefTribePerBlock;
     /// variable has been made constant to cut down on gas costs
```
GNSPS, line 105: Comment seems outdated

contracts/staking/TribalChief.sol
```diff
@@ -94,12 +95,17 @@ contract TribalChief is CoreRef, ReentrancyGuard {
...
-    uint256 private constant ACC_TRIBE_PRECISION = 1e23;
```
GNSPS, line 100: Why are we getting rid of constants? They should be hardcoded into the bytecode, and therefore, usable on a proxy pattern.

I admittedly am not an expert on ERC 1967 but since it uses `delegatecall`, all the bytecode positions should be available in a certain execution frame. The constants are built at compile time, not deploy time.

contracts/staking/TribalChief.sol
```diff
@@ -114,11 +120,24 @@ contract TribalChief is CoreRef, ReentrancyGuard {
     event NewTribePerBlock(uint256 indexed amount);
     event PoolLocked(bool indexed locked, uint256 indexed pid);
 
+    /// @notice The way this function is constructed, you will not be able to 
+    /// call initialize after this function is constructed, effectively 
+    /// only allowing TribalChief to be used through delegate calls.
+    /// @param coreAddress The Core contract address.
+    constructor(address coreAddress) CoreRef(coreAddress) {}
```
GNSPS, line 127: I have thought so much about proxy patterns that I believe it'd be great for us to talk more at length about this, @Joeysantoro . 😄 🙌

contracts/staking/TribalChief.sol
Joeysantoro: I'm pretty sure this is the case as well so we can keep constants

contracts/staking/TribalChief.sol
Joeysantoro: Definitely

Joeysantoro:
Merging using https://github.com/fei-protocol/fei-protocol-core/pull/145

PR #146, opened 2021-08-31
Collateralization Oracle (v2)

eswak:
This PR adds the Collateralization Oracle required for the V2.

In short, this new oracle...

- Is configured by the DAO to follow a list of PCVDeposits (add, remove, swap are `onlyGovernor`)
- Has its own list of oracles (need to call `setOracle(token, address)`)
- Relies on an updated PCVDeposit interface, we will need to deploy wrappers for existing deposits

The new PCVDeposit interface has the following methods : 

- `balanceReportedIn()` the token address used for accounting in this deposit
- `balanceAndFei()` the resistant token balance & the protocol-owned FEI in this deposit

The CR Oracle has a function `pcvStats()` that returns :

- the Protocol Controlled Value 
- the User Circulating FEI
- the Protocol Equity
- a validity status

### DIFF
--- a/contracts/oracle/CollateralizationOracle.sol
+++ b/contracts/oracle/CollateralizationOracle.sol
@@ -308,8 +308,9 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
     ///         Controlled Value) than the circulating (user-owned) FEI, i.e.
     ///         a positive Protocol Equity.
     ///         Note: the validity status is ignored in this function.
-    function isOvercollateralized() external view override whenNotPaused returns (bool) {
-        (,, uint256 _protocolEquity,) = pcvStats();
+    function isOvercollateralized() external override view whenNotPaused returns (bool) {
+        (,, uint256 _protocolEquity, bool _valid) = pcvStats();
+        require(_valid, "CollateralizationOracle: reading is invalid");
         return _protocolEquity > 0;
     }
 }
--- a/contracts/oracle/CollateralizationOracle.sol
+++ b/contracts/oracle/CollateralizationOracle.sol
@@ -182,18 +182,22 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
             (Decimal.D256 memory _price, bool _valid) = IOracle(tokenToOracle[_token]).read();
             _globalValid = _globalValid && _valid;
 
+            uint256 _nTokens = 0;
             // For each deposit...
             for (uint256 j = 0; j < tokenToDeposits[_token].length; j++) {
                 IPCVDepositV2 _deposit = IPCVDepositV2(tokenToDeposits[_token][j]);
 
-                // Increment the total protocol controlled value by the USD value of
-                // the asset held in this PCVDeposit
-                _protocolControlledValue += _price.mul(_deposit.resistantBalance()).asUint256();
+                // Increment the number of tokens held by the protocol
+                _nTokens += _deposit.resistantBalance();
 
                 // Increment the total protocol controlled FEI by the amount of FEI
                 // held by this PCVDeposit
                 _protocolControlledFei += _deposit.resistantProtocolOwnedFei();
             }
+
+            // Increment the total PCV by the number of tokens held times the USD
+            // value of this token.
+            _protocolControlledValue += _price.mul(_nTokens).asUint256();
         }
 
         // The circulating FEI is defined as the total FEI supply, minus the
@@ -245,10 +249,12 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
             require(_valid, "CollateralizationOracle: oracle invalid");
 
             // For each deposit...
+            uint256 _nTokens = 0;
             for (uint256 j = 0; j < tokenToDeposits[_token].length; j++) {
                 IPCVDepositV2 _deposit = IPCVDepositV2(tokenToDeposits[_token][j]);
-                _pcv += _price.mul(_deposit.resistantBalance()).asUint256();
+                _nTokens += _deposit.resistantBalance();
             }
+            _pcv += _price.mul(_nTokens).asUint256();
         }
 
         return _pcv;
@@ -268,11 +274,13 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
             require(_valid, "CollateralizationOracle: oracle invalid");
 
             // For each deposit...
+            uint256 _nTokens = 0;
             for (uint256 j = 0; j < tokenToDeposits[_token].length; j++) {
                 IPCVDepositV2 _deposit = IPCVDepositV2(tokenToDeposits[_token][j]);
-                _pcv += _price.mul(_deposit.resistantBalance()).asUint256();
+                _nTokens += _deposit.resistantBalance();
                 _protocolControlledFei += _deposit.resistantProtocolOwnedFei();
             }
+            _pcv += _price.mul(_nTokens).asUint256();
         }
 
         uint256 _userCirculatingFei = fei().totalSupply() - _protocolControlledFei;
--- a/contracts/oracle/CollateralizationOracle.sol
+++ b/contracts/oracle/CollateralizationOracle.sol
@@ -108,7 +108,7 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
     ///         note : this function reverts if the deposit is already in the list.
     ///         note : this function reverts if the deposit's token has no oracle.
     /// @param _deposit : the PCVDeposit to add to the list.
-    function addDeposit(address _deposit) external onlyGovernor {
+    function addDeposit(address _deposit) public onlyGovernor {
         // if the PCVDeposit is already listed, revert.
         require(depositToToken[_deposit] == address(0), "CollateralizationOracle: deposit duplicate");
 
@@ -131,7 +131,7 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
     ///         the collateralization ratio oracle.
     ///         note : this function reverts if the input deposit is not found.
     /// @param _deposit : the PCVDeposit address to remove from the list.
-    function removeDeposit(address _deposit) external onlyGovernor {
+    function removeDeposit(address _deposit) public onlyGovernor {
         // get the token in which the deposit reports its token
         address _token = depositToToken[_deposit];
 
@@ -158,28 +158,13 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
     /// @param _oldDeposit : the PCVDeposit to remove from the list.
     /// @param _newDeposit : the PCVDeposit to add to the list.
     function swapDeposit(address _oldDeposit, address _newDeposit) external onlyGovernor {
-        // get the token in which the old deposit reports its token
-        address _token = depositToToken[_oldDeposit];
-        address _newToken = IPCVDepositV2(_newDeposit).balanceReportedIn();
-
-        // revert if old deposit is not found
-        require(_token != address(0), "CollateralizationOracle: deposit not found");
-
-        // revert if new deposit is found
-        require(depositToToken[_newDeposit] == address(0), "CollateralizationOracle: deposit duplicate");
+        address _oldToken = depositToToken[_oldDeposit];
+        removeDeposit(_oldDeposit);
+        addDeposit(_newDeposit);
+        address _newToken = depositToToken[_newDeposit];
 
         // revert if token is different
-        require(_token == _newToken, "CollateralizationOracle: deposit has different token");
-
-        // update maps & arrays for faster access
-        depositToToken[_oldDeposit] = address(0);
-        depositToToken[_newDeposit] = _token;
-        tokenToDeposits[_token].remove(_oldDeposit);
-        tokenToDeposits[_token].add(_newDeposit);
-
-        // emit event
-        emit DepositRemove(msg.sender, _oldDeposit);
-        emit DepositAdd(msg.sender, _newDeposit, _token);
+        require(_oldToken == _newToken, "CollateralizationOracle: deposit has different token");
     }
 
     /// @notice Set the price feed oracle (in USD) for a given asset.
--- a/contracts/oracle/CollateralizationOracle.sol
+++ b/contracts/oracle/CollateralizationOracle.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.4;
 
 import "./IOracle.sol";
+import "./ICollateralizationOracle.sol";
 import "../refs/CoreRef.sol";
 import "../pcv/IPCVDepositV2.sol";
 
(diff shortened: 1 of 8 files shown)


contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,287 @@
...
+        token2oracle[_token] = _oracle;
+
+        // emit event
+        emit OracleSet(msg.sender, _token, _oracle);
+    }
+
+    // ----------- IOracle override methods -----------
+    /// @notice update all oracles required for this oracle to work
+    function update() external override whenNotPaused {
+        for (uint256 i = 0; i < tokensInPcv.length; i++) {
```
Joeysantoro, line 153: We're going to have to be extremely gas conscious with this contract. I don't think updating each individual oracle is necessary here, especially because Chainlink is the primary oracle at the moment

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,287 @@
...
+            IOracle(token2oracle[tokensInPcv[i]]).update();
+        }
+    }
+
+    // @notice returns true if any of the oracles required for this oracle to
+    //         work are outdated.
+    function isOutdated() external override view returns (bool) {
```
Joeysantoro, line 160: I think we are going to want to memoize the CR read value, pcv equity etc. Lets not for now, but do some gas analysis once all of the PCV deposits are wired up. For now just have isOutdated return false always. We should use a similar update mechanism to Chainlink i.e. time and threshold based

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,287 @@
...
+        // The total USD value of assets held in the PCV
+        uint256 _protocolControlledValue = 0;
+        // false if any of the token oracles are invalid, or if paused
+        bool _globalValid = !paused();
+
+        // For each token...
+        for (uint256 i = 0; i < tokensInPcv.length; i++) {
+            address _token = tokensInPcv[i];
+            (Decimal.D256 memory _price, bool _valid) = IOracle(token2oracle[_token]).read();
+            _globalValid = _globalValid && _valid;
```
Joeysantoro, line 185: We should think more about how the validity of the CR oracle interplays with the rest of the system. This is a critical contract involved in both the inflation and buyback of TRIBE. If these other contracts revert on a single oracle being outdated we could end up in a bad situation

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,287 @@
...
+
+            // For each deposit...
+            for (uint256 j = 0; j < token2deposits[_token].length; j++) {
+                IPCVDepositV2 _deposit = IPCVDepositV2(token2deposits[_token][j]);
+
+                // Increment the total protocol controlled value by the USD value of
+                // the asset held in this PCVDeposit
+                _protocolControlledValue += _price.mul(_deposit.resistantBalance()).asUint256();
```
Joeysantoro, line 193: Would storing an intermediate value and multiplying by the price once at the end save gas?

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,287 @@
...
+
+                // Increment the total protocol controlled FEI by the amount of FEI
+                // held by this PCVDeposit
+                _protocolControlledFei += _deposit.resistantProtocolOwnedFei();
```
Joeysantoro, line 197: We should consider exposing an api that returns both resistant balance and resistant FEI so we can reduce the number of external calls like how Chainlink does it

Joeysantoro, line 197: Should FEI deposits be treated differently? Some PCV deposits hold only FEI for example and "resistantBalance" would return the number of FEI

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,287 @@
...
+    ///         is defined as the protocol having more assets in its PCV (Protocol
+    ///         Controlled Value) than the circulating (user-owned) FEI.
+    function isOvercollateralized() external view override whenNotPaused returns (bool) {
+        return pcvEquityValue() > 0;
+    }
+
+    /// @notice returns the amount of circulating (user-owned) FEI. User Circulating
+    ///         FEI is defined as the total supply of FEI minus the FEI held in the
+    ///         various protocol's contracts.
+    function userCirculatingFei() external view override whenNotPaused returns (uint256) {
```
Joeysantoro, line 225: I think it would be best to combine all of these functions into a single one that returns userCirculatingFei, pcvValue and pcvEquity to save gas and simplify the contract. The `read` function can just call that one and do the final logic step

contracts/pcv/IPCVDepositV2.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "./IPCVDeposit.sol";
+
+/// @title a PCV Deposit interface for Fei V2
+/// @author eswak
+interface IPCVDepositV2 is IPCVDeposit {
```
Joeysantoro, line 8: Ultimately we should just add these methods to the original interface (but I know that would lead to many compilation errors so we should wait until ready to add all of these methods to existing contracts)

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,287 @@
...
+    ///         version of the tokensInPcv array.
+    mapping(address => uint8) public isTokenInPcv;
+
+    // ----------- Constructor -----------
+
+    /// @notice CollateralizationOracle constructor
+    /// @param _core Fei Core for reference
+    constructor(
+        address _core
+    ) CoreRef(_core) {}
```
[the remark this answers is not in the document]
Joeysantoro: After discussing offline, we will likely do a 2 contract architecture where memoization happens in the immutable outer layer. We maintain upgradeability without employing proxy pattern

contracts/oracle/CollateralizationOracle.sol
Joeysantoro: Maybe having plausibility remove a PCV deposit from the CR oracle can mitigate this issue

contracts/oracle/CollateralizationOracle.sol
Joeysantoro: Lets use a $0 oracle for FEI

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,287 @@
...
+    /// @notice Map from deposit address to token address. It is used like an
+    ///         indexed version of the pcvDeposits array, to check existence
+    ///         of a PCVdeposit in the current config.
+    mapping(address => address) public deposit2token;
+    /// @notice Array of all tokens held in the PCV. Used for iteration on tokens
+    ///         and oracles.
+    address[] public tokensInPcv;
+    /// @notice Map to know if a given token is in the PCV. Used like an indexed
+    ///         version of the tokensInPcv array.
+    mapping(address => uint8) public isTokenInPcv;
```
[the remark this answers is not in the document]
eswak: "bad habit" from embedded software dev ^^ internally, in the EVM, bool is probably an uint8 anyway. But I will update for readability 👍

contracts/oracle/CollateralizationOracle.sol
eswak: I updated the logic to not revert & return a validity status instead

eswak:
I rebased the branch & updated target to `feat/v2`, and implemented the first batch of changes. I'll work on memoization next.

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -75,8 +75,8 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
         // update maps & arrays for faster access
         depositToToken[_deposit] = _token;
         tokenToDeposits[_token].push(_deposit);
-        if (isTokenInPcv[_token] == 0) {
-          isTokenInPcv[_token] = 1;
+        if (isTokenInPcv[_token] == false) {
```
Joeysantoro, line 82: nit: can just do `!isTokenInPcv[_token]`

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -182,18 +182,22 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
             (Decimal.D256 memory _price, bool _valid) = IOracle(tokenToOracle[_token]).read();
             _globalValid = _globalValid && _valid;
 
+            uint256 _nTokens = 0;
```
Joeysantoro: nTokens isn't a super descriptive name. How about `totalTokenBalance` or something

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -245,10 +249,12 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
             require(_valid, "CollateralizationOracle: oracle invalid");
 
             // For each deposit...
+            uint256 _nTokens = 0;
```
Joeysantoro: Same here with nTokens

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -268,11 +274,13 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
...
```
Joeysantoro, line 277: same here with nTokens

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -137,6 +137,49 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
...
     }
 
+    /// @notice Swap a PCVDeposit with a new one, for instance when a new version
+    ///         of a deposit (holding the same token) is deployed.
+    /// @param _oldDeposit : the PCVDeposit to remove from the list.
+    /// @param _newDeposit : the PCVDeposit to add to the list.
+    function swapDeposit(address _oldDeposit, address _newDeposit) external onlyGovernor {
+        // get the token in which the old deposit reports its token
+        address _token = depositToToken[_oldDeposit];
+        address _newToken = IPCVDepositV2(_newDeposit).balanceReportedIn();
```
Joeysantoro, line 147: I think we should require they have the same token

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -137,6 +137,49 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
...
+
+        // revert if new deposit is found
+        require(depositToToken[_newDeposit] == address(0), "CollateralizationOracle: deposit duplicate");
+
+        // revert if token is different
+        require(_token == _newToken, "CollateralizationOracle: deposit has different token");
+
+        // swap the PCVDeposit in the list
+        bool found = false;
+        for (uint256 i = 0; !found; i++) {
```
Joeysantoro, line 160: I'm starting to think keeping a mapping of indices is a better data structure. Compound does this for proposals in the GovernorAlpha

contracts/oracle/ICollateralizationOracle.sol
```diff
@@ -9,11 +9,10 @@ interface ICollateralizationOracle is IOracle {
 
     // ----------- Getters -----------
 
-    function isOvercollateralized() external view returns (bool);
-
-    function userCirculatingFei() external view returns (uint256);
+    // returns the PCV value, User-circulating FEI, and Protocol Equity, as well
+    // as a validity status.
+    function pcvStats() external view returns (uint256, uint256, uint256, bool);
```
Joeysantoro, line 14: equity should be an int256 now right?

eswak: ok, I think it makes sense

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -267,32 +279,32 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
...
-                    if (!_oracleValid) {
-                      validityStatus = false;
+                // ignore deposits that are excluded by the Guardian
+                if (!excludedDeposits[_deposit]) {
+                    // On the first inspected deposit, read the oracle.
+                    // This is done inside the loop, after exclusion check,
+                    // because if all deposits of an asset are excluded, there is
+                    // no need to read the oracle.
+                    if (!_oracleRead) {
+                        (_oraclePrice, _oracleValid) = IOracle(tokenToOracle[_token]).read();
```
Joeysantoro, line 283: Would make more sense to put the read call once outside the loop right?

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,316 @@
...
+                    (uint256 _depositBalance, uint256 _depositFei) = IPCVDepositV2(_deposit).balanceAndFei();
+                    _totalTokenBalance += _depositBalance;
+                    _protocolControlledFei += _depositFei;
+                }
+            }
+
+            protocolControlledValue += _oraclePrice.mul(_totalTokenBalance).asUint256();
+        }
+
+        userCirculatingFei = fei().totalSupply() - _protocolControlledFei;
```
Joeysantoro, line 300: store fei.totalSupply in a variable to save gas

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,316 @@
...
+    ///         of the oracles for tokens held in the PCV are invalid, or if
+    ///         this contract is paused).
+    function pcvStats() public override view returns (
+      uint256 protocolControlledValue,
+      uint256 userCirculatingFei,
+      int256 protocolEquity,
+      bool validityStatus
+    ) {
+        uint256 _protocolControlledFei = 0;
+        address _fei = address(fei());
```
Joeysantoro, line 261: Variable not used

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,316 @@
...
+                    if (!_oracleRead) {
+                        (_oraclePrice, _oracleValid) = IOracle(tokenToOracle[_token]).read();
+                        _oracleRead = true;
+                        if (!_oracleValid) {
+                            validityStatus = false;
+                        }
+                    }
+
+                    // read the deposit
+                    (uint256 _depositBalance, uint256 _depositFei) = IPCVDepositV2(_deposit).balanceAndFei();
```
Joeysantoro, line 291: method should probably be named "resistantBalanceAndFei"

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,316 @@
...
+                }
+            }
+
+            protocolControlledValue += _oraclePrice.mul(_totalTokenBalance).asUint256();
+        }
+
+        userCirculatingFei = fei().totalSupply() - _protocolControlledFei;
+        protocolEquity = int256(protocolControlledValue) - int256(userCirculatingFei);
+
+        userCirculatingFei = fei().totalSupply() - _protocolControlledFei;
```
Joeysantoro, line 303: This line is duplicate of 300

contracts/oracle/ICollateralizationOracle.sol
```diff
@@ -9,11 +9,10 @@ interface ICollateralizationOracle is IOracle {
 
     // ----------- Getters -----------
 
-    function isOvercollateralized() external view returns (bool);
-
-    function userCirculatingFei() external view returns (uint256);
+    // returns the PCV value, User-circulating FEI, and Protocol Equity, as well
+    // as a validity status.
+    function pcvStats() external view returns (uint256, uint256, int256, bool);
```
Joeysantoro, line 14: Probably good to name the return values in interface so they are immediately recognizeable

contracts/pcv/IPCVDepositV2.sol
```diff
@@ -0,0 +1,16 @@
...
+    // ----------- Getters -----------
+
+    // gets the token address in which this deposit returns its balance
+    function balanceReportedIn() external view returns (address);
+
+    // gets the resistant token balance and protocol owned fei of this deposit
+    function balanceAndFei() external view returns (uint256, uint256);
```
Joeysantoro, line 15: method should probably be named "resistantBalanceAndFei"

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,316 @@
...
+
+        // emit event
+        emit DepositRemove(msg.sender, _deposit);
+    }
+
+    /// @notice Swap a PCVDeposit with a new one, for instance when a new version
+    ///         of a deposit (holding the same token) is deployed.
+    /// @param _oldDeposit : the PCVDeposit to remove from the list.
+    /// @param _newDeposit : the PCVDeposit to add to the list.
+    function swapDeposit(address _oldDeposit, address _newDeposit) external onlyGovernor {
```
Joeysantoro, line 160: Thoughts on just calling add then remove? It takes a bit more gas but may reduce complexity. This function is pretty simple so I don't feel strongly about it.

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,316 @@
...
+
+        // emit event
+        emit DepositRemove(msg.sender, _oldDeposit);
+        emit DepositAdd(msg.sender, _newDeposit, _token);
+    }
+
+    /// @notice Set the price feed oracle (in USD) for a given asset.
+    /// @param _token : the asset to add price oracle for
+    /// @param _newOracle : price feed oracle for the given asset
+    function setOracle(address _token, address _newOracle) external onlyGovernor {
```
Joeysantoro, line 188: nit: verify token and oracle are !=0

contracts/oracle/CollateralizationOracle.sol
eswak: best code is no code 👍

contracts/oracle/CollateralizationOracle.sol
eswak: it used to be above the loop, but if all deposits of this token are excluded, there is no need to fetch the oracle, that's why I put it in the loop.

I'll do even better, and fetch the oracle after the loop, if the tokenBalance is != 0

contracts/oracle/CollateralizationOracle.sol
Joeysantoro: Now that the logic is remove + add I don't think we need to enforce the tokens are the same as they are effectively distinct operations

Joeysantoro (approved):
Added one last nit. Contract looks fantastic

contracts/oracle/CollateralizationOracle.sol
```diff
@@ -0,0 +1,289 @@
...
+    /// @param _core Fei Core for reference
+    constructor(
+        address _core
+    ) CoreRef(_core) {}
+
+    // ----------- Convenience getters -----------
+
+    /// @notice returns true if a token is held in the pcv
+    function isTokenInPcv(address token) external view returns(bool) {
+        return tokensInPcv.contains(token);
```
sol-sr, line 62: Aren't we better off looking up a token in the tokenToOracle mapping and ensuring that the resulting address is not address 0 instead of enumerating over a list of addresses?

PR #148, opened 2021-09-02
Add proposal plugin

Joeysantoro:
Adds Idle Finance governor alpha proposal hardhat plugin

https://github.com/Idle-Finance/hardhat-proposals-plugin/blob/main/src/proposals/proposal.ts

Uses it to automatically construct the calldata from the json files in proposals/description

### DIFF
--- a/end-to-end/setup/index.ts
+++ b/end-to-end/setup/index.ts
@@ -11,6 +11,8 @@ import {
 } from './types'
 import { sudo } from '../../scripts/utils/sudo'
 import { exec } from '../../proposals/dao/exec'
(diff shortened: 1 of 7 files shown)


end-to-end/setup/index.ts
```diff
@@ -161,16 +162,4 @@ export class TestEndtoEndCoordinator implements TestCoordinator {
     // @ts-ignore
     return accessControlRoles;
   }
-
-  /**
```
Joeysantoro, line 165: Moved to `end-to-end/setup/loadContracts`

sebastiandelgado (commented):
This is sick

PR #150, opened 2021-09-02
Add new Scripts to proposal flow

Joeysantoro:
Uses the proposals hardhat plugin to quickly simulate proposals in the existing e2e test flow.

Adds two new scripts for generating calldata and checking a live proposal against the validate script.

### DIFF
--- a/proposals/README.md
+++ b/proposals/README.md
@@ -1,10 +1,28 @@
-# Making a DAO Proposal
+# Making and testing an FIP DAO Proposal
+Fei Improvement Proposals (FIPs) are the primary mechanism to update Fei Protocol functionality on-chain. 
+
+The proposal flow is automated and integrated into the end-to-end test suite off of live mainnet fork data. This ensures the effect of proposals can be thoroughly tested during development and pre/post execution.
+
 This guide assumes you already have enough TRIBE to make a proposal on-chain.
 
-## Optional Step 1: Does Proposal Require New Code?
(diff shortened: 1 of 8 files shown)


sebastiandelgado (approved):

proposals/README.md
```diff
@@ -17,61 +35,46 @@ To execute the deployment run:
...
+The goal of this step is to compare the execution results of the calldata to the mock script from step 3. Any setup and teardown calls in the `proposals/dao/fip_x.js` file will still execute, but the run step is replaced by the execution of the proposal steps in `proposals/description/fip_x.json`.
 
-To get the calldata, submit a proposal on https://www.withtally.com/governance/fei/proposal/new with the intended proposal commands, title and description. Click "Submit proposal" but **do not** confirm the Metamask transaction. 
-
-Copy the calldata into the config entry for `fip_x` in `end-to-end/proposals_config.json` for integration testing:
-* exec - set to true
-* proposal_calldata - fill in the copied calldata
-* proposer - fill in proposer address
-* voter - 0xB8f482539F2d3Ae2C9ea6076894df36D1f632775
+Set the exec flag to true into the config entry for `fip_x` in `end-to-end/proposals_config.json`
```
sebastiandelgado, line 71: This step will only be removable once we don't execute the proposal via js script, but exclusively via calldata from json file right?

sebastiandelgado (approved):

proposals/README.md
Joeysantoro: We'd need to remove the ability to include non-deployed contracts to the proposal which requires manipulating the json at runtime after doing the deploy script.

PR #151, opened 2021-09-03
V2 Tribe Reserve Stabilizer

Joeysantoro:
Part of Fei V2 Phase 1 that handles the inflation of TRIBE when the collateralization ratio is low.

Replaces the logic regarding the FEI peg/oracle with collateralization ratio logic. If the CR is below the configured threshold, then exchanging FEI for minted TRIBE becomes possible.

Any TRIBE held by the contract is transferred first before performing a direct mint.

Also restricts TRIBE withdrawals from the contract so TRIBE held is effectively burned barring a mint operation.

Uses the RateLimited utility to put a cap on how much TRIBE can be minted at once.

### DIFF
--- a/contracts/stabilizer/TribeReserveStabilizer.sol
+++ b/contracts/stabilizer/TribeReserveStabilizer.sol
@@ -54,12 +54,13 @@ contract TribeReserveStabilizer is ITribeReserveStabilizer, ReserveStabilizer {
 
     /// @dev reverts. Held TRIBE should only be released by exchangeFei or mint
     function withdraw(address, uint256) external pure override {
-        revert("TribeReserveStabilizer: can't withdraw");
+        revert("TribeReserveStabilizer: can't withdraw TRIBE");
     }
 
-    /// @dev reverts. Held TRIBE should only be released by exchangeFei or mint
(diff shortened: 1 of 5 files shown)


contracts/stabilizer/TribeReserveStabilizer.sol
```diff
@@ -103,8 +106,21 @@ contract TribeReserveStabilizer is ITribeReserveStabilizer, ReserveStabilizer {
         _tribe.setMinter(newMinter);
     }
 
+    // Transfer held TRIBE first, then mint to cover remainder
     function _transfer(address to, uint256 amount) internal override {
-        _mint(to, amount);
+        uint256 _tribeBalance = balance();
```
eswak, line 111: this is using the cached `token` variable from the contract state, while the `_mint()` call uses a fresh `ITribe(address(tribe()))`. Not sure if there could be some bad interaction here, but it would be nice to use the same source of information for both ?

Joeysantoro: Fixed, although it should be fine because token is immutably set to equal tribe();

eswak (approved):

ditchfieldcaleb (approved):

PR #153, opened 2021-09-05
ERC20 Splitter

Joeysantoro:
A contract to split held ERC20 tokens to multiple locations using PCVSplitter.

Also refactors PCVSplitter to extend CoreRef so it can share the setter for the allocation and the Allocate event

This will be used to split TRIBE that is bought back


eswak (approved):

contracts/token/TribeSplitter.sol
```diff
@@ -0,0 +1,31 @@
+pragma solidity ^0.8.0;
+
+import "../pcv/utils/PCVSplitter.sol";
+
+/// @title TribeSplitter
+/// @notice a contract to split TRIBE held to multiple locations
+contract TribeSplitter is PCVSplitter {
```
eswak, line 7: since there is no delegation of voting powers or anything, I guess you could even be more generic and make it an ERC20Splitter. We'll instanciate in on TRIBE through the constructor

Joeysantoro: Changed :)

eswak (approved):

ditchfieldcaleb (approved):

PR #154, opened 2021-09-05
Rate limited FEI minter

Joeysantoro:
Adds a helper abstract class which hard rate limits the speed with which FEI can be minted by a contract.

Uses a buffer that gets depleted as FEI is minted and replenishes over time up to the cap.

Does not use the Timed utility to allow extending contracts to use it for other use cases.

Extents the base _mintFei method of CoreRef so all minting operations are affected.

This class is intended to be used by the PCVEquityMinter once that is merged into the v2 branch

### DIFF
--- a/contracts/utils/RateLimited.sol
+++ b/contracts/utils/RateLimited.sol
@@ -1,0 +1,102 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "../refs/CoreRef.sol";
+import "@openzeppelin/contracts/utils/math/Math.sol";
+
+/// @title abstract contract for putting a rate limit on how fast a contract can perform an action e.g. Minting
+/// @author Fei Protocol
+abstract contract RateLimited is CoreRef {
+
+    /// @notice maximum rate limit per second governance can set for this contract
+    uint256 public immutable MAX_RATE_LIMIT_PER_SECOND;
+
+    /// @notice the rate per second for this contract
+    uint256 public rateLimitPerSecond;
+
+    /// @notice the last time the buffer was used by the contract
+    uint256 public lastBufferUsedTime;
+
+    /// @notice the cap of the buffer that can be used at once
+    uint256 public bufferCap;
+
+    /// @notice a flag for whether to allow partial actions to complete if the buffer is less than amount
+    bool public doPartialAction;
+
+    uint256 private _bufferStored;
+
+    event BufferCapUpdate(uint256 oldBufferCap, uint256 newBufferCap);
+    event RateLimitPerSecondUpdate(uint256 oldRateLimitPerSecond, uint256 newRateLimitPerSecond);
+
+    constructor(uint256 _maxRateLimitPerSecond, uint256 _rateLimitPerSecond, uint256 _bufferCap, bool _doPartialAction) {
+        lastBufferUsedTime = block.timestamp;
+
+        _bufferStored = _bufferCap;
+        _setBufferCap(_bufferCap);
+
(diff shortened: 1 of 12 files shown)


eswak (approved):
art 👌

contracts/utils/RateLimited.sol
```diff
@@ -0,0 +1,102 @@
...
+
+    function _depleteBuffer(uint256 amount) internal returns(uint256) {
+        uint256 newBuffer = buffer();
+        
+        uint256 usedAmount = amount;
+        if (doPartialAction && usedAmount > newBuffer) {
+            usedAmount = newBuffer;
+        }
+
+        require(usedAmount <= newBuffer, "RateLimitedMinter: rate limit hit");
```
eswak, line 70: forgot to rename the contract here

contracts/utils/RateLimited.sol
```diff
@@ -0,0 +1,102 @@
...
+
+        _bufferStored = newBuffer - usedAmount;
```
eswak, line 72: if `doPartialAction` is false, and buffer is not full, the expected behavior is that the call will revert on this line because of an uint underflow ? we could probably revert with a more graceful message

contracts/utils/RateLimited.sol
```diff
@@ -0,0 +1,102 @@
...
+    /// @notice the last time the buffer was used by the contract
+    uint256 public lastBufferUsedTime;
+
+    /// @notice the cap of the buffer that can be used at once
+    uint256 public bufferCap;
+
+    /// @notice a flag for whether to allow partial actions to complete if the buffer is less than amount
+    bool public doPartialAction;
+
+    uint256 private _bufferStored;
```
eswak, line 26: doc

contracts/utils/RateLimited.sol
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /// @notice the amount of action used before hitting limit
+    /// @dev replenishes at rateLimitPerSecond per second up to bufferCap
+    function buffer() public view returns(uint256) { 
+        uint256 elapsed = block.timestamp - lastBufferUsedTime;
+        return Math.min(_bufferStored + (rateLimitPerSecond * elapsed), bufferCap);
+    }
+
+    function _depleteBuffer(uint256 amount) internal returns(uint256) {
```
eswak, line 62: doc

contracts/utils/RateLimited.sol
Joeysantoro: It would revert with the more graceful message on line 70

contracts/utils/RateLimited.sol
Joeysantoro: Not conventionally needed on private vars but probably good to add here

contracts/utils/RateLimited.sol
Joeysantoro: Same as above

eswak (approved):

contracts/utils/RateLimitedMinter.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "./RateLimited.sol";
+
+/// @title abstract contract for putting a rate limit on how fast a contract can mint FEI
+/// @author Fei Protocol
+abstract contract RateLimitedMinter is RateLimited {
+
+    uint256 private constant MAX_FEI_LIMIT_PER_SECOND = 10_000e18; // 10000 FEI/s or ~860m FEI/day
```
ditchfieldcaleb, line 10: No reason to not add in methods here to have this changeable by governor.

Joeysantoro: Will merge as is and add in another PR

PR #155, opened 2021-09-06
Add Resistant Balances

Joeysantoro:
Finishes the work to update all PCVDeposit contracts by adding resistant balance calls

The ETH PCV deposits use the WETH address as balanceReportedIn because ETH doesn't have an address.

The only complex resistant balance calculation is uniswap, which has a derivation in the comments.

TODO: 
- [x] add tests
- [x] consider how to handle FEI PCV deposits and upgrades (possibly a wrapper/adapter)?


eswak:
I know it's not merged yet, but if you feel like spending an afternoon with a pen & paper, would be nice to derive the formula for resistant balance of the Curve pool too

Joeysantoro:
Made wrapper to adapt existing PCV deposits without redeploying: https://github.com/fei-protocol/fei-protocol-core/pull/156

contracts/bondingcurve/BondingCurve.sol
```diff
@@ -107,6 +107,18 @@ contract BondingCurve is IBondingCurve, OracleRef, PCVSplitter, Timed, Incentivi
...
+    /// @notice display the related token of the balance reported
+    /// @dev to allow compatibility with the Collateralization Oracle even though this contract isn't an IPCVDeposit
+    function balanceReportedIn() public view returns (address) {
+        return address(token);
+    }
+
+    /// @notice returns a manipulation resistant account of both the balance of underlying and protocol owned FEI
+    /// @dev to allow compatibility with the Collateralization Oracle even though this contract isn't an IPCVDeposit
+    function resistantBalanceAndFei() public view returns(uint256, uint256) {
+      return (balance(), 0);
```
eswak, line 119: I wonder if protocol contracts should report their current balance of FEI instead of 0. Like there could be a weird attack where FEI is sent to the protocol contracts, and the protocol believes it is undercollateralized & mint TRIBE, where actually it is holding the FEI. At first glance this looks like an attack that is a loss for the attacker, but idk there may be some wierd interaction somewhere ?

contracts/pcv/utils/PCVDepositWrapper.sol
```diff
@@ -0,0 +1,48 @@
...
+
+    /// @notice returns total balance of PCV in the Deposit
+    function balance() public view returns (uint256) {
+        return pcvDeposit.balance();
+    }
+
+    /// @notice returns the resistant balance and FEI in the deposit
+    function resistantBalanceAndFei() public view returns (uint256, uint256) {
+        uint256 resistantBalance = balance();
+        uint256 reistantFei = isProtocolFeiDeposit ? resistantBalance : 0;
```
eswak, line 40: So, FEI deposits add 1000 FEI to PCV and 1000 FEI to protocol-owned FEI, but we'll use an oracle price of 0 for FEI, so it will add 0$ to the PCV and 1000 FEI into circulation, correct ?

eswak (approved):

contracts/bondingcurve/BondingCurve.sol
ditchfieldcaleb: +1 this is definitely something to put some thought into - though, a special version of the Bonding Curve that implements logic susceptible to this kind of attack could just override this method.

If anything, it's better to leave it as the default of zero -  I cold see an attack far more likely where someone sends FEI borrowed from a lending pool to the contract, has the contract count it, and double-counts PCV this way.

contracts/pcv/utils/PCVDepositWrapper.sol
ditchfieldcaleb: ^ this is confusing and should be very clearly spelled out in docs if we go forward with it

contracts/pcv/utils/PCVDepositWrapper.sol
```diff
@@ -0,0 +1,48 @@
...
+  @notice a lightweight contract to wrap old PCV deposits to use the new interface 
+  @author Fei Protocol
+  When upgrading the PCVDeposit interface, there are many old contracts which do not support it.
+  The main use case for the new interface is to add read methods for the Collateralization Oracle.
+  Most PCVDeposits resistant balance method is simply returning the balance as a pass-through
+  If the PCVDeposit holds FEI it may be considered as protocol FEI
+
+  This wrapper can be used in the CR oracle which reduces the number of contract upgrades and reduces the complexity and risk of the upgrade
+*/
+contract PCVDepositWrapper {
```
ditchfieldcaleb, line 15: If this contract is wrapping something else to expose methods for the interface, why doesn't it implement the new interface?

ditchfieldcaleb (changes requested):
Feedback submitted. I still want to review the derivation for the Uniswap resistant balance but I'm sure its fine if it's been checked over a few times.

contracts/pcv/utils/PCVDepositWrapper.sol
Joeysantoro: It only implements the read parts of the interface. If it needed to literally wrap everything then it would have to implement withdraw and be given PCVController access. I guess we could implement the interface and make it a no-op. What do you think?

contracts/pcv/uniswap/IUniswapPCVDeposit.sol
```diff
@@ -1,12 +1,11 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity ^0.8.4;
 
-import "../IPCVDeposit.sol";
 import "@uniswap/v2-periphery/contracts/interfaces/IUniswapV2Router02.sol";
 
 /// @title a PCV Deposit interface
 /// @author Fei Protocol
-interface IUniswapPCVDeposit is IPCVDeposit {
+interface IUniswapPCVDeposit {
```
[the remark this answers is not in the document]
Joeysantoro: Good question, I believe there was an inheritance collision with `PCVDeposit` so I removed it from this interface

contracts/pcv/compound/EthCompoundPCVDeposit.sol
```diff
@@ -11,6 +11,9 @@ interface CEther {
 /// @author Fei Protocol
 contract EthCompoundPCVDeposit is CompoundPCVDepositBase {
 
+    /// @dev used as the reporting token address for balances
+    address public constant WETH = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
```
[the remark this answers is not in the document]
Joeysantoro: I'll make a task to add this if I get time in a separate PR

contracts/bondingcurve/BondingCurve.sol
```diff
@@ -107,6 +107,18 @@ contract BondingCurve is IBondingCurve, OracleRef, PCVSplitter, Timed, Incentivi
         return token.balanceOf(address(this));
     }
 
+    /// @notice display the related token of the balance reported
+    /// @dev to allow compatibility with the Collateralization Oracle even though this contract isn't an IPCVDeposit
```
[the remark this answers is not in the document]
Joeysantoro: This is a fantastic callout and I agree. I'll see how easy it is to refactor out the read methods into another interface but that would also deserve another PR

contracts/pcv/utils/PCVDepositWrapper.sol
Joeysantoro: @eswak 's comment was the intended behavior. Alternatively we can return 0 for the balance on Fei deposits to be explicit. @ditchfieldcaleb is the intended outcome confusing or just the implementation?

contracts/bondingcurve/BondingCurve.sol
Joeysantoro: Default of 0 makes most sense in this context because the FEI is likely held as a deposit receipt in Fuse or whatever. Alternatively we could add a distinct "Fei" PCV deposit that reports in-line with expectations for FEI

contracts/pcv/utils/PCVDepositWrapper.sol
eswak: let's make it inherit the new interface that is not deposit() but that the CR oracle reads & that is both inherited by PCVDeposits and BondingCurves

contracts/pcv/compound/EthCompoundPCVDeposit.sol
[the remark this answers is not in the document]
Joeysantoro: https://github.com/fei-protocol/fei-protocol-core/pull/165/files

PR #157, opened 2021-09-07
Collateralization Memoizer

eswak:
This PR adds a new contract to cache calls to the CollateralizationOracle

### DIFF
--- a/contracts/oracle/CollateralizationMemoizer.sol
+++ b/contracts/oracle/CollateralizationMemoizer.sol
@@ -1,0 +1,204 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "./IOracle.sol";
+import "./ICollateralizationOracle.sol";
+import "../refs/CoreRef.sol";
+
+interface IPausable {
+    function paused() external view returns (bool);
+}
+
+/// @title Fei Protocol's Collateralization Oracle
+/// @author eswak
+/// @notice Reads a list of PCVDeposit that report their amount of collateral
+///         and the amount of protocol-owned FEI they manage, to deduce the
+///         protocol-wide collateralization ratio.
+contract CollateralizationMemoizer is IOracle, CoreRef {
+    using Decimal for Decimal.D256;
+
+    // ----------- Events -----------
+
+    event CachedValueUpdate(
+        address from,
+        uint256 indexed protocolControlledValue,
+        uint256 indexed userCirculatingFei,
+        int256 indexed protocolEquity
+    );
+
+    // ----------- Properties -----------
+
+    /// @notice address of the CollateralizationOracle to memoize
+    address public collateralizationOracle;
+
+    /// @notice cached value of the Protocol Controlled Value
+    uint256 public cachedProtocolControlledValue;
+    /// @notice cached value of the User Circulating FEI
+    uint256 public cachedUserCirculatingFei;
+    /// @notice cached value of the Protocol Equity
+    int256 public cachedProtocolEquity;
+
+    /// @notice timestamp of the last read operation on the CollateralizationOracle
+    uint256 public lastUpdate;
+    /// @notice validity duration of the cached value
+    uint256 public validityDuration;
+    /// @notice deviation threshold to consider cached values outdated, in basis
+    ///         points (base 10_000)
+    uint256 public deviationThreshold;
+
+    // ----------- Constructor -----------
+
+    /// @notice CollateralizationOracle constructor
+    /// @param _core Fei Core for reference
+    constructor(
+        address _core
+    ) CoreRef(_core) {}
+
+    // ----------- IOracle override methods -----------
+    /// @notice update reading of the CollateralizationOracle
+    function update() external override whenNotPaused {
+        // fetch a fresh round of information
+        (
+            uint256 _protocolControlledValue,
+            uint256 _userCirculatingFei,
+            int256 _protocolEquity,
+            bool _validityStatus
+        ) = ICollateralizationOracle(collateralizationOracle).pcvStats();
+
+        // only update if valid
+        require(_validityStatus, "CollateralizationMemoizer: CollateralizationOracle is invalid");
+
+        // Update cached variables
+        _setCache(_protocolControlledValue, _userCirculatingFei, _protocolEquity);
+    }
+
+    // @notice returns true if the cached values are outdated.
+    function isOutdated() external override view returns (bool outdated) {
+        // check if cached value is fresh
+        if (block.timestamp > lastUpdate + validityDuration) {
+            outdated = true;
+        }
+        // check if deviation thresholds are met
+        if (!outdated) {
+            (
+                uint256 _protocolControlledValue,
+                uint256 _userCirculatingFei,
+                , // int256 _protocolEquity,
+                bool _validityStatus
+            ) = ICollateralizationOracle(collateralizationOracle).pcvStats();
+
+            require(_validityStatus, "CollateralizationMemoizer: CollateralizationOracle reading is invalid");
+
+            Decimal.D256 memory _thresholdLow = Decimal.from(10_000 - deviationThreshold).div(10_000);
+            Decimal.D256 memory _thresholdHigh = Decimal.from(10_000 + deviationThreshold).div(10_000);
+
+            // PCV < threshold
+            Decimal.D256 memory pcvRatio = Decimal.from(_protocolControlledValue).div(cachedProtocolControlledValue);
+            outdated = outdated || pcvRatio.lessThan(_thresholdLow);
+            // PCV > threshold
+            if (!outdated) {
+                outdated = outdated || pcvRatio.greaterThan(_thresholdHigh);
+            }
+            // CircFEI < threshold
+            if (!outdated) {
+                Decimal.D256 memory circFeiRatio = Decimal.from(_userCirculatingFei).div(cachedUserCirculatingFei);
+                outdated = outdated || circFeiRatio.lessThan(_thresholdLow);
+                // CircFEI > threshold
+                if (!outdated) {
+                    outdated = outdated || circFeiRatio.greaterThan(_thresholdHigh);
+                }
+            }
+            // no need to check protocol equity, it is deduced from the other 2
+        }
+    }
+
+    /// @notice Get the current collateralization ratio of the protocol, from cache.
+    /// @return collateralRatio the current collateral ratio of the protocol.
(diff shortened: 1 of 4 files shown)


contracts/oracle/CollateralizationMemoizer.sol
```diff
@@ -0,0 +1,204 @@
...
+interface IPausable {
+    function paused() external view returns (bool);
+}
+
+/// @title Fei Protocol's Collateralization Oracle
+/// @author eswak
+/// @notice Reads a list of PCVDeposit that report their amount of collateral
+///         and the amount of protocol-owned FEI they manage, to deduce the
+///         protocol-wide collateralization ratio.
+contract CollateralizationMemoizer is IOracle, CoreRef {
```
Joeysantoro, line 17: Can you rename CollateralizationOracleWrapper to be consistent with other naming?

contracts/oracle/CollateralizationMemoizer.sol
```diff
@@ -0,0 +1,204 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "./IOracle.sol";
+import "./ICollateralizationOracle.sol";
+import "../refs/CoreRef.sol";
```
Joeysantoro, line 6: This contract should use `../utils/Timed.sol` to keep track of the validityDuration and lastUpdate. Is there a reason why you didn't?

contracts/oracle/CollateralizationMemoizer.sol
```diff
@@ -0,0 +1,204 @@
...
+    ///         Otherwise, call the CollateralizationOrache to get fresh data.
+    /// @return protocolControlledValue : the total USD value of all assets held
+    ///         by the protocol.
+    /// @return userCirculatingFei : the number of FEI not owned by the protocol.
+    /// @return protocolEquity : the difference between PCV and user circulating FEI.
+    ///         If there are more circulating FEI than $ in the PCV, equity is 0.
+    /// @return validityStatus : the current oracle validity status (false if any
+    ///         of the oracles for tokens held in the PCV are invalid, or if
+    ///         this contract is paused).
+    function pcvStats() external returns (
```
Joeysantoro, line 138: This should be a view method, likely split in two. pcvStats should simply return the cached values with a validity status equal to whether it is out of date or not

A second view method maybe called pcvStatsCurrent passes through to the CR oracle and returns most up to date values

contracts/oracle/CollateralizationMemoizer.sol
```diff
@@ -0,0 +1,204 @@
...
+    /// @notice cached value of the Protocol Equity
+    int256 public cachedProtocolEquity;
+
+    /// @notice timestamp of the last read operation on the CollateralizationOracle
+    uint256 public lastUpdate;
+    /// @notice validity duration of the cached value
+    uint256 public validityDuration;
+    /// @notice deviation threshold to consider cached values outdated, in basis
+    ///         points (base 10_000)
+    uint256 public deviationThreshold;
```
Joeysantoro, line 47: nit: rename `deviationThresholdBasisPoints`

contracts/oracle/CollateralizationMemoizer.sol
```diff
@@ -0,0 +1,204 @@
...
+
+    /// @notice CollateralizationOracle constructor
+    /// @param _core Fei Core for reference
+    constructor(
+        address _core
+    ) CoreRef(_core) {}
+
+    // ----------- IOracle override methods -----------
+    /// @notice update reading of the CollateralizationOracle
+    function update() external override whenNotPaused {
```
Joeysantoro, line 59: should this method also call update on the CR oracle? I'd lean no but maybe it should

contracts/oracle/CollateralizationMemoizer.sol
```diff
@@ -0,0 +1,204 @@
...
+
+        // only update if valid
+        require(_validityStatus, "CollateralizationMemoizer: CollateralizationOracle is invalid");
+
+        // Update cached variables
+        _setCache(_protocolControlledValue, _userCirculatingFei, _protocolEquity);
+    }
+
+    // @notice returns true if the cached values are outdated.
+    function isOutdated() external override view returns (bool outdated) {
```
Joeysantoro, line 76: I think this function should use early returning liberally to simplify the logic

contracts/oracle/CollateralizationMemoizer.sol
```diff
@@ -0,0 +1,204 @@
...
+            (
+                uint256 _protocolControlledValue,
+                uint256 _userCirculatingFei,
+                , // int256 _protocolEquity,
+                bool _validityStatus
+            ) = ICollateralizationOracle(collateralizationOracle).pcvStats();
+
+            require(_validityStatus, "CollateralizationMemoizer: CollateralizationOracle reading is invalid");
+
+            Decimal.D256 memory _thresholdLow = Decimal.from(10_000 - deviationThreshold).div(10_000);
```
Joeysantoro, line 92: maybe add a helper method that returns % diff between two numbers then invalidate if that number exceeds the threshold. Would be simpler logic

contracts/oracle/CollateralizationMemoizer.sol
Joeysantoro: This method should split in two, isOutdated returning the time condition only and a second called isExceededDeviationThreshold

contracts/oracle/CollateralizationMemoizer.sol
eswak: no reason, I just needed a timestamp, didn't feel like inheriting the whole class, but why not

contracts/oracle/CollateralizationMemoizer.sol
eswak: let's leave it like this for now

contracts/oracle/CollateralizationMemoizer.sol
eswak: I did not use early returns, since the function is intended to be called offchain only, I kept the code as simple as possible without too much branching

contracts/oracle/CollateralizationOracleWrapper.sol
```diff
@@ -0,0 +1,272 @@
...
+        collateralizationOracle = _collateralizationOracle;
+        deviationThresholdBasisPoints = _deviationThresholdBasisPoints;
+    }
+
+    // ----------- Setter methods ----------------------------------------------
+
+    /// @notice set the address of the CollateralizationOracle to inspect, and
+    /// to cache values from.
+    /// @param _newCollateralizationOracle the address of the new oracle.
+    function setCollateralizationOracle(address _newCollateralizationOracle) external onlyGovernor {
```
Joeysantoro, line 81: require !=0

contracts/oracle/CollateralizationOracleWrapper.sol
```diff
@@ -0,0 +1,272 @@
...
+            msg.sender,
+            _oldCollateralizationOracle,
+            _newCollateralizationOracle
+        );
+    }
+
+    /// @notice set the deviation threshold in basis points, used to detect if the
+    /// cached value deviated significantly from the actual fresh readings.
+    /// @param _newDeviationThresholdBasisPoints the new value to set.
+    function setDeviationThresholdBasisPoints(uint256 _newDeviationThresholdBasisPoints) external onlyGovernor {
```
Joeysantoro, line 96: require !=0 && less than 10_000

contracts/oracle/CollateralizationOracleWrapper.sol
```diff
@@ -0,0 +1,272 @@
...
+        (
+            uint256 _protocolControlledValue,
+            uint256 _userCirculatingFei,
+            , // int256 _protocolEquity,
+            bool _validityStatus
+        ) = ICollateralizationOracle(collateralizationOracle).pcvStats();
+
+        require(_validityStatus, "CollateralizationMemoizer: CollateralizationOracle reading is invalid");
+
+        Decimal.D256 memory _thresholdLow = Decimal.from(10_000 - deviationThresholdBasisPoints).div(10_000);
```
Joeysantoro, line 178: I still think a helper method would be cleaner but this looks fine so I won't die on that hill

contracts/oracle/CollateralizationOracleWrapper.sol
```diff
@@ -0,0 +1,272 @@
...
+        protocolControlledValue = cachedProtocolControlledValue;
+        userCirculatingFei = cachedUserCirculatingFei;
+        protocolEquity = cachedProtocolEquity;
+    }
+
+    /// @notice returns true if the protocol is overcollateralized. Overcollateralization
+    ///         is defined as the protocol having more assets in its PCV (Protocol
+    ///         Controlled Value) than the circulating (user-owned) FEI, i.e.
+    ///         a positive Protocol Equity.
+    ///         Note: the validity status is ignored in this function.
```
Joeysantoro, line 238: This comment is innacurate

contracts/oracle/CollateralizationOracleWrapper.sol
```diff
@@ -0,0 +1,272 @@
...
+    /// @return validityStatus : the current oracle validity status (false if any
+    ///         of the oracles for tokens held in the PCV are invalid, or if
+    ///         this contract is paused).
+    function pcvStatsCurrent() external view returns (
+        uint256 protocolControlledValue,
+        uint256 userCirculatingFei,
+        int256 protocolEquity,
+        bool validityStatus
+    ) {
+      bool fetchedValidityStatus;
```
Joeysantoro, line 262: this fetchedValidityStatus isn't used

eswak: resolved, it was a bug

contracts/oracle/CollateralizationOracleWrapper.sol
```diff
@@ -0,0 +1,273 @@
...
+            , // int256 _protocolEquity,
+            bool _validityStatus
+        ) = ICollateralizationOracle(collateralizationOracle).pcvStats();
+
+        require(_validityStatus, "CollateralizationOracleWrapper: CollateralizationOracle reading is invalid");
+
+        Decimal.D256 memory _thresholdLow = Decimal.from(10_000 - deviationThresholdBasisPoints).div(10_000);
+        Decimal.D256 memory _thresholdHigh = Decimal.from(10_000 + deviationThresholdBasisPoints).div(10_000);
+
+        obsolete = paused();
```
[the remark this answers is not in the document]
Joeysantoro: Will merge as is and add a refactor in another PR

Joeysantoro (approved):

PR #179, opened 2021-09-14
Cleanup LBP Withdrawal

Joeysantoro:
Remove slippage protection param from BalancerLBPSwapper because it isn't useful unless an external oracle is integrated

### DIFF
--- a/end-to-end/e2e.spec.ts
+++ b/end-to-end/e2e.spec.ts
@@ -40,6 +40,38 @@ describe('e2e', function () {
     await contracts.uniswapPCVDeposit.deposit()
   })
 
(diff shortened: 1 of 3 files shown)


end-to-end/e2e.spec.ts
```diff
@@ -40,6 +40,38 @@ describe('e2e', function () {
     await contracts.uniswapPCVDeposit.deposit()
   })
 
+  describe('PCV Equity Minter + LBP', async function() {
```
Joeysantoro, line 43: Moved because of an order of operation dependency. CR oracle changes mess this up

ditchfieldcaleb (approved):

PR #185, opened 2021-09-22
Add Admin support on contract level for Fei Protocol

Joeysantoro:
We introduced the concept of a "contract admin" with OA for TribalChief.

This PR adds admin support to contracts throughout the repo following the below general rule of thumb:
* integer parameters and flags OK
* Address changes NOT OK

Notable exceptions:
* CR oracle constituents have admin enabled
* TribeReserveStabilizer parameters do NOT have admin

NOTE: This PR will not affect any existing on-chain contracts. And future contracts will explicitly need to define an admin role and have this role granted to a multisig timelock through the DAO. By itself these contract changes are effectively a no-op


eswak:
the code looks ok, do we add tests for this trivial change ?

Joeysantoro:
No need, modifier already tested in CoreRef test

ditchfieldcaleb (approved):
Approved but see comments in Slack.

PR #187, opened 2021-09-22
Upgrade DAO to OZ Governor

Joeysantoro:
This PR upgrades the Fei governor to OZ Governor with the following changes:
* delay is only 1 block (from 3333)
* voting period is 13000 blocks (from 10000)
* proposal threshold, delay, voting period, and quorum all now configurable

TODO:
- [x] complete DAO deployment steps
- [x] e2e test future proposal
- [x] unit tests
- [x] check if proposals plugin breaks https://github.com/Idle-Finance/hardhat-proposals-plugin/blob/main/src/proposals/compound-alpha.ts


ditchfieldcaleb (approved):

test/unit/dao/FeiDao.test.ts
```diff
@@ -0,0 +1,189 @@
...
+        method: 'hardhat_impersonateAccount',
+        params: [address]
+      });
+
+      impersonatedSigners[address] = await ethers.getSigner(address);
+    }
+  });
+
+  beforeEach(async function () {
+    ({ userAddress } = await getAddresses());
```
ditchfieldcaleb, line 35: Why is this in parens?

test/unit/dao/FeiDao.test.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    await network.provider.request({
+      method: 'hardhat_reset',
+      params: []
+    });
+
+    this.core = await getCore();
```
ditchfieldcaleb, line 42: `this` inside of a mocha test should be avoided; instead, declare vars inside the uppermost "describe" block and reference these to avoid weird scoping issues, especially when nesting describe hooks

source: https://stackoverflow.com/questions/27358338/mocha-this-in-before-and-beforeeach-hooks

test/unit/dao/FeiDao.test.ts
```diff
@@ -0,0 +1,189 @@
...
+    // Deploy timelock with vanilla user admin
+    const timelockDeployer = await ethers.getContractFactory(Timelock.abi, Timelock.bytecode);
+    this.timelock = await timelockDeployer.deploy(userAddress, 1, 1); // delay and min delay 1s
+
+    // Deploy new Fei DAO
+    const feiDAODeployer = await ethers.getContractFactory(FeiDAO.abi, FeiDAO.bytecode);
+    this.feiDAO = await feiDAODeployer.deploy(tribeAddress, this.timelock.address);
+
+    // Update timelock admin to Fei DAO
+    await hre.network.provider.request({
```
ditchfieldcaleb, line 54: Not something you need to fix, but I'm going to add a utility method that does this requesting via hardhat and such for us, so that getting a new signer is a one-liner.

ditchfieldcaleb (approved):
Beautiful unit tests. Left some comments; only thing that really needs changing imo is the `this` thing.

ditchfieldcaleb:
Didn't mean to actually approve the PR since integration tests haven't been written yet, oops.

test/unit/dao/FeiDao.test.ts
Joeysantoro: sick

test/unit/dao/FeiDao.test.ts
Joeysantoro: not sure but ts yells at me if not

Joeysantoro:
Pretty sure the proposals plugin is compatible with OZ Gov the way we use it: 

https://github.com/Idle-Finance/hardhat-proposals-plugin/blob/main/src/proposals/compound-alpha.ts#L98

it uses a static call to get the pid, which would work for OZ gov.

contracts/dao/FeiDAO.sol
```diff
@@ -0,0 +1,156 @@
...
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        string memory description
+    )
+        public
+        override(IGovernor, Governor, GovernorCompatibilityBravo)
+        returns (uint256)
+    {
+        return super.propose(targets, values, calldatas, description);
```
ElliotFriedman, line 110: Does this function need to be permissioned or does the function we call here enforce that?

test/unit/dao/FeiDao.test.ts
```diff
@@ -0,0 +1,189 @@
...
+      method: 'hardhat_reset',
+      params: []
+    });
+
+    this.core = await getCore();
+    const tribeAddress = await this.core.tribe();
+
+    // Deploy timelock with vanilla user admin
+    const timelockDeployer = await ethers.getContractFactory(Timelock.abi, Timelock.bytecode);
+    this.timelock = await timelockDeployer.deploy(userAddress, 1, 1); // delay and min delay 1s
```
ditchfieldcaleb, line 47: Since you already wrote this I hate to ask, but can you remove all references to `this` given the importance of the upgrade?

ditchfieldcaleb (changes requested):

contracts/dao/FeiDAO.sol
Joeysantoro: the supercall checks proposal threshold

test/unit/dao/FeiDao.test.ts
Joeysantoro: Fixed. I didn't type Core because it had some upstream stuff I didn't want to mess with

PR #201, opened 2021-09-28
make addresses a ts object, and fix the bugs this causes

ditchfieldcaleb:
This is PR 1 of 2 for the mainnet addresses stuff.
This one simply changes the addresses json object into a ts object, and fixes the import errors that result from this.
Also deleted extraneous types file and fixed the imports to point to the correct types file.


contract-addresses/mainnetAddresses.ts
```diff
@@ -0,0 +1,261 @@
...
+  rariPool9RaiPCVDeposit: {
+    artifactName: 'ERC20CompoundPCVDeposit',
+    address: '0x9aAdFfe00eAe6d8e59bB4F7787C6b99388A6960D'
+  },
+  ratioPCVController: { artifactName: 'RatioPCVController', address: '0xfC1aD6eb84351597cD3b9B65179633697d65B920' },
+  reflexerStableAssetFusePoolRai: {
+    artifactName: 'CErc20Delegator',
+    address: '0x752F119bD4Ee2342CE35E2351648d21962c7CAfE'
+  },
+  reflexerStableAssetFusePoolRaiPCVDeposit: {
```
ElliotFriedman, line 232: What kind of linter is being used here? We should get everyone on the team set up with the same linter so all of our code conforms to a standard style.

ditchfieldcaleb: `npm run lint:fix` and `npm run prettier-format`

Joeysantoro (approved):

PR #212, opened 2021-10-03
FIP-30

Joeysantoro:
This PR creates the DAO proposal steps for upgrading to the OZ governor. It also introduces a rollback mechanism which allows the guardian to revert back to a backup Governor Alpha in the event that the upgrade fails for one reason or another.

The rollback can be used at most once, and expires automatically on November 1 if not used beforehand.

TODO:
- [x] Add e2e test for rollback
- [x] Deploy FeiDAO
- [x] Fix unit tests and constants with new deploy address
- [x] Write json for proposal


contracts/dao/FeiDAO.sol
```diff
@@ -78,6 +91,33 @@ contract FeiDAO is
...
     }
 
+    /// @notice one-time option to roll back the DAO to old GovernorAlpha
+    /// @dev admin-only, and expires after the deadline. This function is here as a fallback in case something goes wrong.
+    function __rollback(uint256 eta) external {
+        require(msg.sender == _admin, "FeiDAO: caller not admin");
+        // Deleting admin prevents multiple triggers of this function
+        _admin = address(0);
+
+        require(eta <= ROLLBACK_DEADLINE, "FeiDAO: rollback expired");
```
ditchfieldcaleb, line 101: We should also require that eta < now. I know this is checked further along when the timelock checks this too, but checking here makes it simpler to analyze.

contracts/dao/FeiDAO.sol
```diff
@@ -78,6 +91,33 @@ contract FeiDAO is
...
+        _eta = eta;
+
+        ICompoundTimelock _timelock = ICompoundTimelock(payable(timelock()));
+        _timelock.queueTransaction(timelock(), 0, "setPendingAdmin(address)", abi.encode(OLD_GOVERNOR), eta);
```
ditchfieldcaleb, line 105: Is abi.encode here correct or should it be encodePacked? (Are you taking this from an existing example?)

contracts/dao/FeiDAO.sol
Joeysantoro: The timelock will revert if eta is < now. I'd lean against adding it but if you feel strongly its not much different

contracts/dao/FeiDAO.sol
Joeysantoro: This is forked from compound's GovernorAlpha logic. encode is how you format something for an actual external call, encodePacked ignores padding and is used for hashing.

contracts/dao/FeiDAO.sol
ditchfieldcaleb: Fine by me

scripts/deploy/fip_30.ts
```diff
@@ -1,19 +1,26 @@
 import { DeployUpgradeFunc } from '../../types/types';
 import { ethers } from 'hardhat';
 
+const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000';
```
ditchfieldcaleb, line 4: use ethers.constants.AddressZero

test/integration/e2e.spec.ts
```diff
@@ -114,6 +114,30 @@ describe('e2e', function () {
       expect((await feiDAO.votingDelay()).toString()).to.be.equal('10');
       expect((await contracts.daiBondingCurve.duration()).toString()).to.be.equal('11');
     });
+
+    it('rollback succeeds', async function() {
+      const { feiDAO, timelock, governorAlpha } = contracts;
+      const { multisig } = contractAddresses;
+
+      const signer = await ethers.getSigner(multisig);
```
ditchfieldcaleb, line 122: I *believe* this is out of order; I think you need to impersonate the account and then get the signer. Could be wrong.

Either way if you rebase off of master you can use the new helper method I just added.

test/unit/dao/FeiDao.test.ts
```diff
@@ -97,6 +98,52 @@ describe('FeiDAO', function () {
...
       expect(await feiDAO.timelock()).to.be.equal(timelock.address);
     });
+
+    it('rollback deadline', async function () {
+      expect((await feiDAO.ROLLBACK_DEADLINE()).toString()).to.be.equal(
+        '1635724800'
+      );
+    });
+
+    it('old governor', async function () {
```
ditchfieldcaleb, line 108: nit: the descriptions should tell you what the test is doing. this computes to: "FeiDAO old governor" (old governor...what?)

better: "FeiDAO old governor address is correct"

test/unit/dao/FeiDao.test.ts
```diff
@@ -97,6 +98,52 @@ describe('FeiDAO', function () {
...
+      expect(await feiDAO.OLD_GOVERNOR()).to.be.equal('0xE087F94c3081e1832dC7a22B48c6f2b5fAaE579B');
```
ditchfieldcaleb, line 109: this address doesn't match the one set in the FeiDAO smart contract

ditchfieldcaleb (changes requested):
Looks good, but the address set in the FeiDAO smart contract for OLD_GOVERNOR doesn't match the address in the tests.

test/integration/e2e.spec.ts
Joeysantoro: will use going forward

ditchfieldcaleb (approved):

PR #215, opened 2021-10-04
Make LBP swap ACL

Joeysantoro:
makes the LBP swap method only callable by admin. The admin role will be granted to PCVEquityMinter so that whenever mint is called, it passes through to swap.

PR also deletes many unused variables and functions to simplify.

Resolves 3.6 AND 3.12 in Dili audit


ditchfieldcaleb (commented):
This looks fine but I want to review the balancer pool manager in depth tomorrow and get back to it.

Joeysantoro:
> This looks fine but I want to review the balancer pool manager in depth tomorrow and get back to it.

Sounds good

ditchfieldcaleb (approved):

PR #225, opened 2021-10-05
Deploy script FIP-30b

Joeysantoro:


scripts/deploy/fip_30b.ts
```diff
@@ -0,0 +1,25 @@
...
+  const feiDAOTimelock = await factory.connect(signer).deploy(core, feiDAO, delay, minDelay);
+
+  logging && console.log('FeiDAOTimelock deployed to: ', feiDAOTimelock.address);
+
+  return {
+    feiDAOTimelock
+  };
+};
+
+module.exports = { deploy };
```
ditchfieldcaleb, line 25: do `export const  deploy = ` instead of using module.exports

ditchfieldcaleb (changes requested):

scripts/deploy/fip_30b.ts
ditchfieldcaleb: (they are *almost* the same thing but they are not; because the target in `tsconfig.json` is es2018, we should be import/export instead of require/module.exports)

contracts/dao/FeiDAOTimelock.sol
```diff
@@ -39,4 +42,16 @@ contract FeiDAOTimelock is Timelock, CoreRef {
         pendingAdmin = newAdmin;
         emit NewPendingAdmin(newAdmin);
     }
+
+    /// @notice one-time option to roll back the Timelock to old timelock
+    /// @dev guardian-only, and expires after the deadline. This function is here as a fallback in case something goes wrong.
+    function rollback() external onlyGuardianOrGovernor {
+        require(block.timestamp <= ROLLBACK_DEADLINE, "FeiDAOTimelock: rollback expired");
+        
+        IFeiDAO(admin).updateTimelock(OLD_TIMELOCK);
```
ditchfieldcaleb, line 51: Does this pass the onlyGovernance check on updateTimelock?

ditchfieldcaleb: Other than this looks good

ditchfieldcaleb (approved):

contracts/dao/FeiDAOTimelock.sol
Joeysantoro: it does, onlyGovernance requires the msg.senser to be itself if no timelock, otherwise the timelock

Joeysantoro: Will confirm in e2e

PR #232, opened 2021-10-10
Tribe Rate Limit

Joeysantoro:
Separates all Tribe Minting logic out from TribeReserveStabilizer into a TribeMinter contract
  
-------
  @title implementation for a TRIBE Minter Contract
  @author Fei Protocol

  This contract will be the unique TRIBE minting contract. 
  All minting is subject to an annual inflation rate limit.
  For example if circulating supply is 1m and inflation is capped at 10%, then no more than 100k TRIBE can enter circulation in the following year.

  The contract will increase (decrease) the rate limit proportionally as supply increases (decreases)

  Governance and admins can only lower the max inflation %. 
  They can also exclude (unexclude) addresses' TRIBE balances from the circulating supply. 
  The minter's balance is excluded by default.

  ACCESS_CONTROL:
  This contract follows a somewhat unique access control pattern. 
  It has a contract admin which is NOT intended for optimistic approval, but rather for contracts such as the TribeReserveStabilizer.
  An additional potential contract admin is one which automates the inclusion and removal of excluded deposits from on-chain timelocks.

  Additionally, the ability to transfer the tribe minter role is held by the contract *owner* rather than governor or admin.
  The owner will intially be the DAO timelock.
  This keeps the power to transfer or burn TRIBE minting rights isolated.
  
  -------
  Also makes some light changes to RateLimited contract

### DIFF
--- a/contracts/dao/TribeMinter.sol
+++ b/contracts/dao/TribeMinter.sol
@@ -1,0 +1,207 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "./ITribeMinter.sol";
+import "../utils/RateLimited.sol";
+import "../Constants.sol";
+import "@openzeppelin/contracts/utils/math/Math.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+
+/** 
+  @title implementation for a TRIBE Minter Contract
+  @author Fei Protocol
+
+  This contract will be the unique TRIBE minting contract. 
+  All minting is subject to an annual inflation rate limit.
+  For example if circulating supply is 1m and inflation is capped at 10%, then no more than 100k TRIBE can enter circulation in the following year.
+
+  The contract will increase (decrease) the rate limit proportionally as supply increases (decreases)
+
+  Governance and admins can only lower the max inflation %. 
+  They can also exclude (unexclude) addresses' TRIBE balances from the circulating supply. 
+  The minter's balance is excluded by default.
+
+  ACCESS_CONTROL:
+  This contract follows a somewhat unique access control pattern. 
+  It has a contract admin which is NOT intended for optimistic approval, but rather for contracts such as the TribeReserveStabilizer.
+  An additional potential contract admin is one which automates the inclusion and removal of excluded deposits from on-chain timelocks.
+
+  Additionally, the ability to transfer the tribe minter role is held by the contract *owner* rather than governor or admin.
+  The owner will intially be the DAO timelock.
+  This keeps the power to transfer or burn TRIBE minting rights isolated.
+*/
+contract TribeMinter is ITribeMinter, RateLimited, Ownable {
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    uint256 public override annualMaxInflationBasisPoints;
+
+    EnumerableSet.AddressSet internal _lockedTribeAddresses;
+
+    /// @notice Tribe Reserve Stabilizer constructor
+    /// @param _core Fei Core to reference
+    constructor(
+        address _core,
+        uint256 _annualMaxInflationBasisPoints,
+        address _owner,
+        address[] memory _lockedTribeAddressList
+    ) 
+      RateLimited(0, 0, 0, false)
+      CoreRef(_core)
+    {
+        _setAnnualMaxInflationBasisPoints(_annualMaxInflationBasisPoints);
+        _poke();
+
+        // start with a full buffer
+        _resetBuffer();
+
+        transferOwnership(_owner);
+
+        _addLockedTribeAddress(address(this));
+        for (uint256 i = 0; i < _lockedTribeAddressList.length; i++) {
+            _addLockedTribeAddress(_lockedTribeAddressList[i]);
+        }
+    }
+
+    /// @notice update the rate limit per second and buffer cap
+    function poke() public override {
+        (uint256 oldBufferCap, uint256 newBufferCap) = _poke();
+
+        // Increasing the buffer cap shouldn't also increase capacity atomically
+        // Deplete buffer by the newly increased cap difference
+        if (newBufferCap > oldBufferCap) {
+            uint256 increment = newBufferCap - oldBufferCap;
+            _depleteBuffer(increment);
+        }
+    }
+
+    /// @dev no-op, reverts. Prevent admin or governor from overwriting ideal rate limit
+    function setRateLimitPerSecond(uint256) external pure override {
+        revert("no-op");
+    }
+
+    /// @dev no-op, reverts. Prevent admin or governor from overwriting ideal buffer cap
+    function setBufferCap(uint256) external pure override { 
+        revert("no-op");
+    }
+
+    /// @notice mints TRIBE to the target address, subject to rate limit
+    /// @param to the address to send TRIBE to
+    /// @param amount the amount of TRIBE to send
+    function mint(address to, uint256 amount) external override onlyGovernorOrAdmin {
+        // first apply rate limit 
+        _depleteBuffer(amount);
+
+        // then mint
+        _mint(to, amount);
+    }
+
+    /// @notice add an address to the lockedTribe excluded list
+    function addLockedTribeAddress(address lockedTribeAddress) external override onlyGovernorOrAdmin {
+        _addLockedTribeAddress(lockedTribeAddress);
+    }
+
+    /// @notice remove an address from the lockedTribe excluded list
+    function removeLockedTribeAddress(address lockedTribeAddress) external onlyGovernorOrAdmin {
+        _lockedTribeAddresses.remove(lockedTribeAddress);
+        emit RemoveLockedTribeAddress(lockedTribeAddress);
+    }
+
+    /// @notice changes the TRIBE minter address
+    /// @param newMinter the new minter address
+    function setMinter(address newMinter) external override onlyOwner {
+        require(newMinter != address(0), "TribeReserveStabilizer: zero address");
+        ITribe _tribe = ITribe(address(tribe()));
+        _tribe.setMinter(newMinter);
+    }
+
+    /// @notice sets the max annual inflation relative to current supply
+    /// @param newAnnualMaxInflationBasisPoints the new max inflation % denominated in basis points (1/10000)
+    function setAnnualMaxInflationBasisPoints(uint256 newAnnualMaxInflationBasisPoints) external override onlyGovernorOrAdmin {
+        _setAnnualMaxInflationBasisPoints(newAnnualMaxInflationBasisPoints);
+    }
+
+    /// @notice return the ideal buffer cap based on TRIBE circulating supply
+    function idealBufferCap() public view override returns (uint256) {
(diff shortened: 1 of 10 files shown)


contracts/utils/RateLimited.sol
```diff
@@ -103,9 +106,13 @@ abstract contract RateLimited is CoreRef {
...
         // Cap the existing stored buffer
         if (_bufferStored > newBufferCap) {
-            _bufferStored = newBufferCap;
+            _resetBuffer();
         }
 
         emit BufferCapUpdate(oldBufferCap, newBufferCap);
     }
+
+    function _resetBuffer() internal {
```
Joeysantoro, line 115: Added so TribeMinter constructor can start with a full buffer

contracts/stabilizer/TribeReserveStabilizer.sol
```diff
@@ -99,41 +92,8 @@ contract TribeReserveStabilizer is ITribeReserveStabilizer, ReserveStabilizer, R
...
-        }
-        if (mintAmount != 0) {
-            _mint(to, mintAmount);
-        }
-    }
-
-    function _mint(address to, uint256 amount) internal {
-        ITribe _tribe = ITribe(address(token));
-        _tribe.mint(to, amount);
+        tribeMinter.mint(to, amount);
```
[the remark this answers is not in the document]
Joeysantoro: The naming is reasonable but I'm open to suggestions. The TRIBE held in this contract is effectively burned so I treat this operation like minting

contracts/dao/TribeMinter.sol
```diff
@@ -0,0 +1,211 @@
...
+    function setBufferCap(uint256) external pure override { 
+        revert("no-op");
+    }
+
+    /// @notice mints TRIBE to the target address, subject to rate limit
+    /// @param to the address to send TRIBE to
+    /// @param amount the amount of TRIBE to send
+    function mint(address to, uint256 amount) external override onlyGovernorOrAdmin {
+        // first apply rate limit 
+        _depleteBuffer(amount);
```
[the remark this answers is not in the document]
Joeysantoro: Not a logic error, the rate limit applies to transferred and minted TRIBE

contracts/dao/TribeMinter.sol
```diff
@@ -0,0 +1,211 @@
...
+        newBufferCap = idealBufferCap();
+        oldBufferCap = bufferCap;
+        require(newBufferCap != oldBufferCap, "TribeMinter: No rate limit change needed");
+        
+        _setBufferCap(newBufferCap);
+        _setRateLimitPerSecond(newBufferCap / Constants.ONE_YEAR);
+    }
+
+    // Transfer held TRIBE first, then mint to cover remainder
+    function _mint(address to, uint256 amount) internal {
```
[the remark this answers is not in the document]
Joeysantoro: I'm down to break this out into multiple methods, however I think the general functionality is pretty much as expected

test/unit/stablizer/TribeReserveStabilizer.test.ts
```diff
@@ -209,37 +159,6 @@ describe('TribeReserveStabilizer', function () {
     });
```
Joeysantoro, line 159: @ElliotFriedman this test here checks Withdraw

contracts/dao/TribeMinter.sol
```diff
@@ -0,0 +1,211 @@
...
+        _setAnnualMaxInflationBasisPoints(newAnnualMaxInflationBasisPoints);
+    }
+
+    /// @notice return the ideal buffer cap based on TRIBE circulating supply
+    function idealBufferCap() public view override returns (uint256) {
+        return tribeCirculatingSupply() * annualMaxInflationBasisPoints / Constants.BASIS_POINTS_GRANULARITY;
+    }
+
+    /// @notice return the TRIBE supply, subtracting locked TRIBE
+    function tribeCirculatingSupply() public view override returns (uint256) {
```
ditchfieldcaleb, line 134: @Joeysantoro I think your intention for this function is to use it for external orgs that want to get our circulating supply; if this is the case, we should discuss in person imo since I think we'll experience a prohibitively high amount of resistance to this (ie we'll have a hard time convincing important orgs or people to use this instead of the Tribe totalSupply() method

contracts/dao/TribeMinter.sol
```diff
@@ -0,0 +1,211 @@
...
+        IERC20 _tribe = tribe();
+
+        // Remove all locked TRIBE from total supply calculation
+        uint256 lockedTribe;
+        for (uint256 i = 0; i < _lockedTribeAddresses.length(); i++) {
```
ditchfieldcaleb, line 139: Using a potentially unlimited-gas-cap loop with no way to abort and still not throw in the function makes me very uncomfortable for a mission-critical function like this

also this will be very gas expensive if this function is called frequently by users

contracts/dao/TribeMinter.sol
ditchfieldcaleb: one way to mitigate this while still retaining the functionality therein is to split this out into three functions: one that reports the raw supply (without the locked addresses), another that reports the locked supply, and a final third that reports the difference.

contracts/dao/TribeMinter.sol
```diff
@@ -0,0 +1,211 @@
...
+
+    /// @notice alias for tribeCirculatingSupply
+    /// @dev for compatibility with ERC-20 standard for off-chain 3rd party sites
+    function totalSupply() public view override returns (uint256) {
+        return tribeCirculatingSupply();
+    }
+
+    /// @notice return whether a poke is needed or not i.e. is buffer cap != ideal cap
+    function isPokeNeeded() external view override returns (bool) {
+        return idealBufferCap() != bufferCap;
```
ditchfieldcaleb, line 154: we should use a percent range to the buffer cap here.
imagine a scenario where the buffer is 1 wei below the cap - it might not even be possible to hit the cap exactly depending on the precise math to mint 1 wei

and even if it is, it would have to be done via a bot or manually.

contracts/dao/TribeMinter.sol
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    /// @notice return the set of locked TRIBE holding addresses to be excluded from circulating supply
+    function lockedTribeAddresses() external view override returns(address[] memory) {
+        return _lockedTribeAddresses.values();
+    }
+
+    function _addLockedTribeAddress(address lockedTribeAddress) internal {
+        _lockedTribeAddresses.add(lockedTribeAddress);
+        emit AddLockedTribeAddress(lockedTribeAddress);
```
ditchfieldcaleb, line 164: only emit an event if .add returns true; else the address was already in the enumerableSet

contracts/utils/RateLimited.sol
```diff
@@ -26,6 +26,7 @@ abstract contract RateLimited is CoreRef {
     /// @notice the buffer at the timestamp of lastBufferUsedTime
     uint256 private _bufferStored;
 
+    event BufferUsed(uint256 amountUsed, uint256 bufferRemaining);
```
ditchfieldcaleb, line 29: (this comment isn't about this particular line but the whole file)

I strongly think we should extend RateLimited instead of changing it.
At the very least, these changes should go into RateLimitedV2.

But a good pattern here would be:

`contract RateLimitedWithBuffer extends RateLimit`

Adding internal functions especially to existing-deployed contracts that we have out on mainnet is a good recipe for interface-mismatch, and while we might catch it in integration tests, we should always try to keep up as many layers of [swiss-cheese security](https://en.wikipedia.org/wiki/Swiss_cheese_model) as we can.

contracts/dao/TribeMinter.sol
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    /// @notice return the TRIBE supply, subtracting locked TRIBE
+    function tribeCirculatingSupply() public view override returns (uint256) {
+        IERC20 _tribe = tribe();
+
+        // Remove all locked TRIBE from total supply calculation
+        uint256 lockedTribe;
+        for (uint256 i = 0; i < _lockedTribeAddresses.length(); i++) {
+            lockedTribe += _tribe.balanceOf(_lockedTribeAddresses.at(i));
```
ditchfieldcaleb, line 140: Using the literal tribe balance of the locked address adds in a potential current (and if not current, potential future) exploit: atomically flashloaning a TON of Tribe into a locked-tribe-address, calling this function, and using the atomically-reduced-circulating-tribe-supply to inflate greater than intended.

you might say "if the tribe goes to the locked address it's locked!" - but that is not necessarily the case. if we keep the implementation this way we will need to carefully consider every single locked address we add to ensure that there isn't any functionality for recovering tokens accidentally sent to that address, and we'll need to remember this for every future locked address we add

also - consider the scenario in which tokens vest from a timelock. those addresses contain *some* locked and *some* circulating tribe, and allow the owners of the vested-portion to manipulate the Tribe inflation rate this way if they wish. this seems pretty problematic.

contracts/utils/RateLimited.sol
```diff
@@ -43,13 +44,13 @@ abstract contract RateLimited is CoreRef {
...
-    function setRateLimitPerSecond(uint256 newRateLimitPerSecond) external onlyGovernorOrAdmin {
+    function setRateLimitPerSecond(uint256 newRateLimitPerSecond) external virtual onlyGovernorOrAdmin {
         require(newRateLimitPerSecond <= MAX_RATE_LIMIT_PER_SECOND, "RateLimited: rateLimitPerSecond too high");
         _setRateLimitPerSecond(newRateLimitPerSecond);
     }
 
     /// @notice set the buffer cap
-    function setbufferCap(uint256 newBufferCap) external onlyGovernorOrAdmin {
+    function setBufferCap(uint256 newBufferCap) external virtual onlyGovernorOrAdmin {
         _setBufferCap(newBufferCap);
```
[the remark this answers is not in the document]
ditchfieldcaleb: *if not already in the internal method

contracts/dao/TribeMinter.sol
Joeysantoro: no longer using locked addresses

contracts/dao/TribeMinter.sol
Joeysantoro: no longer using locked addresses

contracts/utils/RateLimited.sol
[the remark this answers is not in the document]
Joeysantoro: already in internal

contracts/dao/TribeMinter.sol
Joeysantoro: I'm not sure this is needed. Poke is not incentivized, so theres no griefing or gaming I'm really worried about. If we add an incentive higher up we can add this logic but doesn't belong here imo

ditchfieldcaleb: Yeah I misunderstood the purpose of this function. All good.

ditchfieldcaleb (approved):
This looks great. I'm glad we hardcoded in the Treasury, Dripper, etc.

I think we need to make very sure that the Tribe balance in the Treasury & the Dripper aren't manipulable if we're going to use tribe.balanceOf (as opposed to having them report their own locked balances).

Otherwise, no concerns.

PR #234, opened 2021-10-11
Read pause override

Joeysantoro:
Add the ability for the guardian to enforce reads are valid even while paused.

This is useful when the guardian is manually adjusting the cache values via CollateralizationOracleGuardian.


ElliotFriedman (approved):
LGTM!

PR #237, opened 2021-10-11
fuse pool deploy

Joeysantoro:
Adds a reusable fuse pool deploy script with 3 new pools:
- note pool 54 is not verified yet


scripts/deploy/compoundPCVDeposit.ts
```diff
@@ -0,0 +1,24 @@
+import { ethers } from 'hardhat';
+
+export async function deploy(deployAddress, addresses, logging = false) {
```
ditchfieldcaleb, line 3: type as deployFunc

ElliotFriedman (approved):
LGTM!

PR #241, opened 2021-10-12
moving files

Joeysantoro:
Also deletes useless deploy scripts, the balancer one is covered by https://github.com/fei-protocol/fei-protocol-core/pull/231


proposals/dao/old/feiRari.ts
```diff
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import { RunUpgradeFunc, SetupUpgradeFunc, TeardownUpgradeFunc, ValidateUpgradeFunc } from '../../types/types';
+import { RunUpgradeFunc, SetupUpgradeFunc, TeardownUpgradeFunc, ValidateUpgradeFunc } from '../../../types/types';
```
ditchfieldcaleb, line 2: If we're going to change paths we may as well switch to the @ syntax, which will likelyprevent the need for future path changes

Joeysantoro: done

PR #245, opened 2021-10-13
CR oracle deploy

Joeysantoro:


scripts/deploy/v2Phase1-part2.ts
```diff
@@ -0,0 +1,381 @@
...
+  const creamFeiPCVDepositWrapperFactory = await ethers.getContractFactory('PCVDepositWrapper');
+  const creamFeiPCVDepositWrapper = await creamFeiPCVDepositWrapperFactory.deploy(creamFeiPCVDeposit, fei, true);
+
+  logging && console.log('creamFeiPCVDepositWrapper: ', creamFeiPCVDepositWrapper.address);
+
+  const staticPcvDepositWrapperFactory = await ethers.getContractFactory('StaticPCVDepositWrapper');
+  const staticPcvDepositWrapper = await staticPcvDepositWrapperFactory.deploy(
+    core,
+    toBN(4000000).mul(ethers.constants.WeiPerEther).toString(), // 4M PCV for 100k INDEX @ ~$40
+    toBN(8500000).mul(ethers.constants.WeiPerEther).toString() // 8.5M FEI in Kashi
```
ditchfieldcaleb, line 292: Excellent, excellent. Strong math skills here.

ditchfieldcaleb (changes requested):
Overall it looks great and I don't see any bugs. However, a couple of requests before we go live:

1) @Joeysantoro you're more familiar with which pcv deposits are protocol-owned (which would make the second parameter True). You should definitely check over all of the ones here to make sure this parameter is set correctly for each one.

2) In the "Deploy Actions" comment, there are 28 actions. Can you number each deployment action in the script corresponding to this comment? This will ensure we aren't missing anything.

3) You should do one more check over to make sure there's no PCV deposits we forgot to add.

scripts/deploy/v2Phase1-part2.ts
```diff
@@ -0,0 +1,381 @@
+import { ethers } from 'hardhat';
+import { getAllContractAddresses } from '@test/integration/setup/loadContracts';
+import { DeployUpgradeFunc, NamedContracts } from '@custom-types/types';
+
+const toBN = ethers.BigNumber.from;
+
+const USD_ADDRESS = '0x1111111111111111111111111111111111111111';
```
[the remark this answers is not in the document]
Joeysantoro: Yes, its the constant used in repo to represent USD, abstractly

Joeysantoro:
> 1. @Joeysantoro you're more familiar with which pcv deposits are protocol-owned (which would make the second deploy parameter True). You should definitely check over all of the ones here to make sure this parameter is set correctly for each one.

I had them organized into two groups so it should be false in the first and true in the second



> 2\. In the "Deploy Actions" comment, there are 28 actions. Can you number each deployment action in the script corresponding to this comment? This will ensure we aren't missing anything.

Done

> 3\. You should do one more check over to make sure there's no PCV deposits we forgot to add.

Can't think of any more after we add ERC20 one. I also incremented the manual PC FEI for Idle and BarnBridge. I made the guardian the owner until we ratify on-chain so we can add/remove

contracts/oracle/collateralization/CollateralizationOracle.sol
```diff
@@ -75,12 +76,7 @@ contract CollateralizationOracle is ICollateralizationOracle, CoreRef {
 
     /// @notice returns an array of the addresses of tokens held in the pcv.
     function getTokensInPcv() external view returns(address[] memory) {
-        uint256 _length = tokensInPcv.length();
-        address[] memory tokens = new address[](_length);
-        for (uint256 i = 0; i < _length; i++) {
-            tokens[i] = tokensInPcv.at(i);
-        }
-        return tokens;
+        return tokensInPcv.values();
```
ditchfieldcaleb, line 79: WHAT I didn't know this was a thing omg

ditchfieldcaleb: Ah it's via enumerableSet. Very nice.

proposals/dao/collateralizationOracle.ts
```diff
@@ -0,0 +1,486 @@
...
+  const guardian = addresses.multisig;
+
+  const { dai, weth, dpi, rai, fei} = addresses;
+
+  let staticBalances = await staticPcvDepositWrapper.resistantBalanceAndFei();
+  expect(staticBalances[0]).to.be.equal(ethers.constants.WeiPerEther.mul(4_000_000));
+  expect(staticBalances[1]).to.be.equal(ethers.constants.WeiPerEther.mul(11_500_000));
+
+  const tokens = await collateralizationOracle.getTokensInPcv();
+  expect(tokens.length).to.be.equal(6);
```
ditchfieldcaleb, line 461: I dislike the fact that this reports 6 tokens even though there's only actually five - the "USD" fake-token shouldn't count imo. @Joeysantoro is there any reason not to have this function report -1 on the "total"?

ditchfieldcaleb:
linting & prettier for you, and also pushed the small fix for the typescript-casting for the contracts.

proposals/dao/collateralizationOracle.ts
Joeysantoro: It returns a list so no, we shouldn't exclude USD. We want to know how many deposits are abstractly USD

PR #246, opened 2021-10-14
V2 Interfaces

Joeysantoro:


contracts/pcv/IPCVDepositAggregator.sol
```diff
@@ -0,0 +1,48 @@
...
+    function pcvDeposits() external view returns(IPCVDeposit[] memory deposits, uint256[] memory weights);
+
+    /// @notice current percent of PCV held by the input `pcvDeposit` relative to the total managed by aggregator.
+    /// @param depositAmount a hypothetical deposit amount, to be included in the calculation
+    function percentHeld(IPCVDeposit pcvDeposit, uint256 depositAmount) external view returns(Decimal.D256 memory);
+
+    /// @notice the normalized target percent of PCV held by `pcvDeposit` relative to aggregator total
+    function targetPercentHeld(IPCVDeposit pcvDeposit) external view returns(Decimal.D256 memory);
+
+    /// @notice the raw amount of PCV off of the target percent held by `pcvDeposit`
```
ditchfieldcaleb, line 46: I don't quite understand what you mean by this

contracts/stabilizer/IPegStabilityModule.sol
```diff
@@ -0,0 +1,91 @@
...
+
+    /// @notice mint `amountFeiOut` FEI to address `to` for `amountIn` underlying tokens
+    /// @dev see getMintAmountOut() to pre-calculate amount out
+    function mint(address to, uint256 amountIn)
+        external
+        returns (uint256 amountFeiOut);
+
+    /// @notice redeem `amountFeiIn` FEI for `amountOut` underlying tokens and send to address `to` 
+    /// @dev see getRedeemAmountOut() to pre-calculate amount out
+    function redeem(address to, uint256 amountFeiIn)
```
ElliotFriedman, line 36: Same comment here. There should be a third param that allows users to specify their output currency so that they can receive assets of their choosing.  @Joeysantoro Please confirm that a single PSM contract will be responsible for handling all assets.

contracts/stabilizer/IPegStabilityModule.sol
```diff
@@ -0,0 +1,91 @@
...
+        external
+        returns (uint256 amountOut);
+
+    /// @notice send any surplus reserves to the PCV allocation
+    function allocateSurplus() external;
```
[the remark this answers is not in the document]
Joeysantoro: should be public and incentivized by a keeper contract

contracts/pcv/IPCVDepositAggregator.sol
Joeysantoro: i.e if its 900 vs 1000 threhsold, the % is 10 and the raw amount is 100

PR #252, opened 2021-10-17
tribal chief sync

Joeysantoro:
This contract is a utility to atomically mass update all pools and sync the auto rewards distributor when changing tribalChief.

It executes directly on the Timelock


test/integration/setup/index.ts
```diff
@@ -50,6 +51,7 @@ export class TestEndtoEndCoordinator implements TestCoordinator {
    *
    */
   public async loadEnvironment(): Promise<Env> {
+    await resetFork();
```
ditchfieldcaleb, line 54: nice

ditchfieldcaleb (dismissed):
Looks good, see my two comments tho

contracts/staking/TribalChiefSync.sol
```diff
@@ -0,0 +1,98 @@
...
+            address(tribalChief), 
+            0, 
+            data, 
+            bytes32(0), 
+            salt
+        );
+    }
+
+    /// @notice Sync a pool addition
+    function addPool(
```
ElliotFriedman, line 76: We should add a function to this contract that does the same thing to `resetRewards` and `set`. Those functions only update their pool, but do not account for other pools having their tribePerBlock changing.

ditchfieldcaleb (dismissed):

PR #253, opened 2021-10-18
remove burn req from stabilizer

Joeysantoro:


ElliotFriedman (dismissed):

test/integration/tests/buybacks.ts
```diff
@@ -15,7 +15,7 @@ before(async () => {
   await resetFork();
 });
 
-describe.only('e2e', function () {
+describe('e2e', function () {
```
ditchfieldcaleb, line 18: should be e2e-buybacks

ditchfieldcaleb (changes requested):
One small change, plus rename to ReserveStabilizerV2

PR #258, opened 2021-10-19
Feat/pcv deposit aggregator

ditchfieldcaleb:
PCV Deposit Aggregator. 

Each PCV Deposit Aggregator sits below its parent RewardsAssetManager, and is in charge of allocating to and from individual PCV Deposits for its particular currency. This could be ETH or tokens.

The interface (IPCVDepositAggregator) can be found here

Each PCV Deposit controlled by each aggregator will have a relative "weight", similar to how TribalChief works. When funds are needed to be withdrawn from this particular-currency asset, the RewardsAggregator that sits atop each aggregator will call into the aggregator with withdraw(). For the inverse, deposit()

The PCV Deposit Aggregator will need to intelligently withdraw, deposit, and occasionally rebalance funds so that the target weights remain generally correct (though this cannot be a strict requirement due to fluctuations in underlying value and yield)

Todo:

- [x] write base contract
- [x] fix draft comment issues
- [x] write unit tests
- [x] gas golf to heck and back
- [x] natspec
- [x] add/fix role
- [x] sum of optimal balances vs underlying sum (waiting on Joey response)
- [x] add guardian-only setDepositWeightZero function
- [x] final review


contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+    function rebalance() external virtual override {
+        _rebalance();
+    }
+
+    function rebalanceSingle(address pcvDeposit) external virtual override {
+        _rebalanceSingle(pcvDeposit);
+    }
+
+    function deposit() external virtual override {
```
Joeysantoro, line 73: This shouldn't be a no-op right? It should rebalance if above buffer or something

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+        }
+    }
+
+    function withdrawERC20(address _token, address to, uint256 amount) external virtual override onlyPCVController {
+        IERC20(_token).safeTransfer(to, amount);
+    }
+
+    function withdrawETH(address payable to, uint256 amount) external virtual override onlyPCVController {
+        to.transfer(amount);
+    }
```
Joeysantoro, line 136: These are in PCVDeposit if inherited

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+    function setBufferWeight(uint128 newBufferWeight) external virtual override onlyGuardianOrGovernor {
+        int128 difference = int128(newBufferWeight) - int128(bufferWeight);
+        bufferWeight = uint128(int128(bufferWeight) + difference);
+
+        totalWeight = uint128(int128(totalWeight) + difference);
```
Joeysantoro, line 142: add event

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+    function withdrawERC20(address _token, address to, uint256 amount) external virtual override onlyPCVController {
+        IERC20(_token).safeTransfer(to, amount);
+    }
+
+    function withdrawETH(address payable to, uint256 amount) external virtual override onlyPCVController {
+        to.transfer(amount);
+    }
+
+    function setBufferWeight(uint128 newBufferWeight) external virtual override onlyGuardianOrGovernor {
```
Joeysantoro, line 138: guardian should not be able to set weights, should be admin

contracts/pcv/PCVDepositAggregator.sol
Joeysantoro, line 142: Use safeCast

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+        int128 difference = int128(newBufferWeight) - int128(bufferWeight);
+        bufferWeight = uint128(int128(bufferWeight) + difference);
+
+        totalWeight = uint128(int128(totalWeight) + difference);
+    }
+
+    function setPCVDepositWeight(address depositAddress, uint128 newDepositWeight) external virtual override onlyGuardianOrGovernor {
```
Joeysantoro, line 145: Should be admin or guardian

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+        if (!pcvDepositAddresses.contains(depositAddress)) {
+            revert("Deposit does not exist.");
+        }
```
Joeysantoro, line 148: Why not add instead?

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+        int128 difference = int128(newDepositWeight) - int128(pcvDepositInfos[depositAddress].weight);
+        pcvDepositInfos[depositAddress].weight = uint128(newDepositWeight);
+
+        totalWeight = uint128(int128(totalWeight) + difference);
+    }
+
+    function removePCVDeposit(address pcvDeposit) external virtual override onlyGuardianOrGovernor {
```
Joeysantoro, line 156: admin not guardian

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+        _removePCVDeposit(address(pcvDeposit));
+    }
+
+    function addPCVDeposit(address newPCVDeposit, uint256 weight) external virtual override onlyGovernor {
```
Joeysantoro, line 160: admin can add too

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+    function balance() public view virtual override returns (uint256) {
+        return getTotalBalance();
+    }
+
+    function percentHeld(address pcvDeposit, uint256 depositAmount) external view virtual override returns(Decimal.D256 memory) {
+        uint totalBalanceWithTheoreticalDeposit = getTotalBalance() + depositAmount;
+        uint targetBalanceWithTheoreticalDeposit = IPCVDeposit(pcvDeposit).balance() + depositAmount;
+
+        return Decimal.ratio(targetBalanceWithTheoreticalDeposit, totalBalanceWithTheoreticalDeposit);
```
Joeysantoro, line 215: Whats the difference between theoretical and hypothetical

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    function targetPercentHeld(address pcvDeposit) external view virtual override returns(Decimal.D256 memory) {
```
Joeysantoro, line 218: Target should use the weights not balances

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+    function amountFromTarget(address pcvDeposit) external view virtual override returns(int256) {
+        uint totalBalance = getTotalBalance();
+
+        uint pcvDepositBalance = IPCVDeposit(pcvDeposit).balance();
+        uint pcvDepositWeight = pcvDepositInfos[address(pcvDeposit)].weight;
+
+        uint idealDepositBalance = pcvDepositWeight * totalBalance / totalWeight;
+        
+        return int(idealDepositBalance) - int(pcvDepositBalance);
```
Joeysantoro, line 233: Thoughts on whether positive should mean above or below?

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+            address pcvDepositAddress = pcvDepositAddresses.at(i);
+            _removePCVDeposit(pcvDepositAddress);
+        }
+
+        // Finally, set the new aggregator on the rewards asset manager itself
+        IRewardsAssetManager(rewardsAssetManager).setNewAggregator(address(newAggregator));
+    }
+
+    // ---------- View Functions ---------------
+    function resistantBalanceAndFei() external view virtual override returns (uint256, uint256) {
```
Joeysantoro, line 199: This should probably sum resistant balance and FEI of all underlying

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+        pcvDepositAddresses.add(depositAddress);
+        pcvDepositInfos[depositAddress] = PCVDepositInfo(
+            weight,
+            false
+        );
+
+        totalWeight = totalWeight + weight;
+    }
+    function _removePCVDeposit(address depositAddress) internal {
```
Joeysantoro, line 344: Should this method also withdraw?

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+            IERC20(token).safeTransfer(depositAddress, amountNeeded);
+            IPCVDeposit(depositAddress).deposit();
+        }
+    }
+
+    function _addPCVDeposit(address depositAddress, uint128 weight) internal {
+        if (pcvDepositAddresses.contains(depositAddress)) {
+            revert("Deposit already added.");
+        }
```
Joeysantoro, line 334: Why not require instead of if-revert?

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+        return totalBalance;
+    }
+
+    // ---------- Internal Functions ----------- //
+
+    function _rebalanceSingle(address pcvDeposit) internal {
+        if (!pcvDepositAddresses.contains(pcvDeposit)) {
+            revert("Deposit does not exist.");
+        }
```
Joeysantoro, line 266: why not require

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+        // Let's not forget to get this balance
+        totalBalance += IERC20(token).balanceOf(address(this));
+
+        return totalBalance;
+    }
+
+    // ---------- Internal Functions ----------- //
+
+    function _rebalanceSingle(address pcvDeposit) internal {
```
Joeysantoro, line 263: Call amountFromTarget as a subroutine to simplify the logic here

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+            } else {
+                // Emit event so that we know to do a full rebalance
+            }
+        } else {
+            // PCV deposit balance is exactly where it needs to be. Don't touch it.
+        }
+    }
+
+
+    function _rebalance() internal {
```
Joeysantoro, line 292: We should try to improve the algorithm here. Shouldn't do more than 2 external calls to each PCV deposit, one for balance, and one for withdraw/deposit

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+            IPCVDepositAggregator(newAggregator).addPCVDeposit(pcvDepositAddress, pcvDepositWeight);
+        }
+
+        // Set all weights to zero (except for us)
+        for (uint i=0; i < pcvDepositAddresses.length(); i++) {
+            address pcvDepositAddress = pcvDepositAddresses.at(i);
+            pcvDepositInfos[pcvDepositAddress].weight = uint128(0);
+        }
+
+        // Rebalance (withdraws everything to us)
```
Joeysantoro, line 179: No need to rebalance imo, because all the deposits should have similar weights in new aggregator, let that one handle rebalance

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+        IERC20(token).safeTransfer(address(newAggregator), IERC20(token).balanceOf(address(this)));
+
+        // Call rebalance on the new aggregator
+        IPCVDepositAggregator(newAggregator).rebalance();
+
+        // Remove all deposits.
+        for (uint i=0; i < pcvDepositAddresses.length(); i++) {
+            address pcvDepositAddress = pcvDepositAddresses.at(i);
+            _removePCVDeposit(pcvDepositAddress);
+        }
```
Joeysantoro, line 192: Also probably no need to reset all this state

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    function setNewAggregator(address newAggregator) external virtual override onlyGovernor {
+        // Add each pcvDeposit to the new aggregator
+        for (uint i=0; i < pcvDepositAddresses.length(); i++) {
+            address pcvDepositAddress = pcvDepositAddresses.at(i);
+            uint128 pcvDepositWeight = pcvDepositInfos[pcvDepositAddress].weight;
+
+            IPCVDepositAggregator(newAggregator).addPCVDeposit(pcvDepositAddress, pcvDepositWeight);
+        }
```
Joeysantoro, line 171: If the other PCVDeposit aggregator reverts, should check existence before adding

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+        // Set all weights to zero (except for us)
+        for (uint i=0; i < pcvDepositAddresses.length(); i++) {
+            address pcvDepositAddress = pcvDepositAddresses.at(i);
+            pcvDepositInfos[pcvDepositAddress].weight = uint128(0);
```
Joeysantoro, line 176: No need to reset state imo

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+     * 1. check if the contract has enough in the buffer to cover the withdrawal. if so, just use this
+     * 2. if it doesn't, scan through each of the pcv deposits and withdraw from them their overage amounts,
+     *    up to the total amount needed (less the amount already in the buffer)
+     */
+    function withdraw(address to, uint256 amount) external virtual override onlyPCVController {
+        uint totalBalance = getTotalBalance();
+
+        if (amount > totalBalance) {
+            revert("Not enough balance to withdraw");
+        }
```
Joeysantoro, line 89: require instead of if-revert

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+        // If our buffer is full enough, just transfer from that
+        if (IERC20(token).balanceOf(address(this)) >= amount) {
+            IERC20(token).safeTransfer(to, amount);
+        } else {
+            // We're going to have to pull from underlying deposits
```
Joeysantoro, line 95: What about if we do all the overages and thats not enough? This method should not fail under any circumstance if possible

Joeysantoro (changes requested):

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: Separation of responsibilities. setPCVDepositWeight should do one thing and one thing only: set an (existing) pcv deposit's weight.

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: Nothing? We could use either term.

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+        _setContractAdminRole(keccak256("PCV_CONTROLLER_ROLE"));
+
+        for (uint i=0; i < _initialPCVDepositAddresses.length; i++) {
+            _addPCVDeposit(_initialPCVDepositAddresses[i], _initialPCVDepositWeights[i]);
+        }
+    }
+
+    // ---------- Public Functions -------------
+
+    function rebalance() external virtual override {
```
ElliotFriedman, line 65: This should be pausable.

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: Since we're using the nomenclature "amount [off] from target", a positive number should mean that we have overshot, eg we should take from this deposit. Likewise a negative number means that we are undershot and this deposit needs more.

...which makes my implementation backwards, but happy to correct.

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+            address depositAddress = pcvDepositAddresses.at(i);
+
+            IERC20(token).safeTransfer(depositAddress, amountNeeded);
+            IPCVDeposit(depositAddress).deposit();
+        }
+    }
+
+    function _addPCVDeposit(address depositAddress, uint128 weight) internal {
+        if (pcvDepositAddresses.contains(depositAddress)) {
```
ElliotFriedman, line 332: You should call pcvDepositInfos[depositAddress] and seeing if that deposit has a weight instead of looping over the pcvDepositAddresses to see if there is a match.

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: It does. Well, it does for balance. Should do it for FEI too.

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+                // Needs a deposit. Let's write that down.
+                uint needed = idealDepositBalance - pcvDepositBalance;
+                depositAmountsNeeded[i] = needed;
+            } else {
+                // Do nothing
+                // This accounts for the rare = case, but is definitely necessary.
+            }
+        }
+
+        for (uint i=0; i<depositAmountsNeeded.length; i++) {
```
ElliotFriedman, line 320: Please add spacing like so:

for (uint i = 0; i < depositAmountsNeeded.length; i++)

to make things visually cleaner.

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: This is currently the case, save for the ones we deposit into, where we call deposit() after transferring.

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: Yep, you found a bug. Fixing.

ditchfieldcaleb: Fixed.

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+
+        for (uint i=0; i < pcvDepositAddresses.length(); i++) {
+            _deposits[i] = pcvDepositAddresses.at(i);
+            _weights[i] = pcvDepositInfos[pcvDepositAddresses.at(i)].weight;
+        }
+
+        return (_deposits, _weights);
+    }
+
+    function getTotalBalance() public view virtual override returns (uint256) {
```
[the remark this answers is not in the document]
ditchfieldcaleb: Agreed, but do we want to have a method that reports *just* this contract's balance?

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: Caching implemented.

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,365 @@
...
+import "./IPCVDepositAggregator.sol";
+import "./PCVDeposit.sol";
+import "../refs/CoreRef.sol";
+import "../external/Decimal.sol";
+import "./balancer/IRewardsAssetManager.sol";
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+import "@openzeppelin/contracts/utils/structs/EnumerableMap.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+contract PCVDepositAggregator is IPCVDepositAggregator, CoreRef {
```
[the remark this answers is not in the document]
ditchfieldcaleb: @todo remove IPCVDeposit from AggeregatorInterface

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: Nevermind

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: I can see reasons why it might want to, and reasons why it might not want to.

If we're transferring a pcv deposit to another aggregator or another pcv controller role type entirely, it might not make sense to withdraw.

If we're just removing the pcv deposit because it's not being used anymore, it definitely should withdraw.

Either way to preserve losslessness it needs to withdraw upon removal, though.

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: Took out uint128/int128, no longer needed.

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,456 @@
...
+            if (IPCVDepositAggregator(newAggregator).hasPCVDeposit(pcvDepositAddresses.at(i))) continue;
+
+            address pcvDepositAddress = pcvDepositAddresses.at(i);
+            uint pcvDepositWeight = pcvDepositWeights[pcvDepositAddress];
+
+            IPCVDepositAggregator(newAggregator).addPCVDeposit(pcvDepositAddress, pcvDepositWeight);
+        }
+
+        // Send old aggregator assets over to the new aggregator
+        IERC20(token).safeTransfer(address(newAggregator), IERC20(token).balanceOf(address(this)));
```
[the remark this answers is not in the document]
ditchfieldcaleb: done

.eslintcache
```diff
@@ -0,0 +1 @@
...83"},{"messageId":"2078","fix":"2084","desc":"2080"},{"messageId":"2081","fix":"2085","desc":"2083"},{"messageId":"2078","fix":"2086","desc":"2080"},{"messageId":"2081","fix":"2087","desc":"2083"},{"messageId":"2078","fix":"2088","desc":"2080"},{"messageId":"2081","fix":"2089","desc":"2083"},{"messageId":"2078","fix":"2090","desc":"2080"},{"messageId":"2081","fix":"2091","desc":"2083"},{"messageId":"2078","fix":"2092","desc":"2080"},{"messageId":"2081","fix":"2093","desc":"2083"},"suggestUnknown",{"range":"2094","text":"2095"},"Use `unknown` instead, this will force you to explicitly, and safely assert the type is correct.","suggestNever",{"range":"2094","text":"2096"},"Use `never` instead, this is useful when instantiating generic type parameters that you don't need to know the type of.",{"range":"2097","text":"2095"},{"range":"2097","text":"2096"},{"range":"2098","text":"2095"},{"range":"2098","text":"2096"},{"range":"2099","text":"2095"},{"range":"2099","text":"2096"},{"range":"2100","text":"2095"},{"range":"2100","text":"2096"},{"range":"2101","text":"2095"},{"range":"2101","text":"2096"},[2069,2072],"unknown","never",[1122,1125],[1834,1837],[1873,1876],[1959,1962],[4612,4615]]
```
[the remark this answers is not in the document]
ditchfieldcaleb: you're right, will add it to the gitignore

contracts/pcv/IPCVDepositAggregator.sol
```diff
@@ -11,38 +11,70 @@ import "../external/Decimal.sol";
...
     function rebalance() external;
 
+    /// @notice same as the rebalance function, but for a single deposit
+    function rebalanceSingle(address pcvDeposit) external;
+
     // ----------- Governor only state changing api -----------
     /// @notice adds a new PCV Deposit to the set of deposits
     /// @param weight a relative (i.e. not normalized) weight of this PCV deposit
-    function addPCVDeposit(IPCVDeposit newPCVDeposit, uint256 weight) external;
+    function addPCVDeposit(address newPCVDeposit, uint weight) external;
```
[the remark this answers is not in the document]
ditchfieldcaleb: done

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+
+      // Call deposit on each pcv deposit so that their balances update
+      await pcvDeposit1.deposit();
+      await pcvDeposit2.deposit();
+      await pcvDeposit3.deposit();
+
+      await pcvDepositAggregator
+        .connect(impersonatedSigners[pcvControllerAddress])
+        .withdraw(userAddress, ethers.utils.parseEther('10000'));
+      expect(await token.balanceOf(userAddress)).to.equal(ethers.utils.parseEther('10000'));
```
Joeysantoro, line 441: check balance of aggregator and individuals after as well

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+      expect(await pcvDepositAggregator.totalWeight()).to.equal(130);
+    });
+
+    it('sets the buffer weight and updates the totalweight accordingly', async () => {
+      await pcvDepositAggregator.connect(impersonatedSigners[governorAddress]).setBufferWeight(50);
+      expect(await pcvDepositAggregator.totalWeight()).to.equal(140);
+    });
+
+    it('withdraws when the buffer is not enough to cover the balasnce', async () => {
+      // Mint 4000, 4000, and 4000 tokens to each pcv deposit, respectively
```
Joeysantoro, line 411: comment off

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+    });
+
+    it('sets the pcv deposit weight and updates the totalweight accordingly', async () => {
+      await pcvDepositAggregator
+        .connect(impersonatedSigners[governorAddress])
+        .setPCVDepositWeight(pcvDeposit1.address, 50);
+      expect(await pcvDepositAggregator.totalWeight()).to.equal(130);
+    });
+
+    it('sets the buffer weight and updates the totalweight accordingly', async () => {
```
Joeysantoro, line 405: check ACL

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+      const totalResistantBalanceAndFei = await pcvDepositAggregator.getTotalResistantBalanceAndFei();
+
+      const totalResistantBalance = totalResistantBalanceAndFei[0];
+      const totalResistantFei = totalResistantBalanceAndFei[1];
+
+      expect(totalResistantBalance).to.equal(ethers.utils.parseEther('10000'));
+      expect(totalResistantFei).to.equal(0);
+    });
+
+    it('sets the pcv deposit weight and updates the totalweight accordingly', async () => {
```
Joeysantoro, line 398: check ACL

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+        ethers.utils.parseEther('4000')
+      );
+      expect(await pcvDepositAggregator.amountFromTarget(pcvDeposit2.address)).to.equal(ethers.utils.parseEther('0'));
+      expect(await pcvDepositAggregator.amountFromTarget(pcvDeposit3.address)).to.equal(
+        ethers.utils.parseEther('-3000')
+      );
+    });
+
+    it('reports accurate percentHeld', async () => {
+      // Mint 4000, 4000, and 4000 tokens to each pcv deposit, respectively
```
Joeysantoro, line 344: comment off

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+      await pcvDeposit3.deposit();
+
+      // Rebalance
+      await pcvDepositAggregator.rebalance();
+
+      const pcvDeposit1PercentHeld = (
+        await pcvDepositAggregator.percentHeld(pcvDeposit1.address, ethers.utils.parseEther('10000'))
+      ).value;
+
+      expect(ethers.utils.formatUnits(pcvDeposit1PercentHeld)).to.equal('0.6');
```
Joeysantoro, line 361: show expected math in comments

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+      const pcvDeposit2TargetPercentHeld = await pcvDepositAggregator.targetPercentHeld(pcvDeposit2.address);
+      const pcvDeposit3TargetPercentHeld = await pcvDepositAggregator.targetPercentHeld(pcvDeposit3.address);
+
+      expect(ethers.utils.formatUnits(pcvDeposit1TargetPercentHeld.value)).to.equal('0.2');
+      expect(ethers.utils.formatUnits(pcvDeposit2TargetPercentHeld.value)).to.equal('0.3');
+      expect(ethers.utils.formatUnits(pcvDeposit3TargetPercentHeld.value)).to.equal('0.4');
+    });
+
+    it('reports accurate amountFromTarget', async () => {
+      // Mint 4000, 4000, and 4000 tokens to each pcv deposit, respectively
```
Joeysantoro, line 323: comment off

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+      await token.mint(pcvDeposit1.address, ethers.utils.parseEther('6000'));
+      await token.mint(pcvDeposit2.address, ethers.utils.parseEther('3000'));
+      await token.mint(pcvDeposit3.address, ethers.utils.parseEther('1000'));
+
+      // Call deposit on each pcv deposit so that their balances update
+      await pcvDeposit1.deposit();
+      await pcvDeposit2.deposit();
+      await pcvDeposit3.deposit();
+
+      // Amount from target should be -4000, 0, 3000, respectively
```
Joeysantoro, line 333: I think comment is off here

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+      await pcvDepositAggregator.connect(impersonatedSigners[governorAddress]).addPCVDeposit(pcvDeposit4.address, 10);
+      expect(await pcvDepositAggregator.totalWeight()).to.equal(110);
+    });
+
+    it('removes a pcv deposit', async () => {
+      await pcvDepositAggregator.connect(impersonatedSigners[governorAddress]).removePCVDeposit(pcvDeposit1.address);
+      expect(await pcvDepositAggregator.totalWeight()).to.equal(80);
+    });
+
+    it('reports accurate targetPercentHeld', async () => {
```
Joeysantoro, line 299: This test is also testing rebalancing. Would be useful to see this without a rebalance

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+      const mockPCVDepositDeployer = new MockPCVDepositV2__factory(impersonatedSigners[userAddress]);
+
+      const pcvDeposit4 = await mockPCVDepositDeployer.deploy(core.address, token.address, 0, 0);
+      await pcvDeposit4.deployTransaction.wait();
+
+      await pcvDepositAggregator.connect(impersonatedSigners[governorAddress]).addPCVDeposit(pcvDeposit4.address, 10);
+      expect(await pcvDepositAggregator.totalWeight()).to.equal(110);
+    });
+
+    it('removes a pcv deposit', async () => {
```
Joeysantoro, line 294: Check ACL, also check that the deposit is no longer present using hasPCVDeposit

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+      expect(await token.balanceOf(pcvDepositAggregator.address)).to.equal(ethers.utils.parseEther('4000'));
+    });
+
+    it('adds a pcv deposit', async () => {
+      const mockPCVDepositDeployer = new MockPCVDepositV2__factory(impersonatedSigners[userAddress]);
+
+      const pcvDeposit4 = await mockPCVDepositDeployer.deploy(core.address, token.address, 0, 0);
+      await pcvDeposit4.deployTransaction.wait();
+
+      await pcvDepositAggregator.connect(impersonatedSigners[governorAddress]).addPCVDeposit(pcvDeposit4.address, 10);
```
Joeysantoro, line 290: check PCVDeposit is added, also ACL

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+      expect(await token.balanceOf(pcvDeposit1.address)).to.equal(ethers.utils.parseEther('2000'));
+      expect(await token.balanceOf(pcvDeposit2.address)).to.equal(ethers.utils.parseEther('3000'));
+      expect(await token.balanceOf(pcvDeposit3.address)).to.equal(ethers.utils.parseEther('4000'));
+
+      // Also check the aggregator balance
+      expect(await token.balanceOf(pcvDepositAggregator.address)).to.equal(ethers.utils.parseEther('1000'));
+    });
+
+    it('rebalances a single deposit', async () => {
+      // Mint 4000, 4000, and 4000 tokens to each pcv deposit, respectively
```
Joeysantoro, line 265: comment is off

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+      expect(await token.balanceOf(pcvDeposit1.address)).to.equal(ethers.utils.parseEther('2000'));
+      expect(await token.balanceOf(pcvDeposit2.address)).to.equal(ethers.utils.parseEther('3000'));
+      expect(await token.balanceOf(pcvDeposit3.address)).to.equal(ethers.utils.parseEther('4000'));
+
+      // Also check the aggregator balance
+      expect(await token.balanceOf(pcvDepositAggregator.address)).to.equal(ethers.utils.parseEther('1000'));
+    });
+
+    it('successfully rebalances when some pcv deposits have an overage of tokens and some do not', async () => {
+      // Mint 4000, 4000, and 4000 tokens to each pcv deposit, respectively
```
Joeysantoro, line 237: this comment is off

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+
+      // Call rebalance
+      await pcvDepositAggregator.rebalance();
+
+      // Check pcv deposit balances
+      // Should be 2000, 3000, 4000 in deposits
+      // Should be 1000 in the aggregator
+      expect(await token.balanceOf(pcvDeposit1.address)).to.equal(ethers.utils.parseEther('2000'));
+      expect(await token.balanceOf(pcvDeposit2.address)).to.equal(ethers.utils.parseEther('3000'));
+      expect(await token.balanceOf(pcvDeposit3.address)).to.equal(ethers.utils.parseEther('4000'));
```
Joeysantoro, line 258: should also check .balance() calls maybe

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+
+      // Also check the aggregator balance
+      expect(await token.balanceOf(pcvDepositAggregator.address)).to.equal(ethers.utils.parseEther('1000'));
```
Joeysantoro, line 261: this should use a .balance() call to check too

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+      expect(deposits[0]).to.equal(pcvDeposit1.address);
+      expect(deposits[1]).to.equal(pcvDeposit2.address);
+      expect(deposits[2]).to.equal(pcvDeposit3.address);
+
+      expect(weights[0]).to.equal(20);
+      expect(weights[1]).to.equal(30);
+      expect(weights[2]).to.equal(40);
+    });
+
+    it('successfully rebalances when all pcv deposits need tokens', async () => {
```
Joeysantoro, line 206: Should also check case where all deposits are overfunded

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,444 @@
...
+        core.address,
+        token.address,
+        [pcvDeposit1.address, pcvDeposit2.address, pcvDeposit3.address],
+        [20, 30, 40],
+        10
+      );
+      await pcvDepositAggregator.deployTransaction.wait();
+    });
+
+    it('returns expected values for deposits, weights, balances, and token', async () => {
```
Joeysantoro, line 172: This is similar to the Init functionality normally used but also has some other logic around the depositing behavior

Joeysantoro (commented):
Two unrelated macro issues, one is that withdraw occasionally has a rounding error on the amount withdrawn, which would brick withrdawals if under expected.

the other is that resetting the PCVDeposit aggregator needs testing

contracts/pcv/IPCVDepositAggregator.sol
```diff
@@ -11,38 +11,70 @@ import "../external/Decimal.sol";
...
 
     /// @notice replaces this contract with a new PCV Deposit Aggregator on the rewardsAssetManager
-    function setNewAggregator(IPCVDepositAggregator newAggregator) external;
+    function setNewAggregator(address newAggregator) external;
 
     // ----------- Governor or Guardian only state changing api -----------
```
Joeysantoro, line 44: I don't think the Guardian category exists anymore

contracts/libs/UintArrayOps.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
```
Joeysantoro, line 4: SafeERC20 not needed

contracts/libs/UintArrayOps.sol
```diff
@@ -0,0 +1,43 @@
...
+        _difference = new int256[](a.length);
+
+        for (uint256 i=0; i < a.length; i++) {
+            _difference[i] = a[i].toInt256() - b[i].toInt256();
+        }
+
+        return _difference;
+    }
+
+    function positiveDifference(uint[] memory a, uint[] memory b) internal pure returns (uint[] memory _positiveDifference) {
```
Joeysantoro, line 30: This method is a bit unintuitive from the name. Not sure if there is an action item but at least a comment explaining should help

contracts/libs/UintArrayOps.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/utils/math/SafeCast.sol";
+
+library UintArrayOps {
+    using SafeCast for uint256;
+
+    function sum(uint[] memory array) internal pure returns (uint256 _sum) {
```
Joeysantoro, line 10: uint256 everywhere

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,444 @@
...
+import "../refs/CoreRef.sol";
+import "../external/Decimal.sol";
+import "./balancer/IRewardsAssetManager.sol";
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/utils/math/SafeCast.sol";
+import "../libs/UintArrayOps.sol";
+
+/// @title PCV Deposit Aggregator
+/// @notice A smart contract that aggregates erc20-based PCV deposits and rebalances them according to set weights
```
Joeysantoro, line 15: will probably need an ETH compatible one right? Should still base in WETH but have convertibility of some kind

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,444 @@
...
+        address _token,
+        address[] memory _initialPCVDepositAddresses,
+        uint128[] memory _initialPCVDepositWeights,
+        uint128 _bufferWeight
+    ) CoreRef(_core) {
+        require(_initialPCVDepositAddresses.length == _initialPCVDepositWeights.length, "Addresses and weights are not the same length!");
+        require(_assetManager != address(0x0), "Rewards asset manager cannot be null");
+        require(IRewardsAssetManager(_assetManager).getToken() == _token, "Rewards asset manager must be for the same token as this.");
+
+        assetManager = _assetManager;
```
Joeysantoro, line 53: assetManager should use an internal setter

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,444 @@
...
+        }
+
+        emit AggregatorDeposit();
+    }
+
+    /// @notice tops up a deposit from the aggregator's balance
+    /// @param pcvDeposit the address of the pcv deposit to top up
+    /// @dev this will only pull from the balance that is left over after the aggregator's buffer fills up
+    function depositSingle(address pcvDeposit) public override whenNotPaused {
+        revert("Method not yet written.");
```
Joeysantoro, line 112: lol

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,444 @@
...
+        uint128 _bufferWeight
+    ) CoreRef(_core) {
+        require(_initialPCVDepositAddresses.length == _initialPCVDepositWeights.length, "Addresses and weights are not the same length!");
+        require(_assetManager != address(0x0), "Rewards asset manager cannot be null");
+        require(IRewardsAssetManager(_assetManager).getToken() == _token, "Rewards asset manager must be for the same token as this.");
+
+        assetManager = _assetManager;
+        token = _token;
+        bufferWeight = _bufferWeight;
+        totalWeight = bufferWeight;
```
Joeysantoro, line 56: Should use _setBufferWeight

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,444 @@
...
+        for (uint256 i=0; i < _initialPCVDepositAddresses.length; i++) {
+            require(IPCVDeposit(_initialPCVDepositAddresses[i]).balanceReportedIn() == _token, "Deposit token must be the same as for this aggregator.");
+            _addPCVDeposit(_initialPCVDepositAddresses[i], _initialPCVDepositWeights[i]);
+        }
+    }
+
+    // ---------- Public Functions -------------
+
+    /// @notice deposits tokens into sub-contracts (if needed)
+    /// @dev this is equivalent to half of a rebalance. the implementation is as follows:
```
Joeysantoro, line 69: still half a rebalance?

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,444 @@
...
+        // we should fill up the buffer before sending out to sub-deposits
+        uint256 amountAvailableForUnderlyingDeposits = optimalAggregatorBalance - actualAggregatorBalance;
+
+        // calculate the amount that each pcv deposit needs. if they have an overage this is 0.
+        uint256[] memory optimalUnderlyingBalances = _getOptimalUnderlyingBalances(totalBalance);
+        uint256[] memory amountsNeeded = optimalUnderlyingBalances.positiveDifference(underlyingBalances);
+        uint256 totalAmountNeeded = amountsNeeded.sum();
+
+        // calculate a scalar. this will determine how much we *actually* send to each underlying deposit.
+        Decimal.D256 memory scalar = Decimal.ratio(amountAvailableForUnderlyingDeposits, totalAmountNeeded);
```
Joeysantoro, line 95: should probably assert this is <= 1 right?

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,444 @@
...
+    /// 1. check if the contract has enough in the buffer to cover the withdrawal. if so, just use this
+    /// 2. if not, calculate what the ideal underlying amount should be for each pcv deposit *after* the withdraw
+    /// 3. then, cycle through them and withdraw until each has their ideal amount (for the ones that have overages)
+    /// Note this function will withdraw all of the overages from each pcv deposit, even if we don't need that much to
+    /// actually cover the transfer! This is intentional because it costs the same to withdraw exactly how much we need
+    /// vs the overage amount; the entire overage amount should be moved if it is the same cost as just as much as we need.
+    function withdraw(address to, uint256 amount) external override onlyPCVController whenNotPaused {
+        uint256 aggregatorBalance = balance();
+
+        if (aggregatorBalance > amount) {
```
Joeysantoro, line 126: geq

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,444 @@
...
+        // (pulls from the deposits that have > than their post-withdraw-ideal-underlying-balance)
+        for (uint256 i=0; i < pcvDepositAddresses.length(); i++) {
+            address pcvDepositAddress = pcvDepositAddresses.at(i);
+            uint256 actualPcvDepositBalance = underlyingBalances[i];
+            uint256 idealPcvDepositBalance = idealUnderlyingBalancesPostWithdraw[i];
+
+            if (actualPcvDepositBalance > idealPcvDepositBalance) {
+                // Has post-withdraw-overage; let's take it
+                uint256 amountToWithdraw = actualPcvDepositBalance - idealPcvDepositBalance;
+                IPCVDeposit(pcvDepositAddress).withdraw(address(this), amountToWithdraw);
```
Joeysantoro, line 159: This can withdraw too much, right?

Joeysantoro: amountToWithdraw is unrelated to the originally requested `amount`

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,444 @@
...
+
+    /// @notice sets the weight of a pcv deposit to zero
+    /// @param depositAddress the address of the pcv deposit to set the weight of to zero
+    function setPCVDepositWeightZero(address depositAddress) external override onlyGuardianOrGovernor {
+        require(pcvDepositAddresses.contains(depositAddress), "Deposit does not exist.");
+
+        uint256 oldDepositWeight = pcvDepositWeights[depositAddress];
+        pcvDepositWeights[depositAddress] = 0;
+
+        totalWeight = (totalWeight.toInt256() - oldDepositWeight.toInt256()).toUint256();
```
Joeysantoro, line 197: why is casting needed here?

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,444 @@
...
+    function setNewAggregator(address newAggregator) external override onlyGovernor {
+        require(PCVDepositAggregator(newAggregator).token() == token, "New aggregator must be for the same token as the existing.");
+
+        // Add each pcvDeposit to the new aggregator
+        for (uint256 i=0; i < pcvDepositAddresses.length(); i++) {
+            address pcvDepositAddress = pcvDepositAddresses.at(i);
+            if (!(IPCVDepositAggregator(newAggregator).hasPCVDeposit(pcvDepositAddress))) {
+                uint256 pcvDepositWeight = pcvDepositWeights[pcvDepositAddress];
+                IPCVDepositAggregator(newAggregator).addPCVDeposit(pcvDepositAddress, pcvDepositWeight);
+            }
```
Joeysantoro, line 229: I don't think this is needed. It means this contract has to be an admin of the new aggregator

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,444 @@
...
+    /// @return true if the given address is a PCV Deposit in this aggregator
+    function hasPCVDeposit(address pcvDeposit) public view override returns (bool) {
+        return pcvDepositAddresses.contains(pcvDeposit);
+    }
+
+    function resistantBalanceAndFei() public view override returns (uint256, uint256) {
+        return (balance(), 0);
+    }
+
+    function balanceReportedIn() external view override returns (address) {
```
Joeysantoro, line 268: Natspec

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -0,0 +1,444 @@
...
+        uint256 totalBalance = aggregatorBalance + underlyingSum;
+
+        distancesToTargets = new int256[](pcvDepositAddresses.length());
+
+        for (uint256 i=0; i < distancesToTargets.length; i++) {
+            uint256 idealAmount = totalBalance * pcvDepositWeights[pcvDepositAddresses.at(i)] / totalWeight;
+            distancesToTargets[i] = (idealAmount).toInt256() - (underlyingBalances[i]).toInt256();
+        }
+
+        return distancesToTargets;
```
Joeysantoro, line 324: Do you imagine either of these methods being useful and worth keeping? They don't share code anywhere with the above. I had originally thought they might be useful but now I'm not sure

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: ...oops

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: I don't believe it's possible to be > 1 but wouldn't hurt to add the assert.

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: not, fixed

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: already in the interface (which isn't even IPCVGuardian)

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: could be useful for stats and stuff. doesn't hurt to keep it in.

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: Nope. See the note at the top of this function:

```
    /// Note this function will withdraw all of the overages from each pcv deposit, even if we don't need that much to
    /// actually cover the transfer! This is intentional because it costs the same to withdraw exactly how much we need
    /// vs the overage amount; the entire overage amount should be moved if it is the same cost as just as much as we need.
    ```

File: contracts/pcv/PCVDepositAggregator.sol (hunk 63 above)
Comment (ditchfieldcaleb) [reply contracts/pcv/PCVDepositAggregator.sol:268]: right, reason I didn't add this is because there's no natespec for the pcv deposit interface or pcv deposit implementation to draw from

Comment (ditchfieldcaleb) [reply contracts/pcv/PCVDepositAggregator.sol:268]: added

File: contracts/pcv/PCVDepositAggregator.sol (hunk 53 above)
Comment (ditchfieldcaleb) [reply contracts/pcv/PCVDepositAggregator.sol:15]: yes, I will make a pr onto this pr that implements this (it will probably just subclass the PCVDepositAggregator, ie an ETHPCVDepositAggregator or such)

**Comment by ditchfieldcaleb** [comment]
todo: add tests for pcvDepositSingle

File: contracts/pcv/PCVDepositAggregator.sol (hunk 54 above)
Comment (ditchfieldcaleb) [reply contracts/pcv/PCVDepositAggregator.sol]: already talked about this - can't do this because you can't read immutable vars in the constructor

File: contracts/pcv/PCVDepositAggregator.sol (hunk 60 above)
Comment (Joeysantoro) [reply contracts/pcv/PCVDepositAggregator.sol:185]: If you pair this with deposit, we have the same security concerns as before as atomic withdraw + deposit is equivalent to rebalance

File: contracts/pcv/PCVDepositAggregator.sol [hunk 65]
```diff
@@ -0,0 +1,456 @@
...
+    function removePCVDeposit(address pcvDeposit) external virtual override onlyGovernorOrAdmin {
+        _removePCVDeposit(address(pcvDeposit));
+    }
+
+    function addPCVDeposit(address newPCVDeposit, uint256 weight) external virtual override onlyGovernorOrAdmin {
+        _addPCVDeposit(address(newPCVDeposit), uint128(weight));
+    }
+
+    function setNewAggregator(address newAggregator) external virtual override onlyGovernor {
+        require(PCVDepositAggregator(newAggregator).token() == token, "New aggregator must be for the same token as the existing.");
```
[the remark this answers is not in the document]
Joeysantoro: @ditchfieldcaleb yeah you can add raw variables to an interface. Its the same function selector whether its a variable or a function

contracts/pcv/PCVDepositAggregator.sol
```diff
@@ -108,7 +109,33 @@ contract PCVDepositAggregator is IPCVDepositAggregator, PCVDeposit {
...
+        uint256 actualDepositBalance = IPCVDeposit(pcvDeposit).balance();
+
+        require(actualDepositBalance < optimalDepositBalance, "Deposit does not need topping up.");
+
+        uint256 amountToSend = optimalDepositBalance - actualDepositBalance;
+
+        // If we don't have enough overage to send the whole amount, send as much as we can
+        if (amountToSend > aggregatorOverage) {
+            amountToSend = aggregatorOverage;
+        }
```
Joeysantoro, line 134: nit: I think OZ Math.min() could make this more readable

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,761 @@
...
+    // add any addresses you want to impersonate here
+    const impersonatedAddresses = [userAddress, pcvControllerAddress, governorAddress, guardianAddress];
+
+    for (const address of impersonatedAddresses) {
+      await hre.network.provider.request({
+        method: 'hardhat_impersonateAccount',
+        params: [address]
+      });
+
+      impersonatedSigners[address] = await ethers.getSigner(address);
```
Joeysantoro, line 55: nit: getImpersonatedSigner is cleaner

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: Aaaaaah you're right. I'll remove it.

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,761 @@
...
+    });
+
+    // This test covers the special edge case with the following context:
+    // 1. The buffer is not enough to cover the balances
+    // 2. There is (at least one) pcv deposit that has a defecit
+    // 3. Said defecit is *still* a defecit after the withdrawal
+    // This edge case is special because it is the only time when we DONT pull tokens from every pcv deposit to cover the overage,
+    // and where we'll actually end up pulling more tokens than needed into the aggregator - ie the aggregatort will have an overage
+    // after this method is complete. This is because we don't do deposits of tokens on withdraw - only withdrawals.
+    // @todo
```
Joeysantoro, line 435: what is this todo for?

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,761 @@
...
+
+      const aggregatorBalance = await token.balanceOf(pcvDepositAggregator.address);
+
+      const sum = pcvDeposit1Balance.add(pcvDeposit2Balance).add(pcvDeposit3Balance).add(aggregatorBalance);
+
+      //console.log(`pcv deposit 1 balance: ${ethers.utils.formatUnits(pcvDeposit1Balance)}`);
+      //console.log(`pcv deposit 2 balance: ${ethers.utils.formatUnits(pcvDeposit2Balance)}`);
+      //console.log(`pcv deposit 3 balance: ${ethers.utils.formatUnits(pcvDeposit3Balance)}`);
+      //console.log(`aggregator balance: ${ethers.utils.formatUnits(aggregatorBalance)}`);
+      //console.log(`sum: ${ethers.utils.formatUnits(sum)}`);
```
Joeysantoro, line 542: Remove console logs

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,761 @@
...
+      const pcvDeposit3Balance = await token.balanceOf(pcvDeposit3.address);
+      const aggregatorBalance = await token.balanceOf(pcvDepositAggregator.address);
+
+      const sum = pcvDeposit1Balance.add(pcvDeposit2Balance).add(pcvDeposit3Balance).add(aggregatorBalance);
+
+      //console.log(`pcv deposit 1 balance: ${ethers.utils.formatUnits(pcvDeposit1Balance)}`);
+      //console.log(`pcv deposit 2 balance: ${ethers.utils.formatUnits(pcvDeposit2Balance)}`);
+      //console.log(`pcv deposit 3 balance: ${ethers.utils.formatUnits(pcvDeposit3Balance)}`);
+      //console.log(`aggregator balance: ${ethers.utils.formatUnits(aggregatorBalance)}`);
+      //console.log(`sum: ${sum.toNumber()}`)
```
Joeysantoro, line 574: Remove console logs

test/unit/pcv/PCVDepositAggregator.test.ts
```diff
@@ -0,0 +1,761 @@
...
+
+      expect(sum.toNumber()).to.equal(10000001101);
+    });
+
+    it('deposit-singles', async () => {
```
Joeysantoro, line 579: This should be two tests, one for sad and one for happy path

contracts/pcv/PCVDepositAggregator.sol
ditchfieldcaleb: withdraw is onlyPCVController though

Joeysantoro (approved):

PR #270, opened 2021-10-25
Revoke Burner Role

Joeysantoro:
Adds a RestrictedPermissions contract which ignores or reverts on checks for burner, governor, and pcv controller.

It passes through checks for minter and guardian to the main core contract.

The intended usage is to setCore() on Fei contract so the burner functionality can no longer be used, nor transfer hooks.


contracts/core/RestrictedPermissions.sol
```diff
@@ -0,0 +1,56 @@
...
+
+    // ---------- Deprecated roles for caller ---------
+
+    /// @dev returns false rather than reverting so calls to onlyGuardianOrGovernor don't revert
+    function isGovernor(address) external pure override returns (bool) {
+        return false;
+    }
+
+    function isPCVController(address) external pure override returns (bool) {
+        revert("RestrictedPermissions: PCV Controller deprecated for contract");
```
ditchfieldcaleb, line 50: I'm strongly against a view-or-pure function reverting in all cases, or even at all.
Thoughts on making this just return false?

contracts/core/RestrictedPermissions.sol
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    function isBurner(address) external pure override returns (bool) {
+        revert("RestrictedPermissions: Burner deprecated for contract");
```
ditchfieldcaleb, line 54: (see above)

proposals/dao/permanentlyRevokeBurner.ts
```diff
@@ -0,0 +1,61 @@
+import { ethers } from 'hardhat';
+import chai, { expect } from 'chai';
+import CBN from 'chai-bn';
+import {
+  DeployUpgradeFunc,
+  NamedContracts,
+  SetupUpgradeFunc,
+  TeardownUpgradeFunc,
+  ValidateUpgradeFunc
+} from '../../types/types';
```
ditchfieldcaleb, line 10: use @ syntax

proposals/dao/permanentlyRevokeBurner.ts
```diff
@@ -0,0 +1,61 @@
...
+DAO ACTIONS:
+1. setCore on Fei to restrictedPermissions
+
+*/
+
+export const deploy: DeployUpgradeFunc = async (deployAddress, addresses, logging = false) => {
+  const { core } = addresses;
+
+  if (!core) {
+    throw new Error('An environment variable contract address is not set');
```
ditchfieldcaleb, line 30: this isn't an environment variable

test/integration/tests/bondingcurve.ts
```diff
@@ -47,7 +47,7 @@ describe('e2e-bondingcurve', function () {
     doLogging && console.log(`Environment loaded.`);
   });
 
-  describe('Reserve Stabilizer', async () => {
+  describe.skip('Reserve Stabilizer', async () => {
```
ditchfieldcaleb, line 50: Why is this skipped?

ditchfieldcaleb (changes requested):
See comments.

Honestly the way that this revokes the roles breaks some, imo, very important patterns and makes me very wary as is:

- any contract that might exist that calls fei.getCore() to do some operation via the core will break. even if we don't have any contract that do this (I'm not sure if we don't), there's probably other contracts by external parties that do. one solution to this is to proxy-through all calls that aren't the ones we define in restrictedPermissions, which is not hard to do, and I think probably a much better pattern

- I really dislike reverting when calling a view/pure function, especially one that should be a getter. "isGovernor" should never revert, just return false if disabled as in this case.

ditchfieldcaleb:
also should fix e2e tests or skip them if irrelevant

test/integration/tests/fei.ts
```diff
@@ -0,0 +1,98 @@
...
+import { Signer } from '@ethersproject/abstract-signer';
+const toBN = ethers.BigNumber.from;
+
+before(async () => {
+  chai.use(CBN(ethers.BigNumber));
+  chai.use(solidity);
+  await resetFork();
+});
+
+describe.only('e2e-fei', function () {
```
ditchfieldcaleb, line 19: remove .only

ditchfieldcaleb (dismissed):

ditchfieldcaleb:
Looks good to me - just fix that .only and should be fine. The failing e2e tests are annoying so if you could fix them too that'd be great, but they aren't relevant to this PR.

ditchfieldcaleb (approved):

PR #278, opened 2021-10-31
Fix Check Proposal

Joeysantoro:
Check proposal needs to pass through when proposal has already been voted on and executed.

There is a more elegant way to do this using the proposal state but these were already here just commented out


ditchfieldcaleb (changes requested):
Swallowing all possible errors Isn't great. Can we change this to only swallow the error of "transaction reverted" and have any other errors throw?

I'm the one that originally commented this out because it was swallowing errors and caused us to miss a bug or two in proposal code awhile back.

ditchfieldcaleb (approved):

PR #281, opened 2021-11-03
fix construct proposal

Joeysantoro:


ditchfieldcaleb (approved):

ditchfieldcaleb (changes requested):
@Joeysantoro this is failing the e2e test suite because your changes are causing compile issues in TS

PR #302, opened 2021-11-07
[WIP] ETH PCV Deposit Aggregator

ditchfieldcaleb:


contracts/pcv/ETHPCVDepositAggregator.sol
```diff
@@ -0,0 +1,83 @@
...
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/utils/math/SafeCast.sol";
+import "../libs/UintArrayOps.sol";
+
+/// @title ETH PCV Deposit Aggregator
+/// @notice A smart contract that aggregates erc20-based PCV deposits and rebalances them according to set weights ]
+contract ETHPCVDepositAggregator is PCVDepositAggregator {
+
+    address constant private WETH_ADDRESS = address(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
```
ElliotFriedman, line 19: You should be able to remove this and just use the weth address from the constants library instead.

contracts/pcv/ETHPCVDepositAggregator.sol
```diff
@@ -0,0 +1,83 @@
...
+        _initialPCVDepositAddresses,
+        _initialPCVDepositWeights,
+        _bufferWeight
+    ) { }
+
+    receive() external payable {
+        if (msg.value > 0) {
+            wrapETH(msg.value);
+        }
+    }
```
Joeysantoro, line 41: This is redundant, if fallback automatically wraps.

eswak: there might not be enough gas to wrapETH here, would be better to have a separate function ?

PR #331, opened 2021-11-23
simplify ERC20 compound pcv deposit

Joeysantoro:
Used to make https://etherscan.io/address/0x6026a1559cdd44a63c5ca9a078cc996a9eb68abb#readContract


eswak (dismissed):
cool, that should be more error-proof

PR #315, opened 2021-11-12
[WIP] LUSD Swapper

Joeysantoro:

PR #344, opened 2021-12-03
FIP-50 Yield improvements

ElliotFriedman:


proposals/dao/lusd_swap_grant_swapper_role.ts
```diff
@@ -0,0 +1,57 @@
...
+// cleaned up before doing any validation checks.
+const teardown: TeardownUpgradeFunc = async (addresses, oldContracts, contracts, logging) => {
+  console.log(`No actions to complete in teardown for fip${fipNumber}`);
+};
+
+// Run any validations required on the fip using mocha or console logging
+// IE check balances, check state of contracts, etc.
+const validate: ValidateUpgradeFunc = async (addresses, oldContracts, contracts, logging) => {
+  const { core, optimisticTimelock, lusd, lusdPCVDeposit } = contracts;
+  const currentLUSDBalance = await lusd.balanceOf(lusdPCVDeposit.address);
```
ditchfieldcaleb, line 51: Add in some validations that test near the order of magnitude what the token balances should be. No need to test for the exact amount but it's reasonable to make sure it's got at least x million tokens, where x is the amount that we estimate we'll be pulling out from the swapper.

ditchfieldcaleb (changes requested):

eswak (changes requested):
Instead of granting a role, starting a new swap, etc : we don't want to start another auction, so you should call : 
- feiLusdLBPSwapper.exitPool(feiLusdLBPSwapper)
- ratioPCVController.withdrawRatioERC20(feiLusdLBPSwapper, lusd, lusdPCVDeposit, 10000)
- same for moving the FEI somewhere else, e.g. the OA multisig

this will move all the LUSD to the compound deposit, not start a new auction, and no need to grant a new role

Also don't forget to check the lusdPCVDeposit.balance() after, should be at least 100M

proposals/dao/lusd_swap_grant_swapper_role.ts
ElliotFriedman: A validation 2 lines down verifies that the lusdPCVDeposit has at least 90m LUSD.

ElliotFriedman:
> Instead of granting a role, starting a new swap, etc : we don't want to start another auction, so you should call :
> 
> * feiLusdLBPSwapper.exitPool(feiLusdLBPSwapper)
> * ratioPCVController.withdrawRatioERC20(feiLusdLBPSwapper, lusd, lusdPCVDeposit, 10000)
> * same for moving the FEI somewhere else, e.g. the OA multisig
> 
> this will move all the LUSD to the compound deposit, not start a new auction, and no need to grant a new role
> 
> Also don't forget to check the lusdPCVDeposit.balance() after, should be at least 100M

I like using exit pool better than calling swap. Does exiting the pool end the auction? Why would we exit the pool and send all of the ERC20 tokens directly back to the pool? It makes more sense to exitPool and use the ERC20CompoundPCVDeposit for LUSD as the token recipient.

To actually end the auction, we need to call the setSwapEnabled which requires the governor or admin role.

Exiting the pool to the ERC20CompoundPCVDeposit for LUSD, transferring all FEI in the ERC20CompoundPCVDeposit to the OA timelock, and then calling setSwapEnabled on the BalancerLBPSwapper and setting the swap enabled to false seems like the best path forward.

Joeysantoro:
> I like using exit pool better than calling swap. Does exiting the pool end the auction? Why would we exit the pool and send all of the ERC20 tokens directly back to the pool? It makes more sense to exitPool and use the ERC20CompoundPCVDeposit for LUSD as the token recipient.

We don't need to end the auction we just need the tokens. ExitPool is the right approach, good catch @eswak. We'd send the tokens elsewhere

> Exiting the pool to the ERC20CompoundPCVDeposit for LUSD, transferring all FEI in the ERC20CompoundPCVDeposit to the OA timelock, and then calling setSwapEnabled on the BalancerLBPSwapper and setting the swap enabled to false seems like the best path forward.

This seems find to me, although the setSwapEnabled isn't strictly necessary. I think we should have some light governance around where to send LUSD. Pool 7 looks nice for at least 10m.

contracts/mock/MockChainlinkOracle.sol
```diff
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.4;
 
 import "@chainlink/contracts/src/v0.6/interfaces/AggregatorV3Interface.sol";
+import "hardhat/console.sol";
```
Joeysantoro, line 5: todo clean up

proposals/dao/fip_50.ts
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const validate: ValidateUpgradeFunc = async (addresses, oldContracts, contracts, logging) => {
+  // Check BAMM holdings via lens
+  console.log(await contracts.bammLens.balance());
+
+  // Check balancer LBPSwapper balance is near 0 (note these are still reported in wei)
+  const remainingBalances = await contracts.feiLusdLens.resistantBalanceAndFei();
+  expectApprox(remainingBalances[0], 99229997);
+  expectApprox(remainingBalances[1], 1009159);
```
eswak, line 64: after an exitPool there should be 0 left in the pool ?

proposals/dao/fip_50.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  // check Pool7LUSDDeposit holding 10m
+  expect(await contracts.rariPool7LusdPCVDeposit.balance()).to.be.bignumber.equal(
+    ethers.constants.WeiPerEther.mul(10_000_000)
+  );
+
+  // Check CR Oracle
+  // Check stETH sitting on about 48k ETH
+  expectApprox(await contracts.ethLidoPCVDeposit.balance(), ethers.constants.WeiPerEther.mul(48_700));
```
eswak, line 73: what exactly is the boundary checked here ? I got confused multiple times with `expectApprox` because it checks with orders of magnitude, and I had my tests passing but actually I was testing something totally irrelevant

Check would be more readable with `.to.be.at.least(...)`

proposals/dao/fip_50.ts
```diff
@@ -0,0 +1,76 @@
...
+  await overwriteChainlinkAggregator(addresses.chainlinkEthUsdOracle);
+};
+
+const teardown: TeardownUpgradeFunc = async (addresses, oldContracts, contracts, logging) => {
+  console.log(`No actions to complete in teardown for fip${fipNumber}`);
+};
+
+const validate: ValidateUpgradeFunc = async (addresses, oldContracts, contracts, logging) => {
+  // Check BAMM holdings via lens
+  console.log(await contracts.bammLens.balance());
```
eswak, line 59: add a proper check here, like "at least 89M"

proposals/description/fip_50.ts
```diff
@@ -0,0 +1,94 @@
...
+      arguments: [['{feiLusdLens}']],
+      description: 'Remove Old PCV Deposits from Collateralization Oracle'
+    },
+    {
+      target: 'collateralizationOracle',
+      values: '0',
+      method: 'addDeposits(address[])',
+      arguments: [['{bammLens}']],
+      description: 'Add New PCV Deposits to Collateralization Oracle'
+    }
```
eswak, line 89: you could use swapDeposit here but it does not matter much

test/helpers.ts
```diff
@@ -4,6 +4,7 @@ import CBN from 'chai-bn';
 import { Core, Core__factory } from '@custom-types/contracts';
 import { BigNumber, BigNumberish, Signer } from 'ethers';
 import { NamedAddresses } from '@custom-types/types';
+import { sign } from 'crypto';
```
eswak, line 7: doesn't seem to be used

test/helpers.ts
```diff
@@ -187,6 +188,18 @@ const balance = {
   }
 };
 
+async function overwriteChainlinkAggregator(chainlink) {
```
eswak, line 191: you could add the value (4000) and decimals (8) as args of the helper

test/integration/proposals_config.ts
```diff
@@ -16,23 +17,29 @@ const proposals: ProposalsConfigMap = {
...
-    proposal: peg_stability_module
+    proposal: backstop_proposal
   },
-  backstop: {
+  peg_stability_module: {
     deploy: true,
     skipDAO: false,
     totalValue: 0,
-    proposal: backstop_proposal
+    proposal: peg_stability_module
```
eswak, line 30: weird diff, you inverted the position of the two objects ?

test/integration/tests/fei.ts
```diff
@@ -49,7 +49,7 @@ describe('e2e-fei', function () {
   });
 
   describe('Fei Functionality', async function () {
-    it('setIncentiveContract', async function () {
+    it.skip('setIncentiveContract', async function () {
```
eswak, line 52: ?

contract-addresses/mainnetAddresses.ts
```diff
@@ -378,6 +383,10 @@ const MainnetAddresses = {
   masterKashi: { artifactName: 'unknown', address: '0x2cba6ab6574646badc84f0544d05059e57a5dc42' },
   multisend: { artifactName: 'IERC20Airdropper', address: '0x0B36b0F351ea8383506F596743a2DA7DCa204cc3' },
   multisig: { artifactName: 'unknown', address: '0xB8f482539F2d3Ae2C9ea6076894df36D1f632775' },
+  lusdPCVDeposit: {
```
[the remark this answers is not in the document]
Joeysantoro: It will be used later for the plugin from https://github.com/fei-protocol/fei-protocol-core/pull/315 . Not relevant for this PR

contracts/mock/MockChainlinkOracle.sol
```diff
@@ -49,7 +49,7 @@ contract MockChainlinkOracle is AggregatorV3Interface {
         uint256 updatedAt,
         uint80 answeredInRound
     ) {
-      return (_roundId, _value, _startedAt, _updatedAt, _answeredInRound);
```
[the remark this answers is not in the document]
Joeysantoro: It didn't break any tests that we currently have

proposals/dao/fip_50.ts
Joeysantoro: Like uniswap, there is dust

proposals/dao/fip_50.ts
Joeysantoro: Yeah its honestly a pretty weird method. I'll change the usage here.

proposals/dao/fip_50.ts
Joeysantoro: good catch I meant to do this

contracts/pcv/liquity/BAMMDeposit.sol
```diff
@@ -0,0 +1,71 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.0;
```
eswak, line 2: I noticed sometimes we use ^0.8.0 and sometimes ^0.8.4, do you think it matters ? local config is 0.8.10 anyway

contracts/pcv/liquity/BAMMDeposit.sol
```diff
@@ -0,0 +1,71 @@
...
+        BAMM.deposit(amount);
+    }
+
+    /// @notice withdraw LUSD from B Protocol BAMM
+    function withdraw(address to, uint256 amount) external override onlyPCVController {
+        uint256 totalSupply = BAMM.totalSupply();
+        uint256 lusdValue = stabilityPool.getCompoundedLUSDDeposit(address(BAMM));
+        uint256 shares = (amount * totalSupply / lusdValue) + 1; // extra unit to prevent truncation errors
+
+        // Withdraw the LUSD from BAMM (also withdraws LQTY and dust ETH)
```
eswak, line 47: you might need a receive fonction on the contract if some ETH is sent, like when unwrapping WETH ?

contracts/pcv/liquity/BAMMDeposit.sol
```diff
@@ -0,0 +1,71 @@
...
+        uint256 eth2usdPrice = BAMM.fetchPrice();
+        require(eth2usdPrice != 0, "chainlink is down");
+
+        uint256 ethUsdValue = ethBalance * eth2usdPrice / PRECISION;
+
+        uint256 bammLusdValue = stabilityPool.getCompoundedLUSDDeposit(address(BAMM));
+        return (bammLusdValue + ethUsdValue) * BAMM.balanceOf(address(this)) / BAMM.totalSupply();
+    }
+
+    function claim() public {
```
eswak, line 68: I called the function `claimRewards` in other deposits. Not a big deal because they take different arguments anyway, but for coherence you could use it too

contracts/pcv/liquity/BAMMDeposit.sol
Joeysantoro: shouldn't matter much but good to be consistent. We can pick one and update everything

contracts/pcv/liquity/BAMMDeposit.sol
Joeysantoro: Will do

contracts/pcv/liquity/BAMMDeposit.sol
Joeysantoro: good call this contract does receive ETH

contracts/pcv/liquity/BAMMDeposit.sol
ElliotFriedman: we should just lock all contracts to a specific version by doing `pragma solidity 0.8.10` or `pragma solidity =0.8.10` https://github.com/ConsenSys/smart-contract-best-practices/issues/125

test/integration/proposals_config.ts
ElliotFriedman: Maybe the linter made this happen?

proposals/description/fip_50.ts
```diff
@@ -0,0 +1,101 @@
...
+      method: 'deposit()',
+      arguments: [],
+      description: 'Deposit LUSD to pool 7'
+    },
+    {
+      target: 'lusd',
+      values: '0',
+      method: 'approve(address,uint256)',
+      arguments: ['{ratioPCVControllerV2}', '1000000000000000000000000000'],
+      description: 'Approve 1bn LUSD to RarioPCVControllerV2'
```
eswak, line 46: typo Ra**r**io

proposals/description/fip_50.ts
```diff
@@ -0,0 +1,101 @@
...
+      target: 'core',
+      values: '0',
+      method: 'grantRole(bytes32,address)',
+      arguments: ['0x471cfe1a44bf1b786db7d7104d51e6728ed7b90a35394ad7cc424adf8ed16816', '{optimisticTimelock}'],
+      description: 'Grant SWAP_ADMIN_ROLE to the OA'
+    },
+    {
+      target: 'feiLusdLBPSwapper',
+      values: '0',
+      method: 'exitPool(address)',
```
eswak, line 23: Do we burn the ~1M FEI that will be also sent on the Timelock along with the 99M LUSD ? They will count as user-circulating otherwise. Not sure we can know the exact amount that will land there though (someone could swap in the LUSD LBP)

proposals/description/fip_50.ts
Joeysantoro: fixing

proposals/description/fip_50.ts
Joeysantoro: Sending to OA Timelock

Joeysantoro (approved):

PR #346, opened 2021-12-04
Fix failing TC test

ElliotFriedman:


Joeysantoro:
Part of https://github.com/fei-protocol/fei-protocol-core/pull/347

PR #350, opened 2021-12-04
PCV Guardian Fix

Joeysantoro:


ditchfieldcaleb (commented):
TBH I don't think this is needed.

Allowing safe addresses that are already set to be "set" again seems fine, and will allow for the case where we forget or otherwise miss that we already have something as safe - we don't want to accidentally fail a whole dao proposal because of this.

(this what my intentional thinking when designing this part)

open to alternative opinions though, what's your rationale here @Joeysantoro ?

Joeysantoro:
> we don't want to accidentally fail a whole dao proposal because of this.

I feel that loud failure is far better here because our scripts should catch this early. Could swallow a worse issue quietly otherwise. Will not die on this hill but I think its a no-brainer from that perspective.

ditchfieldcaleb:
I'm fine with either @Joeysantoro (even if I do think the original is better :P )

PR #353, opened 2021-12-05
FIP-55: PCV Guardian

Joeysantoro:
https://tribe.fei.money/t/fip-55-pcv-guardian/3744

TODO
- [x] integration test


proposals/dao/fip_55.ts
```diff
@@ -0,0 +1,62 @@
+import { ethers } from 'hardhat';
+import chai, { expect } from 'chai';
+import CBN from 'chai-bn';
+import {
+  DeployUpgradeFunc,
+  NamedContracts,
+  SetupUpgradeFunc,
+  TeardownUpgradeFunc,
+  ValidateUpgradeFunc
+} from '../../types/types';
```
ditchfieldcaleb, line 10: use shorthand

ditchfieldcaleb (changes requested):
Needs an integration test or two, otherwise good.

ditchfieldcaleb:
@Joeysantoro I think we should add in a check when withdrawing to a safe address *other than the DAOTimelock* that the source and destination underlying token is the same. Otherwise, we could accidentally burn tokens by transferring them into a pcv deposit that can't handle them.

ditchfieldcaleb:
Also there's an e2e test on this that is failing that I'd like to see fixed before approving.

eswak (approved):

eswak:
> @Joeysantoro I think we should add in a check when withdrawing to a safe address _other than the DAOTimelock_ that the source and destination underlying token is the same. Otherwise, we could accidentally burn tokens by transferring them into a pcv deposit that can't handle them.

We shouldn't be able to burn tokens this way, because deposits have a `withdrawERC20` function

Joeysantoro:
> @Joeysantoro I think we should add in a check when withdrawing to a safe address _other than the DAOTimelock_ that the source and destination underlying token is the same. Otherwise, we could accidentally burn tokens by transferring them into a pcv deposit that can't handle them.

Interesting. I think we need to sync on this because at a glance its a good idea but it removes the flexibility to rely on withdrawERC20. I think we should put this check inside the `depositAfter` logic

proposals/dao/fip_55.ts
```diff
@@ -0,0 +1,62 @@
...
+export const setup: SetupUpgradeFunc = async (addresses, oldContracts, contracts, logging) => {
+  logging && console.log('No setup');
+};
+
+export const teardown: TeardownUpgradeFunc = async (addresses, oldContracts, contracts, logging) => {
+  logging && console.log('No teardown');
+};
+
+export const validate: ValidateUpgradeFunc = async (addresses, oldContracts, contracts) => {
+  const pcvGuardian: PCVGuardian = contracts.pcvGuardian as PCVGuardian;
```
ElliotFriedman, line 58: We shouldn't need to cast to type of PCVGuardian as the artifact that is loaded from contracts should be correct.

test/integration/tests/pcv.ts
```diff
@@ -53,6 +53,41 @@ describe('e2e-pcv', function () {
...
+      await pcvGuardian.withdrawToSafeAddress(
+        contractAddresses.rariPool8FeiPCVDeposit,
+        contractAddresses.feiDAOTimelock,
+        ethers.constants.WeiPerEther,
+        true,
+        false
+      );
+
+      expect(await contracts.rariPool8FeiPCVDeposit.paused()).to.be.true;
+      expect(await contracts.fei.balanceOf(contractAddresses.feiDAOTimelock)).to.be.bignumber.equal(
```
ElliotFriedman, line 84: We should add test to see delta between starting and ending balance.

ditchfieldcaleb (approved):

PR #355, opened 2021-12-05
Add PCV Deposit to PCV Guardian

Joeysantoro:


contracts/pcv/IPCVGuardian.sol
```diff
@@ -67,20 +67,20 @@ interface IPCVGuardian {
     /// @param safeAddress the destination address to withdraw to
     /// @param amount the amount to withdraw
     /// @param pauseAfter if true, the pcv contract will be paused after the withdraw
```
ditchfieldcaleb, line 69: Add natspec for new param

contracts/pcv/IPCVGuardian.sol
```diff
@@ -67,20 +67,20 @@ interface IPCVGuardian {
...
-    function withdrawToSafeAddress(address pcvDeposit, address safeAddress, uint256 amount, bool pauseAfter) external;
+    function withdrawToSafeAddress(address pcvDeposit, address safeAddress, uint256 amount, bool pauseAfter, bool depositAfter) external;
 
     /// @notice governor-or-guardian-only method to withdraw funds from a pcv deposit, by calling the withdraw() method on it
     /// @param pcvDeposit the address of the pcv deposit contract
     /// @param safeAddress the destination address to withdraw to
     /// @param amount the amount of tokens to withdraw
     /// @param pauseAfter if true, the pcv contract will be paused after the withdraw
-    function withdrawETHToSafeAddress(address pcvDeposit, address payable safeAddress, uint256 amount, bool pauseAfter) external;
+    function withdrawETHToSafeAddress(address pcvDeposit, address payable safeAddress, uint256 amount, bool pauseAfter, bool depositAfter) external;
```
ditchfieldcaleb, line 77: Add natspec for new param

contracts/pcv/IPCVGuardian.sol
```diff
@@ -67,20 +67,20 @@ interface IPCVGuardian {
...
 
     /// @notice governor-or-guardian-only method to withdraw funds from a pcv deposit, by calling the withdraw() method on it
     /// @param pcvDeposit the deposit to pull funds from
     /// @param safeAddress the destination address to withdraw to
     /// @param token the token to withdraw
     /// @param amount the amount of funds to withdraw
     /// @param pauseAfter whether to pause the pcv after withdrawing
-    function withdrawERC20ToSafeAddress(address pcvDeposit, address safeAddress, address token, uint256 amount, bool pauseAfter) external;
+    function withdrawERC20ToSafeAddress(address pcvDeposit, address safeAddress, address token, uint256 amount, bool pauseAfter, bool depositAfter) external;
```
ditchfieldcaleb, line 85: Add natspec for new param

contracts/pcv/PCVGuardian.sol
```diff
@@ -75,7 +75,13 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
     /// @param safeAddress the destination address to withdraw to
     /// @param amount the amount to withdraw
     /// @param pauseAfter if true, the pcv contract will be paused after the withdraw
```
ditchfieldcaleb, line 77: Add natspec for new param

contracts/pcv/PCVGuardian.sol
```diff
@@ -86,6 +92,10 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
             pcvDeposit._pause();
         }
 
+        if (depositAfter) {
+            IPCVDeposit(safeAddress).deposit();
```
ditchfieldcaleb, line 96: I think we might want to use a try catch here. If we're withdrawing a deposit to a safe address it's likely to be urgent/an emergency, and we don't want to risk the deposit part of this withdrawal reverting the entire thing.

contracts/pcv/PCVGuardian.sol
```diff
@@ -94,7 +104,13 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
     /// @param safeAddress the destination address to withdraw to
     /// @param amount the amount of tokens to withdraw
     /// @param pauseAfter if true, the pcv contract will be paused after the withdraw
```
ditchfieldcaleb, line 106: Add natspec for new param

contracts/pcv/PCVGuardian.sol
```diff
@@ -105,6 +121,10 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
             pcvDeposit._pause();
         }
 
+        if (depositAfter) {
+            IPCVDeposit(safeAddress).deposit();
```
ditchfieldcaleb, line 125: (see above comment about try catch)

contracts/pcv/PCVGuardian.sol
```diff
@@ -113,7 +133,14 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
     /// @param safeAddress the destination address to withdraw to
     /// @param amount the amount of funds to withdraw
     /// @param pauseAfter whether to pause the pcv after withdrawing
```
ditchfieldcaleb, line 135: Add natspec for new param

contracts/pcv/PCVGuardian.sol
```diff
@@ -124,6 +151,10 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
             pcvDeposit._pause();
         }
 
+        if (depositAfter) {
+            IPCVDeposit(safeAddress).deposit();
```
ditchfieldcaleb, line 155: (see above comment about trycatch)

ditchfieldcaleb (changes requested):
Good addition; some natspec fixes to do and please see my comment about try/catch for the deposit.

contracts/pcv/PCVGuardian.sol
Joeysantoro: Disagree, I think it try catch is an extremely dangerous pattern and this is code that should fail loud.

Joeysantoro: Plus in an emergency, Defender would alert us if this would fail in which case we just do false.

ditchfieldcaleb: The scenario in which we're using this is both Defender and manual. Either way, if we're using it, it means that *something has gone wrong* and we need to pull funds out of the danger zone, immediately. Depositing into the underlying deposit is sugar on top and isn't necessary and is another possible point of failure.

ditchfieldcaleb: Try catch isn't inherently dangerous. It's only dangerous if it's a problem if only part of your code succeeds. Here, it's not.

Joeysantoro: > Depositing into the underlying deposit is sugar on top and isn't necessary and is another possible point of failure. 

This is the part where we are sticking. Its not sugar on top, its critical functionality to have atomically. Thats why I made this PR. In many cases undeposited tokens aren't counted in CR oracle. In a perfect storm, this could open up the backstop or change risk curve dramatically. Better to control these variables.

contracts/pcv/PCVGuardian.sol
```diff
@@ -49,6 +49,7 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
     /// @notice batch version of setSafeAddress
     /// @param _safeAddresses the addresses to set as safe, as calldata
     function setSafeAddresses(address[] calldata _safeAddresses) external override onlyGovernorOrAdmin {
+        require(_safeAddresses.length != 0, "empty");
```
ditchfieldcaleb, line 52: Why add this? Seems unnecessary.

ditchfieldcaleb (approved):
Seems good but somewhat confused about the unnecessary require addition.

ditchfieldcaleb (dismissed):
Unit tests are failing, please fix before final approval.

eswak (approved):
ok, just need to fix the unit tests (add a parameter for the new `depositAfter` boolean arg, or add a 2nd signature in the contract that uses `false` by default)

contracts/pcv/PCVGuardian.sol
Joeysantoro: good practice to sanity check inputs

Joeysantoro:
> Unit tests are failing, please fix before final approval.

done

PR #379, opened 2021-12-14
Named Static Cached PCVDeposit

ElliotFriedman:


contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,180 @@
...
+
+  Returns PCV in USD terms
+*/
+contract NamedStaticPCVDepositWrapper is IPCVDepositBalances, CoreRef {
+    using SafeCast for *;
+
+    // -------------- Events ---------------
+    event BalanceUpdate(uint256 oldBalance, uint256 newBalance, uint256 oldFEIBalance, uint256 newFEIBalance);
+
+    event DepositRemoved(uint256 index);
```
Joeysantoro, line 20: Should absolutely emit the address and probably depositName of the removed deposit. With indexing

Joeysantoro: Also add a "DepositAdded" event

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,180 @@
...
+
+        uint256 oldBalance = balance;
+        uint256 oldFEIBalance = feiReportBalance;
+
+        balance += newPCVDeposit.usdAmount;
+        feiReportBalance += newPCVDeposit.feiAmount;
+        pcvDeposits.push(newPCVDeposit);
+
+        emit BalanceUpdate(oldBalance, balance, oldFEIBalance, feiReportBalance);
+    }
```
Joeysantoro, line 78: Addd depositAdded event

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,180 @@
...
+        updatePCVDeposit.usdAmount = usdAmount;
+        updatePCVDeposit.depositName = depositName;
+        updatePCVDeposit.underlyingTokenAmount = underlyingTokenAmount;
+        updatePCVDeposit.underlying = underlying;
+
+        uint256 oldBalance = balance;
+        uint256 oldFEIBalance = feiReportBalance;
+
+        balance = (balance.toInt256() + usdAmountDelta).toUint256();
+        feiReportBalance = (feiReportBalance.toInt256() + feiAmountDelta).toUint256();
```
Joeysantoro, line 108: subtract old then add new is a safer and more intuitive pattern than using unchecked integer arithmetic

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,180 @@
...
+
+        emit BalanceUpdate(oldBalance, balance, oldFEIBalance, feiReportBalance);
```
Joeysantoro, line 110: Should probably have an editDeposit event too

test/unit/pcv/NamedStaticPCVDepositWrapper.test.ts
```diff
@@ -0,0 +1,296 @@
...
+    // add any addresses you want to impersonate here
+    const impersonatedAddresses = [
+      addresses.userAddress,
+      addresses.pcvControllerAddress,
+      addresses.governorAddress,
+      addresses.pcvControllerAddress,
+      addresses.minterAddress,
+      addresses.burnerAddress,
+      addresses.beneficiaryAddress1,
+      addresses.beneficiaryAddress2
```
Joeysantoro, line 31: remove unused addresses
```suggestion
      addresses.governorAddress
```

test/unit/pcv/NamedStaticPCVDepositWrapper.test.ts
```diff
@@ -0,0 +1,296 @@
+import { getCore, getAddresses, expectRevert, expectEvent } from '../../helpers';
```
Joeysantoro, line 1: use typescript path shorthand
```suggestion
import { getCore, getAddresses, expectRevert, expectEvent } from '@test/helpers';
```

test/unit/pcv/NamedStaticPCVDepositWrapper.test.ts
```diff
@@ -0,0 +1,296 @@
...
+      addresses.beneficiaryAddress2
+    ];
+
+    for (const address of impersonatedAddresses) {
+      await hre.network.provider.request({
+        method: 'hardhat_impersonateAccount',
+        params: [address]
+      });
+
+      impersonatedSigners[address] = await ethers.getSigner(address);
```
Joeysantoro, line 40: use getImpersonatedSigner helper method NOTE: also needs to be imported from helpers

```suggestion

      impersonatedSigners[address] = await getImpersonatedSigner(address);
```

test/unit/pcv/NamedStaticPCVDepositWrapper.test.ts
```diff
@@ -0,0 +1,296 @@
...
+    }
+  });
+
+  beforeEach(async function () {
+    ({ governorAddress } = await getAddresses());
+
+    newDepositName = 'Visor Finance Deposit';
+    balance = '2000';
+    fei = '1000';
```
Joeysantoro, line 49: overloaded param name, change to something like `feiBalance`

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /// @notice function to bulk add deposits
+    function bulkAddDeposits(
+        PCVDeposit[] calldata newPCVDeposit
+    ) external onlyGovernorOrAdmin {
+        for (uint256 i = 0; i < newPCVDeposit.length; i++) {
+            _addDeposit(newPCVDeposit[i]);
+        }
+    }
```
Joeysantoro, line 127: nit
```suggestion
    function bulkAddDeposits(
        PCVDeposit[] calldata newPCVDeposits
    ) external onlyGovernorOrAdmin {
        for (uint256 i = 0; i < newPCVDeposits.length; i++) {
            _addDeposit(newPCVDeposits[i]);
        }
    }
```

Joeysantoro (changes requested):

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,180 @@
...
+
+    event DepositRemoved(uint256 index);
+
+    /// @notice struct to store info on each PCV Deposit
+    struct PCVDeposit {
+        string depositName;
+        uint256 usdAmount; /// USD equivalent in this deposit, not including FEI value
+        uint256 feiAmount; /// amount of FEI in this deposit
+        uint256 underlyingTokenAmount; /// amount of underlying token in this deposit
+        address underlying; /// address of the underlying token this deposit is reporting
```
eswak, line 28: for coherence, you could name it `underlyingToken` because you also have `underlyingTokenAmount`. Or rename `underlyingTokenAmount` to `underlyingAmount`

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /// @notice helper method to edit a PCV deposit
+    function _editDeposit(
+        uint256 index,
+        uint256 usdAmount,
+        uint256 feiAmount,
+        uint256 underlyingTokenAmount,
+        string calldata depositName,
+        address underlying
```
eswak, line 87: could you keep the same order than in the structure ? this way, all functions will take arguments in a coherent order

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,180 @@
...
+contract NamedStaticPCVDepositWrapper is IPCVDepositBalances, CoreRef {
+    using SafeCast for *;
+
+    // -------------- Events ---------------
+    event BalanceUpdate(uint256 oldBalance, uint256 newBalance, uint256 oldFEIBalance, uint256 newFEIBalance);
+
+    event DepositRemoved(uint256 index);
+
+    /// @notice struct to store info on each PCV Deposit
+    struct PCVDeposit {
```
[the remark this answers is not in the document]
ElliotFriedman: updated

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,180 @@
...
+
+    /// @notice funciont to remove a PCV Deposit
+    function removeDeposit(uint256 index) external onlyGovernorOrAdmin {
+        _removeDeposit(index);
+    }
+
+    // ----------- Getters -----------
+
+    function numDeposits() public view returns (uint256) {
+        return pcvDeposits.length;
```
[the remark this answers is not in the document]
ElliotFriedman: fixed with our discussed solution

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,180 @@
...
+        PCVDeposit storage removePCVDeposit = pcvDeposits[index];
+
+        uint256 depositBalance = removePCVDeposit.usdAmount;
+        uint256 feiDepositBalance = removePCVDeposit.feiAmount;
+
+        balance -= depositBalance;
+        feiReportBalance -= feiDepositBalance;
+        delete pcvDeposits[index];
+
+        emit BalanceUpdate(balance + depositBalance, balance, feiReportBalance + feiDepositBalance, feiReportBalance);
```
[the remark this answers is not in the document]
ElliotFriedman: fixed by using the stack

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: If we do that, it will be more gas inefficient as we will have to write to storage 2 times per deposit instead of the 1 time we currently have to.

test/unit/pcv/NamedStaticPCVDepositWrapper.test.ts
ElliotFriedman: fixed

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: fixed

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: fixed

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: added

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: done

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: fixed

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: added

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /// @notice function to bulk add deposits
+    function bulkAddDeposits(
+        DepositInfo[] calldata newPCVDeposits
+    ) external onlyGovernorOrAdmin {
+        for (uint256 i = 0; i < newPCVDeposits.length; i++) {
+            _addDeposit(newPCVDeposits[i]);
+        }
+    }
```
Joeysantoro, line 148: Add a bulk remove method and we good I think

Joeysantoro (dismissed):

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: The reason this doesn't work is that if you pass an array of indexes to remove, by the time you remove the first one, all of the other ones will be incorrect. The only bulk remove method that works would be one that deletes all items.

ditchfieldcaleb: ++

this isn't simple though

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,202 @@
...
+
+    /// @notice function to edit an existing deposit
+    function editDeposit(
+        uint256 index,
```
ditchfieldcaleb, line 152: Really don't like the fact that the only identifier here is index. Would be super easy to accidentally edit the wrong deposit.

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,202 @@
...
+        return (balance, feiReportBalance);
+    }
+
+    /// @notice display the related token of the balance reported
+    function balanceReportedIn() public pure override returns (address) {
+        return Constants.USD;
+    }
+    
+    /// @notice function to return all of the different tokens deposited into this contract
+    function getAllUnderlying() public view returns (address[] memory) {
```
ditchfieldcaleb, line 192: This will return duplicates.
Suggest building an enumerableSet from the addresses and then using the method on enumerableSet to return all elements in the set - this will dedupe for you.

test/unit/pcv/NamedStaticPCVDepositWrapper.test.ts
```diff
@@ -0,0 +1,302 @@
...
+        usdAmount: balance, /// USD equivalent in this deposit, not including FEI value
+        feiAmount: feiBalance, /// amount of FEI in this deposit
+        underlyingTokenAmount: underlyingTokenAmount, /// amount of underlying token in this deposit
+        underlyingToken: await core.fei()
+      }
+    ]);
+  });
+
+  describe('init', function () {
+    it('reported in USD', async function () {
```
ditchfieldcaleb, line 52: reported in USD --> balance is correctly reported in USD

test/unit/pcv/NamedStaticPCVDepositWrapper.test.ts
```diff
@@ -0,0 +1,302 @@
...
+      expect(await deposit.balanceReportedIn()).to.be.equal('0x1111111111111111111111111111111111111111');
+    });
+
+    it('depositName', async function () {
```
ditchfieldcaleb, line 56: depositName --> depositName is correct

test/unit/pcv/NamedStaticPCVDepositWrapper.test.ts
```diff
@@ -0,0 +1,302 @@
...
+      const { depositName } = await deposit.pcvDeposits(0);
+      expect(depositName).to.be.equal(newDepositName);
+    });
+
+    it('usdAmount', async function () {
```
ditchfieldcaleb, line 61: Generally just make these more descriptive.
On the cli, reading "NamedStaticPCVDepositWrapper init: usdAmount failed, reason: expected (somenumber) to equal (someothernumber)" doesn't tell you what the error is and you have to pull up the code to find out"

Reading "NamedStaticPCVDepositWrapper init: usdAmount should equal balance" does

test/unit/pcv/NamedStaticPCVDepositWrapper.test.ts
```diff
@@ -0,0 +1,302 @@
...
+
+      const resistantBalances = await deposit.resistantBalanceAndFei();
+
+      expect(resistantBalances[0]).to.be.equal(balance);
+      expect(resistantBalances[1]).to.be.equal(feiBalance);
+    });
+  });
+
+  describe('addDeposit', function () {
+    it('add new deposit', async function () {
```
ditchfieldcaleb, line 103: "successfully adds a new deposit"

(again, entire test string should read like natural language)

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: It could work if the array of indexes was properly ordered from greatest indexes to least, but if unordered or not sorted, it does not work.

ditchfieldcaleb (dismissed):
Three main things:

1) see my idea for a uint256-type UUID for the set; if we implement this, we'll need to also ensure that the hash-params of the pcv deposit are immutable; alternatively we can just use a random UUID (but I do like the deterministic approach)

2) error messages should be readable in natural-language-style and understandable as to what exactly they check against

3) make sure that array index parameters are checked for out-of-bounds issues

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,202 @@
...
+            _addDeposit(newPCVDeposits[i]);
+        }
+    }
+
+    // ----------- Helper methods to change state -----------
+
+    /// @notice helper method to delete a PCV deposit
+    function _removeDeposit(uint256 index) internal {
+        if (pcvDeposits.length == 0) { return; }
+
```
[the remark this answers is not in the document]
ElliotFriedman: I agree I can make the change

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,202 @@
...
+        uint256 depositBalance = removePCVDeposit.usdAmount;
+        uint256 feiDepositBalance = removePCVDeposit.feiAmount;
+
+        uint256 oldBalance = balance;
+        uint256 oldFeiReportBalance = feiReportBalance;
+
+        uint256 lastIndex = pcvDeposits.length - 1;
+
+        if (lastIndex != index) {
+            DepositInfo storage lastvalue = pcvDeposits[lastIndex];
```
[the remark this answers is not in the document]
ElliotFriedman: It's the same amount of gas because either way you are pointing to the same storage slot

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: Disagree. if you're editing the deposits, you are going to query them beforehand, ensure you are modifying the correct index, and then submit your DAO action.

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: This is pretty easy to implement, is this the right decision to make? My concerns are, you are trading off spending more on gas for something that can be filtered out client side.

I could add a view function `getAllUnderlyingUnique` that returns only unique underlying addresses and uses a double for loop to filter out duplicates.

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,202 @@
...
+        address underlyingToken; /// address of the underlying token this deposit is reporting
+    }
+
+    /// @notice a list of all pcv deposits
+    DepositInfo[] public pcvDeposits;
+
+    /// @notice the PCV balance
+    uint256 public override balance;
+
+    /// @notice the reported FEI balance
```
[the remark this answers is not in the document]
ElliotFriedman: added more context

test/unit/pcv/NamedStaticPCVDepositWrapper.test.ts
ElliotFriedman: Not sure why we need to change the names on these init functions as this has been convention to do function name for init.

test/unit/pcv/NamedStaticPCVDepositWrapper.test.ts
ElliotFriedman: If you would like to make a change to our unit testing conventions, please suggest a change here and we can review as a team: https://docs.google.com/document/d/1iDjcVXTc8vM9TSLHLSQJU7ZB9MfhlCmRs7YF1Gnuc2Y/edit

test/unit/pcv/NamedStaticPCVDepositWrapper.test.ts
ElliotFriedman: If you would like to make a change to our unit testing conventions, please suggest a change here and we can review as a team: https://docs.google.com/document/d/1iDjcVXTc8vM9TSLHLSQJU7ZB9MfhlCmRs7YF1Gnuc2Y/edit

ditchfieldcaleb: This isn't a change to our testing conventions; this is just good code style

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,201 @@
...
+
+        DepositInfo storage updatePCVDeposit = pcvDeposits[index];
+
+        uint256 oldBalance = balance;
+        uint256 oldFEIBalance = feiReportBalance;
+
+        balance = balance - updatePCVDeposit.usdAmount + usdAmount;
+        feiReportBalance = feiReportBalance - updatePCVDeposit.feiAmount + feiAmount;
+
+        updatePCVDeposit.usdAmount = usdAmount + updatePCVDeposit.usdAmount;
```
Joeysantoro, line 95: is the feiAmount getting updated? If no we need to fix and add a unit test

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,201 @@
...
+    ) internal {
+        require(index < pcvDeposits.length, "NamedStaticPCVDepositWrapper: cannot edit index out of bounds");
+
+        DepositInfo storage updatePCVDeposit = pcvDeposits[index];
+
+        uint256 oldBalance = balance;
+        uint256 oldFEIBalance = feiReportBalance;
+
+        balance = balance - updatePCVDeposit.usdAmount + usdAmount;
+        feiReportBalance = feiReportBalance - updatePCVDeposit.feiAmount + feiAmount;
```
Joeysantoro, line 93: nit: can store the intermediate sums in a local var to reuse

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -0,0 +1,201 @@
...
+        // add all pcv deposits
+        for (uint256 i = 0; i < newPCVDeposits.length; i++) {
+            _addDeposit(newPCVDeposits[i]);
+        }
+    }
+
+    // ----------- Helper methods to change state -----------
+
+    /// @notice helper method to add a PCV deposit
+    function _addDeposit(DepositInfo memory newPCVDeposit) internal {
```
Joeysantoro, line 62: Should we store a mapping of names -> DepositInfo for lookups? Can also be used to ensure that no duplicate names are added to @ditchfieldcaleb 's earlier concerns

Joeysantoro (changes requested):
Main issue is missing feiAmount in edit

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: Good catch!

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: We only use this value once so no need to store intermediately.

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: We shouldn't need to think adversarially about deposit names as only OA and the governor can update the list and guardian can remove deposits instantly.

If we were to do this feature, the data structure would be name -> DepositInfo Index so that we aren't storing each struct and it's data twice.

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: moved to a cached system to save 2 SLOADs when emitting event

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
```diff
@@ -88,9 +88,11 @@ contract NamedStaticPCVDepositWrapper is IPCVDepositBalances, CoreRef {
 
         uint256 oldBalance = balance;
         uint256 oldFEIBalance = feiReportBalance;
+        uint256 newBalance = balance - updatePCVDeposit.usdAmount + usdAmount;
+        uint256 newFeiReportBalance = feiReportBalance - updatePCVDeposit.feiAmount + feiAmount;
```
Joeysantoro, line 92: Use oldBalance and oldFeiBalance to save another sload

Joeysantoro (approved):
One nit and LGTM

contracts/pcv/utils/NamedStaticPCVDepositWrapper.sol
ElliotFriedman: added

Joeysantoro (approved):

PR #389, opened 2021-12-17
Meta-governance update

eswak:
The purpose of this PR is to enable meta-governance with the protocol-owned governance tokens :
- stkAAVE / AAVE
- COMP
- ANGLE
- CRV/CVX
- TOKE (reactor voting)
- INDEX (already live)

BAL tokens are not included in this PR, because our BAL are paired in the BAL/WETH pool.

A forum thread will be created to discuss the initial delegates and present/discuss the update with the broader community.

This pull-request adds several contracts, which are described below.

# ERC20PermissionlessMover
This contract allows permissionless movements of an ERC20 held on a contract to somewhere else. The whitelist is defined by the DAO, to know which token can be moved from where to where. We need such a contract, because most rewards accrue on the PCVDeposits themselves (they provide liquidity, so they earn the rewards), but we want to separate the voting/governance logic from the funds management logic. Therefore, we need a way to move ERC20s from one place to another, without having to go through a DAO vote every time. The use of a whitelist allows anyone to move these reward tokens from the liquidity-providing PCVDeposits to the protocol contracts that can actually delegate voting power and handle meta-governance.

# SnapshotDelegatorPCVDeposit
This contract existed already, I just moved it to a new `metagov` folder. The constructor also now include the setting of a contract admin role `_setContractAdminRole(keccak256("METAGOV_ADMIN_ROLE"));`, to enable OA to change delegates.

Used for `INDEX` (already deployed on-chain)

# DelegatorPCVDeposit
This contract is similar to the `SnapshotDelegatorPCVDeposit` but calls `token.delegate(address)` instead of using the Snapshot registry. If a token supports `delegate(address)` call, it's better to use this deposit, because usually the delegation works on Snapshot too.

Used for `COMP`.

# AaveDelegatorPCVDeposit
This contract is a `SnapshotDelegatorPCVDeposit` that can also wrap/unwrap between `AAVE` and `stkAAVE`, and claim rewards on Aave. We have to get in touch with the Aave team to allow this contract to claim `stkAAVE` rewards on behalf of our `RAI` and `ETH` PCVDeposits on Aave.

Used for `stkAAVE`.

# ConvexDelegatorPCVDeposit
This contract is a `SnapshotDelegatorPCVDeposit` that can also "play the Convex game" - I'll paste the natspec of the contract : 
```
/// @title Convex Delegator PCV Deposit
/// This contract is designed to be whitelisted as the destination for CVX and CRV
/// rewards earned by various Curve and Convex PCVDeposits.
/// CRV are converted to cvxCRV and staked on Convex. They earn :
///   - CRV rewards, that are cycled back in the same circuit
///   - CVX rewards, that are locked for voting (see following paragraph)
///   - 3crv rewards, that accrue on this contract & should be sent somewhere
///     else, for instance a PCVDeposit that deposit them with FEI in the Curve
///     metapool [3crv, FEI], or a deposit that cash them out for DAI.
/// CVX are vote-locked for 16 weeks, to be able to vote in Convex governance,
/// and earn cvxCRV rewards + eventual bribes. The cvxCRV rewards can be staked
/// to enter the CRV circuit described above (earn CRV/CVX/3crv). The CVX vote
/// for gauge weights happen biweekly through Snapshot (vote.convexfinance.com).
/// @author Fei Protocol
```

Used for `CVX` (and `CRV` that are cycled to more `CVX`)

# AngleDelegatorPCVDeposit
This contract is a `SnapshotDelegatorPCVDeposit` that can also vote-escrow tokens (`VoteEscrowTokenManager`), vote and propose on an OZ governor (`OZGovernorVoter`), and stake tokens in gauges and claim their rewards (`LiquidityGaugeManager`).

Used for `ANGLE`.

# Workflow
- [x] Initial implementation
- [ ] First review / adjustments
- [ ] Unit/Integration tests
- [ ] Community talk
- [ ] Snapshot
- [ ] On-chain vote


contracts/metagov/AaveDelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,95 @@
...
+/// @author Fei Protocol
+contract AaveDelegatorPCVDeposit is SnapshotDelegatorPCVDeposit {
+
+    /// @notice the Aave Incentives Controller.
+    IAaveIncentivesController public constant aaveIncentivesController = IAaveIncentivesController(address(0xd784927Ff2f95ba542BfC824c8a8a98F3495f6b5));
+
+    /// @notice the AAVE token address.
+    IERC20 public constant aave = IERC20(address(0x7Fc66500c84A76Ad7e9c93437bFc5Ac33E2DDaE9));
+    /// @notice the stkAAVE token address.
+    IStkAave public constant stkaave = IStkAave(address(0x4da27a545c0c5B758a6BA100e3a049001de870f5));
```
ElliotFriedman, line 32: No need to cast to an address, then cast to an interface. You can just remove the cast to address.

contracts/metagov/ConvexDelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,139 @@
...
+/// CVX are vote-locked for 16 weeks, to be able to vote in Convex governance,
+/// and earn cvxCRV rewards + eventual bribes. The cvxCRV rewards can be staked
+/// to enter the CRV circuit described above (earn CRV/CVX/3crv). The CVX vote
+/// for gauge weights happen biweekly through Snapshot (vote.convexfinance.com).
+/// @author Fei Protocol
+contract ConvexDelegatorPCVDeposit is SnapshotDelegatorPCVDeposit {
+
+    /// @notice the CvxLocker contract to lock CVX tokens on Convex and earn
+    /// rewards / be eligible for voting in the Convex governance.
+    ICvxLocker public constant cvxLocker = ICvxLocker(address(0xD18140b4B819b895A3dba5442F959fA44994AF50));
```
ElliotFriedman, line 45: remove cast to address

contracts/metagov/DelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,75 @@
...
+        onlyPCVController
+    {
+        _withdrawERC20(address(token), to, amountUnderlying);
+    }
+
+    /// @notice no-op
+    function deposit() external override {}
+
+    /// @notice returns total balance of PCV in the Deposit
+    function balance() public view override returns (uint256) {
```
ElliotFriedman, line 52: Does this contract need to override the resistantBalanceAndFei method? I'm not sure if any of these contracts will hold FEI in them, but if they do, we should override

contracts/pcv/utils/ERC20PermissionlessMover.sol
```diff
@@ -0,0 +1,64 @@
...
+        address _oldTo = whitelist[_token][_from];
+        whitelist[_token][_from] = address(0);
+
+        emit WhitelistRemove(_token, _from, _oldTo);
+    }
+
+    /// @notice Move a token from a deposit to a whitelisted destination
+    /// @param _token to transfer
+    /// @param _from source of the token
+    function move(address _token, address _from) public whenNotPaused {
```
ElliotFriedman, line 55: if this is permissionless, we would be open to grief attacks with this architecture as anyone can move assets out of PCVDeposits into others without any conditionals being met other than this contract not being paused.

eswak: I think that's fine, because we'd usually not allow movement of an ERC20 on a deposit that does its accounting in it, e.g. the ETH Compound PCVDeposit will have a whitelisted movement of COMP, not a whitelisted movement of WETH, so these tokens are not accounted for in the PCV anyway

contracts/metagov/DelegatorPCVDeposit.sol
eswak: they should not hold FEI

PR #399, opened 2021-12-20
set types to interfaces

ditchfieldcaleb:


types/types.ts
```diff
@@ -225,7 +225,7 @@ export interface MainnetContracts {
   feiDAO: FeiDAO;
   autoRewardsDistributor: AutoRewardsDistributor;
   rewardsDistributorAdmin: RewardsDistributorAdmin;
-  restrictedPermissions: RestrictedPermissions;
+  restrictedPermissions: Contract;
```
Joeysantoro, line 228: why rename this?

ditchfieldcaleb: reverted

Joeysantoro (approved):

PR #404, opened 2021-12-21
Create STWBulkHarvest Contract and Deploy Script

ElliotFriedman:
Create stateless contract to allow defender autotask to harvest multiple STW's quickly and in a gas efficient manner. Previously, when calling multiple STW's in a single script, defender would drop tx's as it manually would set each tx to the same nonce. This eliminates that problem and is more gas efficient because each call only has 1 signature verification cost. 

First harvest tx: https://etherscan.io/tx/0x95b807e4cc4853ccc4b81ec07bba3a2869a6dead748ec9f120322401e5651680


contracts/staking/STWBulkHarvest.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity ^0.8.4;
+
+import "./StakingTokenWrapper.sol";
+
+/// @notice contract to bulk harvest multiple Staking Token Wrappers in a single transaction
+/// stateless contract with no storage and can only call harvest on the STW's
+contract STWBulkHarvest {
+    function bulkHarvest(StakingTokenWrapper[] calldata stw) external {
```
Joeysantoro, line 8: It would be cool to expose a method for harvesting a canonical list from storage too I think

ElliotFriedman: that costs way more gas because you have to do a SLOAD for each address you pull out of storage.

ElliotFriedman: If it pulls from storage, then you would need access controls, and have to permission functions to append or delete from the list. This contract should be stateless to keep it simple.

eswak (approved):
cool idea

contract-addresses/mainnetAddresses.ts
```diff
@@ -1311,6 +1311,11 @@ const MainnetAddresses: MainnetAddresses = {
     address: '0x27Fae9E49AD955A24bB578B66Cdc962b5029fbA9',
     category: AddressCategory.Deprecated
   },
+  stwBulkHarvest: {
```
Joeysantoro, line 1314: Please keep these ordered by category + alphabetical. This belongs in the Rewards section

Joeysantoro (changes requested):

Joeysantoro (approved):

PR #407, opened 2021-12-22
dependencies

Joeysantoro:


contract-addresses/dependencies.ts
```diff
@@ -1,5 +1,564 @@
...
+    contractDependencies: []
+  },
+  optimisticTimelock: {
+    contractDependencies: ['core', 'rewardsDistributorAdmin', 'tribalChiefSync', 'rariPool8Comptroller']
+  },
+  rariTimelock: {
+    contractDependencies: []
+  },
+  tribeRariDAO: {
+    contractDependencies: []
```
ElliotFriedman, line 146: isn't this dependent on the tribe token?

contract-addresses/dependencies.ts
```diff
@@ -1,5 +1,564 @@
...
+    contractDependencies: ['core']
+  },
+  daiBondingCurve: {
+    contractDependencies: ['core']
+  },
+  daiPCVDripController: {
+    contractDependencies: ['core']
+  },
+  daiPSM: {
+    contractDependencies: ['core']
```
ElliotFriedman, line 164: this is also dependent on the compoundPCVDripController

contract-addresses/dependencies.ts
```diff
@@ -1,5 +1,564 @@
...
+  },
+  ethReserveStabilizer: {
+    contractDependencies: ['core']
+  },
+  tribeReserveStabilizer: {
+    contractDependencies: ['core']
```
ElliotFriedman, line 170: wouldn't both reserve stabilizers be dependent on their respective drippers?

contract-addresses/dependencies.ts
```diff
@@ -1,5 +1,735 @@
...
+    contractDependencies: ['core', 'fei', 'aaveEthPCVDeposit', 'ethReserveStabilizer']
+  },
+  bondingCurve: {
+    contractDependencies: ['core', 'fei', 'aaveEthPCVDeposit', 'compoundEthPCVDeposit', 'chainlinkEthUsdOracleWrapper']
+  },
+  compoundEthPCVDripController: {
+    contractDependencies: ['core', 'fei', 'compoundEthPCVDeposit', 'ethReserveStabilizer']
+  },
+  daiBondingCurve: {
+    contractDependencies: ['core', 'fei', 'compoundDaiPCVDeposit', 'chainlinkDaiUsdOracleWrapper']
```
ElliotFriedman, line 231: because this is deprecated, can't we just remove?

contract-addresses/dependencies.ts
```diff
@@ -1,5 +1,735 @@
...
+    contractDependencies: ['core', 'tribeUsdCompositeOracle', 'tribeMinter', 'collateralizationOracleWrapper', 'tribe']
+  },
+  aaveEthPCVDeposit: {
+    contractDependencies: ['core', 'aaveEthPCVDripController', 'bondingCurve', 'pcvGuardian']
+  },
+  aaveFeiPCVDeposit: {
+    contractDependencies: ['core', 'fei']
+  },
+  aaveRaiPCVDeposit: {
+    contractDependencies: ['core']
```
ElliotFriedman, line 265: this should also be dependent on the aRAI token

contract-addresses/dependencies.ts
```diff
@@ -1,5 +1,735 @@
...
+    ]
+  },
+  tribeReserveStabilizer: {
+    contractDependencies: ['core', 'tribeUsdCompositeOracle', 'tribeMinter', 'collateralizationOracleWrapper', 'tribe']
+  },
+  aaveEthPCVDeposit: {
+    contractDependencies: ['core', 'aaveEthPCVDripController', 'bondingCurve', 'pcvGuardian']
+  },
+  aaveFeiPCVDeposit: {
+    contractDependencies: ['core', 'fei']
```
ElliotFriedman, line 262: this should also be dependent on the aFEI token

contract-addresses/dependencies.ts
```diff
@@ -1,5 +1,735 @@
...
+  },
+  rariPool79FeiPCVDeposit: {
+    contractDependencies: ['core', 'fei']
+  },
+  rariPool7FeiPCVDeposit: {
+    contractDependencies: ['core', 'fei']
+  },
+  rariPool7LusdPCVDeposit: {
+    contractDependencies: ['core']
```
ElliotFriedman, line 358: missing lusd, and the corresponding fToken

ElliotFriedman:
It seems that this PR does not factor into account external dependencies which I would think it should.

contract-addresses/dependencies.ts
Joeysantoro: aRAI is external

contract-addresses/dependencies.ts
Joeysantoro: not deprecated yet

contract-addresses/dependencies.ts
Joeysantoro: aFEI is external

contract-addresses/dependencies.ts
Joeysantoro: both external

Joeysantoro:
> It seems that this PR does not factor into account external dependencies which I would think it should.

disagree. Maybe we can add external later but this PR is already unwieldy. The point is to keep a mental map of internal contract dependencies primarily. The extra effort to add external deps isn't really worth it. I think it should be merged as-is.

ditchfieldcaleb (dismissed):

PR #411, opened 2021-12-23
Votium Briber adapter

eswak:
Add a new VotiumBriber contract that can collect a rewards stream from the Tribal Chief and bribe with it.

I only did e2e tests because unit tests would just be testing my mock behaviors.

Open question : who do we empower to make the bribe calls ? Could be the OA multisig (not timelock), one of our addresses, a new multisig, a new multisig with timelock... Let's discuss on the forum & on discord.


Joeysantoro:
I changed it to TRIBAL_CHIEF_ADMIN_ROLE for now and we can have OA do it.

test/integration/tests/votium-bribe.ts
```diff
@@ -0,0 +1,104 @@
...
+      );
+    });
+
+    it('should allow VotiumBriber to bribe', async function () {
+      // Should receive a non-zero rewards stream from TribalChief
+      const briberBalanceBefore = await contracts.tribe.balanceOf(contracts.votiumBriberD3pool.address);
+      await contracts.stakingTokenWrapperBribeD3pool.harvest();
+      const briberBalanceAfter = await contracts.tribe.balanceOf(contracts.votiumBriberD3pool.address);
+      const bribeAmount = briberBalanceAfter.sub(briberBalanceBefore);
+      expect(bribeAmount).to.be.at.least('1');
```
ElliotFriedman, line 94: There should be a higher quality test here that does a bit more than just checking the bribe amount is at least 1. With current allocation points, it should be thousands of tribe if you have 250 AP

contracts/pcv/convex/VotiumBriber.sol
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /// @notice Spend tokens on Votium to bribe for a given pool.
+    /// @param _proposal id of the proposal on snapshot
+    /// @param _choiceIndex index of the pool in the snapshot vote to vote for
+    /// @dev the call will revert if Votium has not called initiateProposal with
+    /// the _proposal ID, if _choiceIndex is out of range, or of block.timestamp
+    /// is after the deadline for bribing (usually 6 hours before Convex snapshot
+    /// vote ends).
+    function bribe(bytes32 _proposal, uint256 _choiceIndex) public onlyGovernorOrAdmin whenNotPaused {
```
ElliotFriedman, line 46: This contract needs some unit tests.

ElliotFriedman: It looks fine, but we should always show our work with IT + Unit tests

ElliotFriedman (dismissed):

contracts/pcv/convex/VotiumBriber.sol
Joeysantoro: Don't think it needs unit tests they wouldn't be testing much.

ditchfieldcaleb: Gonna agree with @ElliotFriedman here, we should definitely unit test most of anything new, especially if the function sends value anywhere.

ditchfieldcaleb (dismissed):

contracts/pcv/convex/VotiumBriber.sol
ElliotFriedman: I agree that there isn't much to test, but just because there is a small amount to test, doesn't mean we shouldn't test it.

If it's a small amount of code, we should be happy we don't have tons to test and just write the few cases that are needed.

ElliotFriedman (approved):
I would add an IT that tests the pausing. Other than that, LGTM!

ditchfieldcaleb (approved):

PR #416, opened 2021-12-25
proposal number check

Joeysantoro:
adds proposal id to proposals_config and uses this to streamline check-proposal and e2e flows


ElliotFriedman (approved):
LGTM!

PR #419, opened 2021-12-27
FIP-61 maintenance upgrades

Joeysantoro:
Adds an auto decrease for the TribalChief

https://www.diffchecker.com/1IrDVILm diff between two files


contracts/staking/TribalChiefSyncV2.sol
```diff
@@ -0,0 +1,191 @@
...
+
+    function isRewardDecreaseAvailable() public view returns(bool) {
+        return nextRewardTimestamp() < block.timestamp;
+    }
+
+    function nextRewardTimestamp() public view returns(uint256) {
+        return rewardsSchedule[nextRewardsRate()];
+    }
+
+    function nextRewardsRate() public view returns(uint256) {
```
ditchfieldcaleb, line 107: Consider adding an explicit array-index guard here, with a custom revert message. Helps with debugging.

ditchfieldcaleb (dismissed):

contracts/staking/TribalChiefSyncV2.sol
```diff
@@ -0,0 +1,191 @@
...
+        return rewardsArray[rewardsArray.length - 1];
+    }
+
+    /// @notice Sync a rewards rate change
+    function decreaseRewards(uint256 tribePerBlock, bytes32 salt) external update {
```
ElliotFriedman, line 112: prefer inheriting existing TribalChiefSync instead of copying this code for the last 4 functions.

proposals/description/fip_61.ts
```diff
@@ -0,0 +1,87 @@
...
+      values: '0',
+      method: 'becomeAdmin()',
+      arguments: [],
+      description: 'Become timelock admin'
+    },
+    {
+      target: 'optimisticTimelock',
+      values: '0',
+      method: 'grantRole(bytes32,address)',
+      arguments: ['0xd8aa0f3194971a2a116679f7c2090f6939c8d4e01a2a8d7e41d55e5351469e63', '{tribalChiefSyncV2}'],
```
ElliotFriedman, line 52: would be nice to be able to do substitution for these roles so instead of putting illegible 64 char string, we just put '{EXECUTOR_ROLE}' and under the hood it gets replaced for the corresponding bytes32

contracts/staking/TribalChiefSyncV2.sol
ElliotFriedman: I would agree if this function changed state, but for a view only function it's not needed

Joeysantoro: I added a check inside isRewardDecreaseAvailable

contracts/staking/TribalChiefSyncV2.sol
Joeysantoro: Plan on deprecating old contract. Rather not inherit for readabiity

proposals/description/fip_61.ts
Joeysantoro: Make a ticket

contracts/staking/TribalChiefSyncV2.sol
```diff
@@ -0,0 +1,191 @@
...
+ */
+contract TribalChiefSyncV2 {
+    ITribalChief public immutable tribalChief;
+    IAutoRewardsDistributor public immutable autoRewardsDistributor;
+    ITimelock public immutable timelock;
+
+    /// @notice a mapping from reward rates to timestamps after which they become active
+    mapping(uint256 => uint256) public rewardsSchedule;
+
+    uint256[] internal rewardsArray;
```
ElliotFriedman, line 38: prefer making this public so it's easy to pull all values out of the rewardsSchedule mapping

contracts/staking/TribalChiefSyncV2.sol
ditchfieldcaleb: @ElliotFriedman it doesn't change state directly, but a function that *does* change state relies on it.

ditchfieldcaleb (approved):

PR #423, opened 2021-12-30
deprecated test

Joeysantoro:
Adds deprecated contracts signoff category for proposals_config.

Adds an integration test to require that these contracts are in fact deprecated and not contained anywhere in dependencies map


test/integration/tests/dependencies.ts
```diff
@@ -23,7 +23,17 @@ describe('e2e-dependencies', function () {
         for (let j = 0; j < contracts.length; j++) {
           const contract = contracts[j];
           const category = addresses[contract].category;
-          if (category === 'External' || category === 'Deprecated') {
+          if (category === 'External') {
```
ditchfieldcaleb, line 26: You should make an enum for these categories in the types file and import it here instead of using string literals

ditchfieldcaleb: (and also probably use said enum everywhere else if you can do that without manually editing 500+ lines of code)

test/integration/tests/dependencies.ts
```diff
@@ -37,6 +47,34 @@ describe('e2e-dependencies', function () {
...
+          const contract = deprecated[j];
+          const category = addresses[contract].category;
+          expect(category).to.be.equal('Deprecated');
+
+          expect(contracts).to.not.contain(contract);
+        }
+      }
+    });
+
     it('all dependencies bidirectional', async function () {
```
ditchfieldcaleb, line 78: I know this wasn't an edit on this pr but the string "Check dependencies all dependencies bidirectional" doesn't tell me what it's checking, maybe something like "Check dependencies are listed bidirectionally" would?

ditchfieldcaleb (changes requested):

test/integration/tests/dependencies.ts
Joeysantoro: it already was an enum. Used it here

test/integration/tests/dependencies.ts
Joeysantoro: changed this and others

ditchfieldcaleb (approved):

PR #430, opened 2022-01-02
FIP-60 - Fuse guardian

Joeysantoro:
Creates a Fuse pause/borrow cap admin wrapper which gives access to Guardian and TribalChiefAdmins the ability to pause pools and set caps.

It also includes a utility for pausing/capping by underlying rather than the cToken. This is important because the cToken address cannot be pre-computed so this feature allows atomic deployment/pausing/capping.

### DIFF
--- a/contracts/feirari/FuseGuardian.sol
+++ b/contracts/feirari/FuseGuardian.sol
@@ -1,0 +1,109 @@
+pragma solidity ^0.8.0;
+
+import "../refs/CoreRef.sol";
+import "../external/Unitroller.sol";
+
+/// @title a Fuse pause and borrow cap guardian used to expand access control to more Fei roles
+/// @author joeysantoro
+contract FuseGuardian is CoreRef {
+
+    /// @notice the fuse comptroller
+    Unitroller public immutable comptroller;
+
+    /// @param _core address of core contract
+    /// @param _comptroller the fuse comptroller
+    constructor(
+        address _core,
+        Unitroller _comptroller
+    ) CoreRef(_core) {
+        comptroller = _comptroller;
+        /// @notice The reason we are reusing the tribal chief admin role is it consolidates control in the OA,
+        /// and means we don't have to do another governance action to create this role in core
+        _setContractAdminRole(keccak256("TRIBAL_CHIEF_ADMIN_ROLE"));
+    }
+
+    // ************ BORROW GUARDIAN FUNCTIONS ************
+    /**
+      * @notice Set the given supply caps for the given cToken markets. Supplying that brings total underlying supply to or above supply cap will revert.
+      * @dev Admin or borrowCapGuardian function to set the supply caps. A supply cap of 0 corresponds to unlimited supplying.
+      * @param cTokens The addresses of the markets (tokens) to change the supply caps for
+      * @param newSupplyCaps The new supply cap values in underlying to be set. A value of 0 corresponds to unlimited supplying.
+      */
+    function _setMarketSupplyCaps(CToken[] memory cTokens, uint[] memory newSupplyCaps) public isGovernorOrGuardianOrAdmin {
+        comptroller._setMarketSupplyCaps(cTokens, newSupplyCaps);
+    }
+
+    function _setMarketSupplyCapsByUnderlying(address[] calldata underlyings, uint[] calldata newSupplyCaps) external {
+        _setMarketSupplyCaps(_underlyingToCTokens(underlyings), newSupplyCaps);
+    }
+
+    function _underlyingToCTokens(address[] memory underlyings) internal view returns (CToken[] memory) {
+        CToken[] memory cTokens = new CToken[](underlyings.length);
+        for (uint256 i = 0; i < underlyings.length; i++) {
+            address cToken = comptroller.cTokensByUnderlying(underlyings[i]);
+            require(cToken != address(0), "cToken doesn't exist");
+            cTokens[i] = CToken(cToken);
+        }
+        return cTokens;
+    }
+
+    /**
+      * @notice Set the given borrow caps for the given cToken markets. Borrowing that brings total borrows to or above borrow cap will revert.
+      * @dev Admin or borrowCapGuardian function to set the borrow caps. A borrow cap of 0 corresponds to unlimited borrowing.
+      * @param cTokens The addresses of the markets (tokens) to change the borrow caps for
+      * @param newBorrowCaps The new borrow cap values in underlying to be set. A value of 0 corresponds to unlimited borrowing.
+      */
+    function _setMarketBorrowCaps(CToken[] memory cTokens, uint[] memory newBorrowCaps) public isGovernorOrGuardianOrAdmin {
+        comptroller._setMarketBorrowCaps(cTokens, newBorrowCaps);
+    }
+
+    function _setMarketBorrowCapsByUnderlying(address[] calldata underlyings, uint[] calldata newBorrowCaps) external {
+        _setMarketBorrowCaps(_underlyingToCTokens(underlyings), newBorrowCaps);
+    }
+
+    /**
+     * @notice Admin function to change the Borrow Cap Guardian
+     * @param newBorrowCapGuardian The address of the new Borrow Cap Guardian
+     */
+    function _setBorrowCapGuardian(address newBorrowCapGuardian) external onlyGovernor {
+        comptroller._setBorrowCapGuardian(newBorrowCapGuardian);
+    }
+
+    // ************ PAUSE GUARDIAN FUNCTIONS ************
+    /**
+     * @notice Admin function to change the Pause Guardian
+     * @param newPauseGuardian The address of the new Pause Guardian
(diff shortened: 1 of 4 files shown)


contracts/feirari/FuseGuardian.sol
```diff
@@ -0,0 +1,109 @@
...
+        comptroller._setMarketSupplyCaps(cTokens, newSupplyCaps);
+    }
+
+    function _setMarketSupplyCapsByUnderlying(address[] calldata underlyings, uint[] calldata newSupplyCaps) external {
+        _setMarketSupplyCaps(_underlyingToCTokens(underlyings), newSupplyCaps);
+    }
+
+    function _underlyingToCTokens(address[] memory underlyings) internal view returns (CToken[] memory) {
+        CToken[] memory cTokens = new CToken[](underlyings.length);
+        for (uint256 i = 0; i < underlyings.length; i++) {
```
transmissions11, line 42: unchecked

contracts/feirari/FuseGuardian.sol
```diff
@@ -0,0 +1,109 @@
...
+            address cToken = comptroller.cTokensByUnderlying(underlyings[i]);
+            require(cToken != address(0), "cToken doesn't exist");
+            cTokens[i] = CToken(cToken);
+        }
+        return cTokens;
```
transmissions11, line 47: if u used a named return arg u could just assign cTokens to it

contracts/feirari/FuseGuardian.sol
```diff
@@ -0,0 +1,109 @@
...
+      */
+    function _setMarketSupplyCaps(CToken[] memory cTokens, uint[] memory newSupplyCaps) public isGovernorOrGuardianOrAdmin {
+        comptroller._setMarketSupplyCaps(cTokens, newSupplyCaps);
+    }
+
+    function _setMarketSupplyCapsByUnderlying(address[] calldata underlyings, uint[] calldata newSupplyCaps) external {
+        _setMarketSupplyCaps(_underlyingToCTokens(underlyings), newSupplyCaps);
+    }
+
+    function _underlyingToCTokens(address[] memory underlyings) internal view returns (CToken[] memory) {
```
transmissions11, line 40: this is only used by functions that take calldata so could make calldata

contracts/feirari/FuseGuardian.sol
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    // ************ BORROW GUARDIAN FUNCTIONS ************
+    /**
+      * @notice Set the given supply caps for the given cToken markets. Supplying that brings total underlying supply to or above supply cap will revert.
+      * @dev Admin or borrowCapGuardian function to set the supply caps. A supply cap of 0 corresponds to unlimited supplying.
+      * @param cTokens The addresses of the markets (tokens) to change the supply caps for
+      * @param newSupplyCaps The new supply cap values in underlying to be set. A value of 0 corresponds to unlimited supplying.
+      */
+    function _setMarketSupplyCaps(CToken[] memory cTokens, uint[] memory newSupplyCaps) public isGovernorOrGuardianOrAdmin {
```
transmissions11, line 32: this is only used by functions that take calldata so could make calldata

contracts/feirari/FuseGuardian.sol
```diff
@@ -0,0 +1,109 @@
...
+        return cTokens;
+    }
+
+    /**
+      * @notice Set the given borrow caps for the given cToken markets. Borrowing that brings total borrows to or above borrow cap will revert.
+      * @dev Admin or borrowCapGuardian function to set the borrow caps. A borrow cap of 0 corresponds to unlimited borrowing.
+      * @param cTokens The addresses of the markets (tokens) to change the borrow caps for
+      * @param newBorrowCaps The new borrow cap values in underlying to be set. A value of 0 corresponds to unlimited borrowing.
+      */
+    function _setMarketBorrowCaps(CToken[] memory cTokens, uint[] memory newBorrowCaps) public isGovernorOrGuardianOrAdmin {
```
transmissions11, line 56: see above

contracts/feirari/FuseGuardian.sol
ditchfieldcaleb: @transmissions11 does this save gas or is it just a visual / code style thing

transmissions11: honestly cant remember, this is why y'all need gas snapshotting infra :)

proposals/description/fip_60.ts
```diff
@@ -44,6 +44,51 @@ const fip_60: ProposalDescription = {
...
+      arguments: ['{fuseGuardian}'],
+      description: 'Set Fuse borrow cap guardian'
+    },
+    {
+      target: 'fuseGuardian',
+      values: '0',
+      method: '_setMarketSupplyCapsByUnderlying(address[],uint256[])',
+      arguments: [
+        ['{curveD3pool}', '{curve3Metapool}', '{gUniFeiDaiLP}'],
+        ['25000000000000000000000000', '25000000000000000000000000', '2500000000000000000000000000']
```
ditchfieldcaleb, line 88: We should come up with a better way to spell out amounts than an inline string, it's bound to cause a mistake eventually

contracts/feirari/FuseGuardian.sol
```diff
@@ -0,0 +1,109 @@
...
+        require(cToken != address(0), "cToken doesn't exist");
+        _setBorrowPaused(CToken(cToken), state);
+    }
+
+    function _setBorrowPaused(CToken cToken, bool state) public isGovernorOrGuardianOrAdmin returns (bool) {
+        return comptroller._setBorrowPaused(cToken, state);
+    }
+
+    function _setTransferPaused(bool state) public isGovernorOrGuardianOrAdmin returns (bool) {
+        return comptroller._setSeizePaused(state);
```
ElliotFriedman, line 103: should call `_setTransferPaused ` not `_setSeizePaused`

contracts/feirari/FuseGuardian.sol
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * @notice Admin function to change the Pause Guardian
+     * @param newPauseGuardian The address of the new Pause Guardian
+     * @return uint 0=success, otherwise a failure. (See enum Error for details)
+     */
+    function _setPauseGuardian(address newPauseGuardian) public onlyGovernor returns (uint) {
+        return comptroller._setPauseGuardian(newPauseGuardian);
+    }
+
+    function _setMintPausedByUnderlying(address underlying, bool state) external {
```
ElliotFriedman, line 82: I don't like how this function calls a public function that has a modifier on it and has no modifier on itself in the function signature. It obfuscates who is allowed to call it.

Prefer splitting out the _setMintPaused into an internal helper function and permissioning both functions.

ElliotFriedman (dismissed):

contracts/feirari/FuseGuardian.sol
Joeysantoro: I'm probably going to leave as-is given this is an admin function and I'd rather optimize for readability than gas

contracts/feirari/FuseGuardian.sol
Joeysantoro: same point

contracts/feirari/FuseGuardian.sol
Joeysantoro: Great callout, fixed

proposals/description/fip_60.ts
Joeysantoro: can probably just do ethers weiperether

Joeysantoro: I do think we're covered by invariant checking in the validate step

contracts/feirari/FuseGuardian.sol
```diff
@@ -0,0 +1,117 @@
...
+      * @notice Set the given supply caps for the given cToken markets. Supplying that brings total underlying supply to or above supply cap will revert.
+      * @dev Admin or borrowCapGuardian function to set the supply caps. A supply cap of 0 corresponds to unlimited supplying.
+      * @param cTokens The addresses of the markets (tokens) to change the supply caps for
+      * @param newSupplyCaps The new supply cap values in underlying to be set. A value of 0 corresponds to unlimited supplying.
+      */
+    function _setMarketSupplyCaps(CToken[] memory cTokens, uint[] calldata newSupplyCaps) public isGovernorOrGuardianOrAdmin {
+        comptroller._setMarketSupplyCaps(cTokens, newSupplyCaps);
+    }
+
+    function _setMarketSupplyCapsByUnderlying(address[] calldata underlyings, uint[] calldata newSupplyCaps) external {
```
ElliotFriedman, line 36: I don't like how this function calls a public function that has a modifier on it and has no modifier on itself in the function signature. It obfuscates who is allowed to call it.

contracts/feirari/FuseGuardian.sol
ElliotFriedman: I've seen gas savings from using a named parameter in the return value

PR #435, opened 2022-01-04
ETH PSM with Specific Redemption Pausing

ElliotFriedman:


contracts/utils/PauserV2.sol
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * @dev Contract module which allows children to implement an emergency stop
+ * mechanism that can be triggered by an authorized account.
+ *
+ * This module is used through inheritance. It will make available the
+ * modifiers `whenNotPaused` and `whenPaused`, which can be applied to
+ * the functions of your contract. Note that they will not be pausable by
+ * simply including this module, only once the modifiers are put in place.
+ */
+abstract contract PauserV2 {
```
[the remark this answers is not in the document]
ElliotFriedman: so just write this logic directly into the EthPSM class?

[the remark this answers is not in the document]
ElliotFriedman: done

contracts/stabilizer/EthPegStabilityModule.sol
```diff
@@ -0,0 +1,71 @@
...
+    function pauseRedeem() public onlyGuardianOrGovernor whileRedemptionsNotPaused {
+        redeemPaused = true;
+        emit RedemptionsPaused(msg.sender);
+    }
+
+    /// @notice set secondary pausable methods to unpaused
+    function unpauseRedeem() public onlyGuardianOrGovernor whileRedemptionsPaused {
+        redeemPaused = false;
+        emit RedemptionsUnpaused(msg.sender);
+    }
```
Joeysantoro, line 61: add isGovernorOrGuardianOrAdmin or whatever. Also should that modifier be renamed to only* instead of is* for consistency

contracts/stabilizer/EthPegStabilityModule.sol
```diff
@@ -0,0 +1,71 @@
...
+contract EthPegStabilityModule is PegStabilityModule {
+
+    /// @notice boolean switch that indicates whether redemptions are paused
+    bool public redeemPaused;
+
+    /// @notice event that is emitted when redemptions are paused
+    event RedemptionsPaused(address account);
+    
+    /// @notice event that is emitted when redemptions are unpaused
+    event RedemptionsUnpaused(address account);
```
Joeysantoro, line 17: I know it changes the base class but I think these should be canon for all PSMs not just ETH

ElliotFriedman: it feels like the scope keeps changing here.

ElliotFriedman: if we do this for all base psms, then they should have separate pause switches for mint and redeem outside of oz

contracts/stabilizer/EthPegStabilityModule.sol
ElliotFriedman: i can add isGovernorOrGuardianOrAdmin, but I can't modify the base CoreRef class modifier without blowing up other contracts.

ElliotFriedman: I could extend coreref to have correct naming on the isGovernorOrGuardianOrAdmin modifier.

contracts/stabilizer/EthPegStabilityModule.sol
ElliotFriedman: that would mean we have 3 separate pieces of code doing pausing and at that point we should just move to function selector pausing.

contracts/stabilizer/EthPegStabilityModule.sol
Joeysantoro: just do isGovernorOrGuardianOrAdmin whatever

contracts/stabilizer/EthPegStabilityModule.sol
Joeysantoro: Don't feel strongly. ETH can be unique

PR #443, opened 2022-01-08
Fei Authorities

Joeysantoro:
Adapt solmate authority for Fei


transmissions11:
this is cool, ur kinda shimming ur existing auth system into the Authority interface tho, this is not rly how its meant to be used

authority is supposed to manage which roles can call which and such, but i get the desire to be heavily opinionated

ditchfieldcaleb (commented):
gonna agree with @transmissions11 here, this isn't very clean
will talk with you irl about how we could improve it

Joeysantoro:
Totally fair, I think we should move to deprecate Core completely and use Solmate the right way. Will close this and go back to drawing board

PR #454, opened 2022-01-10
FIP-62 Dependency IT

ElliotFriedman:


Joeysantoro:
Closed bc I don't think this change is neccessary atm and we'll probably to a larger refactor of the deps system

PR #479, opened 2022-01-19
FIP-68

Joeysantoro:


eswak (approved):
You can also remove this from `proposals_config.ts` if you want, but it can wait, nbd

```
import fip_67 from '@proposals/description/fip_67';
```

PR #490, opened 2022-01-23
ERC20LensFactory

Joeysantoro:


eswak (commented):
Code looks good, you could add an integration test and then 👍

contracts/pcv/utils/ERC20LensFactory.sol
```diff
@@ -0,0 +1,19 @@
+pragma solidity ^0.8.4;
```
thomas-waite, line 1: Minor point, need to add the license statement

thomas-waite (commented):
Also looks good to me once we've added some tests

PR #491, opened 2022-01-24
withdrawal

Joeysantoro:
Adds withdrawal contracts for the following PCV deposits held by the OA Timelock:
* Idle (Tranches and Best Yield)
* BarnBridge
* Kashi (DPI, ETH, TRIBE, XSushi)

It removes them from the CR oracle NamedStaticPCVDeposit along with Visor (which will be removed in another OA action).

The redeemers are completely permissionless, but can only redeem to the Optimistic Timelock for minimal attack surface and max flexibility. They also include a sweeper for gov tokens earned, if any (e.g. IDLE)


eswak (dismissed):

contracts/pcv/barnbridge/SmartYieldRedeemer.sol
```diff
@@ -0,0 +1,46 @@
...
+    ) { 
+        target = _target; 
+        underlying = _underlying;
+    }
+
+    /// @notice redeem BarnBridge SmartYield juniors
+    /// @param bbJunior smart yield token to redeem from
+    /// @param amount tokens to redeem
+    function redeem(ISmartYield bbJunior, uint256 amount) external { 
+        bbJunior.sellTokens(amount, 0, block.timestamp);
```
eswak, line 37: somewhat risky to put a minimum amount out of 0, but given the amounts at stake, ok for me

ElliotFriedman: if we are selling into an amm, I think we risk getting front and back run so this function should have a minAmountOut param to limit damage from slippage

ElliotFriedman (changes requested):

eswak (approved):

PR #497, opened 2022-01-24
Debt Ceiling PSM

ElliotFriedman:


contracts/peg/SafeCeilingPSM.sol
```diff
@@ -0,0 +1,77 @@
+pragma solidity ^0.8.4;
+
+import "./SafePSM.sol";
+
+contract SafeCeilingPSM is SafePSM {
```
Joeysantoro, line 5: Doesn't look too complex but I'm pretty worried about adding this much inheritance and hidden logic in our PSM architecture.

IMO we shouldn't ship this any time soon and rather use semi-automated sentinels and policies in the short-medium term

PR #506, opened 2022-01-26
PCVSentinel

ditchfieldcaleb:


contracts/pcv/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,173 @@
...
+pragma solidity ^0.8.4;
+
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+import "../../refs/CoreRef.sol";
+import "./IPCVSentinel.sol";
+import "../IPCVDeposit.sol";
+import "../../libs/CoreRefPauseableLib.sol";
+import "./IGuard.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
```
eswak, line 11: Can you sort ? 😅

contracts/pcv/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,173 @@
...
+        override
+        returns (address[] memory all) 
+    {
+        all = new address[](guards.length());
+
+        for (uint i = 0; i < guards.length(); i++) {
+            all[i] = guards.at(i);
+        }
+
+        return all;
```
eswak, line 61: guards.values() should be enough here ?

contracts/pcv/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,173 @@
...
+        for (uint256 i=0; i<guards.length(); i++) {
+            if(IGuard(guards.at(i)).check()) {
+                numGuardsCanProtec++;
+            }
+        }
+
+        address[] memory guardsWhoCanProtec = new address[](numGuardsCanProtec);
+
+        for (uint256 i=0; i<guards.length(); i++) {
+            if(IGuard(guards.at(i)).check()) {
```
eswak, line 86: you could cache these check() results in a boolean array to save external calls (I know it's not supposed to be called onchain, but...)

contracts/pcv/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,173 @@
...
+    function protec(address guard) 
+        external
+        override
+    {
+        require(guards.contains(guard), "Guard does not exist.");
+
+        bool needsProtec = IGuard(guard).check();
+
+        if (needsProtec) {
+            (address[] memory targets, bytes[] memory datas) = IGuard(guard).getProtecActions();
```
eswak, line 142: "data" has no plural

contracts/pcv/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,173 @@
...
+
+            for(uint256 i=0; i<targets.length; i++) {
+                require(targets[i] != address(this), "Nyeh!");
+                targets[i].call(datas[i]); // Unopinionated about reverts
```
eswak, line 146: What do you mean by "Unopinionated about reverts" ?

test/unit/pcv/sentinel/PCVSentinel.test.ts
```diff
@@ -0,0 +1,202 @@
...
+import chai from 'chai';
+import { forceEth, forceSpecificEth } from '@test/integration/setup/utils';
+
+// This will theoretically make the error stack actually print!
+chai.config.includeStack = true;
+
+// Import if needed, just a helper.
+// const toBN = ethers.BigNumber.from;
+
+describe.only('PCV Sentinel', function () {
```
eswak, line 26: don't commit the .only ^^

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: we use openzeppelin v3 not v4 so we don't have access to .values :(

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: technically you are correct, but I add an s to help differentiate an array of data-objects from a single data object

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: the pcv sentinel does not care if the actions it takes are reverting. adds complexity; the guard can handle that.

test/unit/pcv/sentinel/PCVSentinel.test.ts
ditchfieldcaleb: oop will fix thanks

contracts/pcv/sentinel/PCVSentinel.sol
Joeysantoro: whaa really? Thats surprising

contracts/pcv/sentinel/PCVSentinel.sol
Joeysantoro: calldatas?

contracts/pcv/sentinel/PCVSentinel.sol
Joeysantoro: I actually feel pretty strongly we should fail loud

contracts/pcv/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,173 @@
...
+     * @dev we use try-catch here in case any guards fail, this could be the case
+     * if something changes before the transaction hits the chain
+     * @param whichGuards the guards to try to activate
+     */
+    function protecMany(address[] calldata whichGuards)
+        public 
+        override
+    {
+        for(uint256 i=0; i<whichGuards.length; i++) {
+            try this.protec(whichGuards[i]) {
```
Joeysantoro, line 166: Swallowing errors is a huge defi antipattern.

I think we are far more at risk of missing something here than being screwed because we failed loud. Need some pretty specific contrary evidence to change my mind

Joeysantoro (changes requested):

contracts/pcv/sentinel/guards/examples/BalanceGuard.sol
```diff
@@ -0,0 +1,33 @@
...
+contract BalanceGuard is IGuard {
+    address constant private ZERO = address(0x0);
+
+    function check() 
+        external 
+        view
+        override 
+        returns (bool) 
+    {
+        if(address(this).balance % 2 == 0) return true;
```
Joeysantoro, line 15: assuming this is just a test?

ditchfieldcaleb: yeah, it lets me send eth to it and change the trigger condition

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: Require flawless execution of proective actions is an antipattern in security.
I agree that the risk of not knowing if something has failed is large, but I think the risk of not preventing a hack is much larger.

We could add in a bool that lets the caller decide though.

contracts/pcv/sentinel/PCVSentinel.sol
ElliotFriedman, line 146: you should emit an event here with the result of each call, true or false, the target address, and the returned data if any.

contracts/pcv/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,173 @@
...
+            (address[] memory targets, bytes[] memory datas) = IGuard(guard).getProtecActions();
+
+            for(uint256 i=0; i<targets.length; i++) {
+                require(targets[i] != address(this), "Nyeh!");
+                targets[i].call(datas[i]); // Unopinionated about reverts
+            }
+
+            emit Protected(guard);
+        } else {
+            revert("No need to protec.");
```
ElliotFriedman, line 151: no need to revert if there is nothing to protec

instead, have the function return true or false and return true if anything was protected and false otherwise

contracts/pcv/sentinel/PCVSentinel.sol
ElliotFriedman: I agree with Joey here, we should not just ignore errors when we are trying to protec.

contracts/pcv/sentinel/guards/examples/BalanceGuard.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
ElliotFriedman, line 1: this should go in the mocks folder

contracts/pcv/sentinel/guards/examples/MultiActionGuard.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
ElliotFriedman, line 1: this should go in the mocks folder

contracts/pcv/sentinel/guards/examples/NoOpGuard.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
ElliotFriedman, line 1: this should go in the mocks folder

contracts/pcv/sentinel/guards/examples/ReEntrancyGuard.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
ElliotFriedman, line 1: this should go in the mocks folder

contracts/pcv/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,173 @@
...
+
+    // ---------- Public State-Changing API ----------
+
+    /**
+     * @notice checks the guard and runs its protec actions if needed
+     * @param guard the guard to activate
+     */
+    function protec(address guard) 
+        external
+        override
```
ElliotFriedman, line 135: this function should have a reentrancy guard

ElliotFriedman (changes requested):

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: > whaa really? Thats surprising

yeah I spent a few hours once trying to fix it but that requires using *both* versions in our contract compilation process for anything that depends on core (since we use it in core!) and that became a nightmare

wouldn't recommend it

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: yeah if I want to use this onchain I'll definitely make an optimized version of this, makes it harder to read & audit though

ditchfieldcaleb: wait nvm this isn't hard. good callout.

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: @Joeysantoro @ElliotFriedman neither of you have addressed the fact that the only way to hear about errors is events here.

> If someone calls into this contract from another contract and it reverts, we won't see those. We'll see the events if it doesn't revert though.

Therefore the best way to fail loudly is to *not* revert.

I added in a bool that lets the caller decide if it should revert or not.

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: switched this to ensuring it doesn't fail. I really don't think we need an event for each individual sub-call though.

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: Reverting lets us bubble up errors and emit events when executing multiple protec actions

contracts/pcv/sentinel/guards/examples/BalanceGuard.sol
ditchfieldcaleb: It's not a mock, it's an example

contracts/pcv/sentinel/guards/examples/MultiActionGuard.sol
ditchfieldcaleb: (See above)

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: Now that we're not using delegatecall, and we're ensuring that we aren't calling ourselves with the address(this) check, no re-entrancy guard is needed

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: (it's important to remember here that we won't be the only ones calling into this contract)

contracts/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,180 @@
...
+     * @param allowFailure whether or not the entire transaction should revert on failure
+     * @param whichGuards the guards to try to activate
+     */
+    function protecMany(bool allowFailure, address[] calldata whichGuards)
+        public 
+        override
+    {
+        for(uint256 i=0; i<whichGuards.length; i++) {
+            if (allowFailure) {
+                try this.protec(whichGuards[i]) {
```
thomas-waite, line 170: Why not turn `protec` into a `public` function and just call `try protec(whichGuards[i])`? 

Not generally a fan of calling smart contract methods from within the contract itself using `this.` syntax

thomas-waite (commented):
~Question about syntax~

contracts/sentinel/PCVSentinel.sol
thomas-waite: Ignore - needed to use `try`

test/unit/sentinel/PCVSentinel.test.ts
```diff
@@ -0,0 +1,216 @@
...
+import chai from 'chai';
+import { forceEth, forceSpecificEth } from '@test/integration/setup/utils';
+
+// This will theoretically make the error stack actually print!
+chai.config.includeStack = true;
+
+// Import if needed, just a helper.
+// const toBN = ethers.BigNumber.from;
+
+describe.only('PCV Sentinel', function () {
```
ElliotFriedman, line 26: remove .only

test/unit/sentinel/PCVSentinel.test.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  describe('sentinel with basic condition check', async () => {
+    it('adds and checks and protecs', async () => {
+      await pcvSentinel.connect(impersonatedSigners[guardianAddress]).knight(balanceGuard.address);
+      expect(await pcvSentinel.isGuard(balanceGuard.address)).to.equal(true);
+      await expect(pcvSentinel.protec(balanceGuard.address)).to.emit(pcvSentinel, 'Protected');
+    });
+
+    it('reverts if condition not met', async () => {
+      await pcvSentinel.connect(impersonatedSigners[guardianAddress]).knight(balanceGuard.address);
```
ElliotFriedman, line 122: add this into a beforeEach hook instead of duplicating this logic across multiple tests

contracts/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,180 @@
...
+                    // the emit happens in protec
+                } catch {
```
ElliotFriedman, line 172: If there is a failure, you should emit that error string and target guard smart contract so that it can be easily debugged

contracts/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,180 @@
...
+
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+import "../refs/CoreRef.sol";
+import "../pcv/IPCVDeposit.sol";
+import "./IPCVSentinel.sol";
+import "./IGuard.sol";
+
+/**
+ * @title PCV Sentinel
```
ElliotFriedman, line 12: **style:** all comments should use `///` and not multi line comments to conform to our codebase

contracts/sentinel/guards/examples/ReEntrancyGuard.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
ElliotFriedman, line 1: Example contracts should be moved to the mocks folder

contracts/pcv/sentinel/PCVSentinel.sol
ElliotFriedman: That is a case for only allowing all or nothing tx's.

Given that ethereum is a dark forest and a highly adversarial environment, we should not allow our contracts to halfway execute a transaction that is supposed to protect many of them at the same time.

contracts/pcv/sentinel/PCVSentinel.sol
ElliotFriedman: I'm not convinced this is the right decision.

If you pass in an array of guards to protec, and only half of them can be protec'd, why should we revert when the revert is ignored?

Also, any errors that try to bubble up get swallowed and are forgotten so this error string is literally pointless unless you try to call protec individually, and even then, it's much nicer to return a boolean than fail loudly.

If protecMany reverted on errors to protec, then I would be fine with this pattern.

contracts/pcv/sentinel/PCVSentinel.sol
ElliotFriedman: I would rather be safe than sorry, especially when executing arbitrary calldata against targets. We should add a reentrancy guard here.

ElliotFriedman (changes requested):
Things to be fixed before approval:
- Reentrancy guard on protec
- Natspec comments
- Enforce all or nothing execution on protecMany and remove `allowFailure` param
- Move examples to mocks folder

contracts/sentinel/guards/examples/ReEntrancyGuard.sol
ditchfieldcaleb: This isn't a mock, it need to be here so that people see it as an example.

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: > That is a case for only allowing all or nothing tx's.
> 
> Given that ethereum is a dark forest and a highly adversarial environment, we should not allow our contracts to halfway execute a transaction that is supposed to protect many of them at the same time.

Sorry I don't really know what you're saying here

contracts/sentinel/PCVSentinel.sol
ditchfieldcaleb: Both are natspec and should be acceptable imo

contracts/sentinel/PCVSentinel.sol
ditchfieldcaleb: Good call, will do

test/unit/sentinel/PCVSentinel.test.ts
ditchfieldcaleb: Which ones we want knighted aren't the same across tests, otherwise I would

contracts/pcv/sentinel/PCVSentinel.sol
ditchfieldcaleb: There is literally no way that re-entrancy can happen. We're good.

ElliotFriedman:
@ditchfieldcaleb lmk when you're ready for another review after changes have been pushed

contracts/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,109 @@
...
+        bool needsProtec = IGuard(guard).check();
+
+        if (needsProtec) {
+            (address[] memory targets, bytes[] memory calldatas) = IGuard(guard)
+                .getProtecActions();
+
+            for (uint256 i = 0; i < targets.length; i++) {
+                require(targets[i] != address(this), "Cannot target self.");
+                (bool success, ) = targets[i].call(calldatas[i]);
+                require(success, "Sub-call failed."); // Fail this call if any of the sub-calls fail
```
ditchfieldcaleb, line 101: @TODO: pass through the error message

ElliotFriedman, line 101: let's bubble up errors here if possible

ElliotFriedman:
I'm really liking the new PCV Sentinel! It is very minimal and keeps things simple.

contracts/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,109 @@
...
+    // ---------- Governor-or-Admin-Or-Guardian-Only State-Changing API ----------
+
+    /**
+     * @notice adds a guard contract to the PCV Sentinel
+     * @param guard the guard-contract to add
+     */
+    function knight(address guard)
+        external
+        override
+        isGovernorOrGuardianOrAdmin
```
ElliotFriedman, line 60: Should a guardian be able to add a guard? This is essentially saying that guardian can add a contract into the system that can construct calldata that will be executed by a contract with PCV Controller role. It feels like this gives the guardians too much power.

ditchfieldcaleb: @ElliotFriedman no that's not the case, where are you getting the idea that this is a PCVController still?

(It was, but shouldn't be anymore)

ditchfieldcaleb: Just confirmed, yeah, this will use the PCVGuardian to pull funds to safety :)

contracts/sentinel/PCVSentinel.sol
ElliotFriedman, line 101: bubble up error here

ElliotFriedman (changes requested):

contracts/sentinel/PCVSentinel.sol
ditchfieldcaleb: Closing due to lack of further response, feel free to re-open @ElliotFriedman if you have further concerns.

ditchfieldcaleb:
> Things to be fixed before approval:
> 
> * Reentrancy guard on protec
> * Natspec comments
> * Enforce all or nothing execution on protecMany and remove `allowFailure` param
> * Move examples to mocks folder

- Reentrancy guard: added
- Examples aren't mocks. Yeah, perhaps it'd be better form to move them to the mocks folder, but practicality -> form, and in the mocks folder the examples would never be seen unless you already knew they existed. The examples are a reference for knowing how to create actual guards. Going to leave this as-is.
- Removed allowFailure & protecMany, vastly simplified
- Comments natspec'd.

ditchfieldcaleb:
@Joeysantoro @ElliotFriedman final review requested. I've made a major and potentially controversial change in the last commit: added receive() and fallback() as payable functions, and adjusted the Guard ABI to send an array of values along with the addresses and bytes/data array, which is used when calling the provided calldata.

Justification on receives() and payable(): any contracts that might incentivize calling them in eth. I don't think we do this now but decent chance we will at some point.

Justification on sending value: automation of non-critical amounts of pcv. Topping off small but necessary contracts or eoas. Sending funds received via payable() and receive() to some pre-determined place.

Flexible here, not super attached to the value in the calls, but fairly convinced that having receive() and fallback() is a good idea. Though if we remove the value in the calls we'll need to add in an admin method for recovering eth in the contract received via receive() and fallback().

contracts/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * @dev receive() and fallback() have been added and made payable for cases where the contract
+     * needs to be able to receive eth as part of an operation - such as receiving an incentivization
+     * (in eth) from a contract as part of operation. For similar (and not unlikely) edge cases,
+     * the contract also has the capability of sending eth inside when instructed by a guard to do so.
+     */
+    receive() external payable {}
+
+    fallback() external payable {}
```
ElliotFriedman, line 130: These two functions cannot exist as long as this contract is not a PCV deposit. If it's a PCV deposit then you can add them because you have a way to remove the eth that this contract could potentially receive. Right now, any eth that is received is locked forever.

ElliotFriedman (dismissed):

contracts/sentinel/PCVSentinel.sol
Joeysantoro: I think its fine

This contract isn't meant to hold meaningful amounts of ETH. Any "safe ETH" moves will be sent to DAO timelock

ElliotFriedman: It also can specify how much eth to send in the `protec()` function so it's good to go at this point.

contracts/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,131 @@
...
+            (
+                address[] memory targets,
+                bytes[] memory calldatas,
+                uint256[] memory values
+            ) = IGuard(guard).getProtecActions();
+
+            for (uint256 i = 0; i < targets.length; i++) {
+                require(targets[i] != address(this), "Cannot target self.");
+                (bool success, bytes memory returndata) = targets[i].call{
+                    value: values[i]
```
ElliotFriedman, line 107: We should add a test that rescues ERC20 tokens and eth using this function and some rescue guards.

ElliotFriedman: The rescue guards can be hardcoded to send to an immutable address, and in prod this can be the timelock.

ditchfieldcaleb: Good call, adding

contracts/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,131 @@
...
+
+    // The set of all guards
+    EnumerableSet.AddressSet private guards;
+
+    /**
+     * @notice Creates a PCV Sentinel with no guards
+     * @param _core the Tribe core address
+     */
+    constructor(address _core) CoreRef(_core) {
+        _setContractAdminRole(keccak256("PCV_SENTINEL_ADMIN_ROLE"));
```
Joeysantoro, line 30: I think we should deprecate the "contract admin role" concept and use TribeRoles.

contracts/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,131 @@
...
+    // ---------- Public State-Changing API ----------
+
+    /**
+     * @notice checks the guard and runs its protec actions if needed
+     * @param guard the guard to activate
+     */
+    function protec(address guard) external override nonReentrant {
+        require(guards.contains(guard), "Guard does not exist.");
+
+        if (IGuard(guard).check()) {
```
Joeysantoro, line 97: require(guard.check()) is cleaner than if-revert

contracts/sentinel/PCVSentinel.sol
```diff
@@ -0,0 +1,131 @@
...
+     * @param guard the address to check
+     */
+    function isGuard(address guard) external view override returns (bool) {
+        return guards.contains(guard);
+    }
+
+    /**
+     * @notice returns a list of all guards
+     */
+    function allGuards() external view override returns (address[] memory all) {
```
Joeysantoro, line 46: I think we should add two methods for automation purposes:
1. GetAllProtecableGuards
2. protecAll

so the defender script can run protecAll(getAllProtecableGuards())

Joeysantoro (commented):
Left a few comments, none are dealbreakers but all are nice to have and worth it imo

Joeysantoro:
Love this contract btw. Attack surface is minimal and I believe it is prod-ready after the above changes are reviewed

contracts/sentinel/PCVSentinel.sol
ditchfieldcaleb: This can/will be implemented in a helper contract, no need to add it to the core API of the Sentinel

ditchfieldcaleb: (esp because protecAll must necessarily be opinionated about how to handle reverts within single guards when running all of them, etc etc)

contracts/sentinel/PCVSentinel.sol
```diff
@@ -124,7 +116,7 @@ contract PCVSentinel is IPCVSentinel, CoreRef, ReentrancyGuard {
     }
 
     /**
-     * @dev receive() and fallback() have been added and made payable for cases where the contract
+     * @dev receive() and fallback() have been added and made payable for cases wrhere the contract
```
Joeysantoro, line 120: typo

Joeysantoro (approved):

PR #515, opened 2022-01-31
Global Rate Limited Minter

ElliotFriedman:
Global Rate Limited Minter is a new component of our system that will allow all FEI mint calls to pass through it.

There are three separate types of rate limit:

1. Is a rate limit and buffer per address. This allows us to throttle how much each individual component in our system can mint FEI or do some other rate limited action.
2. Global rate limit and buffer. This allows us to set global maximums on how much FEI can be minted or some other action be taken by our system within a fixed time period.
3. Sub Governor Max Rate Limit Per Second. This is the maximum amount the sub governor role can add to a new address.

The goals of this contract is to reduce the attack surface of our system by only giving the minter role to the GlobalRateLimitedMinter which significantly reduces the amount of contracts and roles that will receive the minter role.

Rate Limited in the Global Rate Limited Minter is always hardcoded to never allow partial buffer depletions, so either the mint succeeds and it mints all the FEI it was supposed to, or it fails and does not.

A new function `mintMaxAllowableFei` was added to the Global Rate Limited Minter to allow timelocks and other governance bound actors to mint the maximum amount of FEI available at a given point in time.


contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,170 @@
...
+    )
+        CoreRef(coreAddress)
+        RateLimited(_maxRateLimitPerSecond, _rateLimitPerSecond, _bufferCap, _doPartialAction)
+    {}
+
+    /// @notice add an authorized minter contract
+    /// @param _minter the new address to add as a minter
+    /// @param _rateLimitPerSecond the rate limit per second for this minter
+    /// @param _bufferCap  the buffer cap for this minter
+    function addAddress(address _minter, uint256 _rateLimitPerSecond, uint256 _bufferCap) external virtual override onlyGovernorOrAdmin {
```
eswak, line 43: You're at a level of abstraction here where addresses are not necessarily "minters", but are rate-limited on anything. I wouldn't call the addresses "minter", because we could rate-limit redemptions or swaps or anything else

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,170 @@
...
+    function setIndividualBufferCap(address minter, uint256 newBufferCap) external virtual override onlyGovernorOrAdmin {
+        require(newBufferCap <= bufferCap, "MultiRateLimited: new buffer cap is over global max");
+
+        _setIndividualBufferCap(minter, newBufferCap);
+    }
+
+    /// @notice the amount of action used before hitting limit
+    /// @dev replenishes at rateLimitPerSecond per second up to bufferCap
+    /// @param minter the address whose buffer will be returned
+    function buffer(address minter) public view override returns(uint256) {
```
eswak, line 104: maybe name this `individualBuffer(address)` because all other methods are individualSomething. Also `buffer()` will return the global buffer, so it's confusing to have 2 functions with the same name

contracts/utils/MultiRateLimited.sol
ElliotFriedman: what should we call them instead?

contracts/utils/MultiRateLimited.sol
ElliotFriedman: good callout

contracts/mock/MockMinter.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "./../utils/GlobalRateLimitedMinter.sol";
+
+contract MockMinter {
+    GlobalRateLimitedMinter globalRateLimitedMinter;
```
thomas-waite, line 7: Is it worthwhile making this state variable visibility explicit? It's `internal` by default sure - what convention for explicit/implicit do we use normally?

contracts/utils/GlobalRateLimitedMinter.sol
```diff
@@ -0,0 +1,43 @@
...
+    /// @param coreAddress address of the core contract
+    /// @param _globalMaxRateLimitPerSecond maximum amount of fei that can replenish per second
+    /// @param _globalRateLimitPerSecond maximum global buffer cap
+    /// @param _bufferCap maximum rate limit per second per minter
+    /// @param maxBufferCap maximum buffer cap per minter
+    /// @param _doPartialAction boolean to describe whether partial actions are allowed or not
+    constructor(
+        address coreAddress,
+        uint256 _globalMaxRateLimitPerSecond,
+        uint256 _globalRateLimitPerSecond,
```
thomas-waite, line 20: I'm not sure if these argument names (`_globalMaxRateLimitPerSecond`,  `_globalRateLimitPerSecond`) are the clearest or how they map up to the description in the natspec comments. 

`_globalRateLimitPerSecond maximum global buffer cap` seems to suggest this is a maximum global constant, rather than a rate? Would `_globalMaxBufferCap` be more accurate?

contracts/utils/MultiRateLimited.sol
thomas-waite: `_throttledAddress`, `_rateLimitedAddress`?

thomas-waite (commented):
WIP review. Would be awesome to see a short PR description too that calls out key design choices to help give me more context + effectively review :)

contracts/utils/IMultiRateLimited.sol
```diff
@@ -0,0 +1,58 @@
...
+    /// @notice the last time the buffer was used by each minter
+    function individualLastBufferUsedTime(address) external view returns(uint256);
+
+    /// @notice the cap of the buffer that can be used at once
+    function individualBufferCap(address) external view returns(uint256);
+
+    // ----------- Events -----------
+    
+    /// @notice emitted when a buffer is eaten into
+    event IndividualBufferUsed(address minter, uint256 amountUsed, uint256 bufferRemaining);
```
thomas-waite, line 20: `minter` references here should also be updated to `throttledAddress` or whichever variable name we use to reference that this contract applies more broadly

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,170 @@
...
+
+    /// @notice the buffer at the timestamp of lastBufferUsedTime
+    mapping (address => uint256) private _bufferStored;
+
+    constructor(
+        address coreAddress,
+        uint256 _maxRateLimitPerSecond,
+        uint256 _rateLimitPerSecond,
+        uint256 _bufferCap,
+        uint256 maxBufferCap,
```
thomas-waite, line 32: `maxBufferCap` looks like an unused variable? Can be removed from the constructor?

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,170 @@
...
+import "@openzeppelin/contracts/utils/math/Math.sol";
+
+/// @title contract for putting a rate limit on how fast an address can perform an action e.g. Minting
+/// there are two buffers, one buffer which is each individual addresses's current buffer,
+/// and then there is a global buffer which is the buffer that each individual address must respect as well
+/// @author Fei Protocol
+contract MultiRateLimited is RateLimited, IMultiRateLimited {
+
+    /// @notice the rate per second for each address
+    mapping (address => uint256) public individualRateLimitPerSecond;
```
thomas-waite, line 16: Should these `public` state variables not have the `override` specifier applied? 

They'll generate getters and you're overriding `function individualRateLimitPerSecond(address) external view returns(uint256);` etc on the inherited interface

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,170 @@
...
+
+        individualRateLimitPerSecond[_minter] = _rateLimitPerSecond;
+        individualLastBufferUsedTime[_minter] = block.timestamp;
+        individualBufferCap[_minter] = _bufferCap;
+        _bufferStored[_minter] = _bufferCap;
+
+        emit IndividualRateLimitPerSecondUpdate(_minter, 0, _rateLimitPerSecond);
+    }
+
+    /// @notice add an authorized minter contract
```
thomas-waite, line 54: Minor nitpick, the @notice needs updating

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,170 @@
...
+    /// @param _bufferCap  the new buffer cap for this minter
+    function updateAddress(address _minter, uint256 _rateLimitPerSecond, uint256 _bufferCap) external virtual override onlyGovernorOrAdmin {
+        require(_bufferCap <= bufferCap, "MultiRateLimited: buffercap too high");
+
+        individualRateLimitPerSecond[_minter] = _rateLimitPerSecond;
+        individualLastBufferUsedTime[_minter] = block.timestamp;
+        individualBufferCap[_minter] = _bufferCap;
+        _bufferStored[_minter] = _bufferCap;
+
+        emit IndividualRateLimitPerSecondUpdate(_minter, 0, _rateLimitPerSecond);
```
thomas-waite, line 66: On this event, the second argument shouldn't be `0` right? It is `uint256 oldRateLimitPerSecond` so should be `individualRateLimitPerSecond[_minter]`?

contracts/mock/MockMinter.sol
ElliotFriedman: It's a mock contract so I don't think it's worth it

contracts/utils/MultiRateLimited.sol
ElliotFriedman: done

contracts/utils/MultiRateLimited.sol
ElliotFriedman: removed

contracts/utils/IMultiRateLimited.sol
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+/// @title interface for putting a rate limit on how fast a contract can perform an action, e.g. Minting
+/// @author Fei Protocol
+interface IMultiRateLimited {
+
+    /// @notice the rate per second for each address
+    function individualRateLimitPerSecond(address) external view returns(uint256);
```
ditchfieldcaleb, line 9: would prefer getter syntax here: getRateLimit(address)

contracts/utils/IMultiRateLimited.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    /// @notice the last time the buffer was used by each address
+    function individualLastBufferUsedTime(address) external view returns(uint256);
```
ditchfieldcaleb, line 12: same as above: getLastBufferUsedTime(address)

contracts/utils/IMultiRateLimited.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    /// @notice the cap of the buffer that can be used at once
+    function individualBufferCap(address) external view returns(uint256);
```
ditchfieldcaleb, line 15: getBufferCap(address)

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,202 @@
...
+
+        delete rateLimitPerAddress[rateLimitedAddress];
+
+        emit IndividualRateLimitPerSecondUpdate(rateLimitedAddress, oldRateLimitPerSecond, 0);
+    }
+
+    /// @notice set the rate limit per second
+    /// @param rateLimitedAddress the address whose buffer will be set
+    /// @param newRateLimitPerSecond the new rate limit per second for this rateLimitedAddress
+    function setIndividualRateLimitPerSecond(address rateLimitedAddress, uint112 newRateLimitPerSecond) external virtual override onlyGovernorOrAdmin {
```
ditchfieldcaleb, line 90: would prefer syntax of setRateLimit(address, uint112) but not too attached to the idea

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,202 @@
...
+    /// @param rateLimitedAddress the address whose buffer will be returned
+    function individualBuffer(address rateLimitedAddress) public view override returns(uint144) {
+        RateLimitData memory rateLimitData = rateLimitPerAddress[rateLimitedAddress];
+
+        uint256 elapsed = block.timestamp - rateLimitData.lastBufferUsedTime;
+        return uint144(Math.min(rateLimitData.bufferStored + (rateLimitData.rateLimitPerSecond * elapsed), rateLimitData.bufferCap));
+    }
+
+    /// @notice the rate per second for each address
+    function individualRateLimitPerSecond(address limiter) external override view returns(uint256) {
```
ditchfieldcaleb, line 117: shouldn't `limiter` be `limited`?

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,202 @@
...
+    /// 2. Reverts
+    /// Depending on whether doPartialAction is true or false
+    /// @param rateLimitedAddress the address whose buffer will be depleted
+    /// @param amount the amount to remove from the rateLimitedAddress's buffer
+    function _depleteBuffer(address rateLimitedAddress, uint256 amount) internal returns(uint256) {
+        _depleteBuffer(amount);
+
+        uint256 newBuffer = individualBuffer(rateLimitedAddress);
+        
+        uint256 usedAmount = amount;
```
ditchfieldcaleb, line 143: I think we talked about removing partial mints on Discord, thoughts?

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,202 @@
...
+
+        rateLimitPerAddress[rateLimitedAddress].lastBufferUsedTime = _blockTimestamp();
+
+        emit IndividualBufferUsed(rateLimitedAddress, usedAmount, newBuffer - usedAmount);
+
+        return usedAmount;
+    }
+
+    /// @notice return current block timestamp as a uint32
+    function _blockTimestamp() private view returns (uint32) {
```
ditchfieldcaleb, line 161: This function is one line, I don't see a need for it

ditchfieldcaleb (changes requested):
Looking great, love the revamped architecture. 

My main complaint is the usage of the word `individual` everywhere. See the comments here and there for my suggested update, but it's not absolutely necessary.

Thought we talked about removing partial mints though.

contracts/utils/MultiRateLimited.sol
ElliotFriedman: its cleaner to call `_blockTimestamp()` instead of this `block.timestamp.toUint32()`

contracts/utils/MultiRateLimited.sol
ElliotFriedman: I don't remember us reaching a consensus on this

contracts/utils/MultiRateLimited.sol
ElliotFriedman: In the RateLimitedContract, there is a method `setRateLimitPerSecond`, so I would like to keep this naming convention

ditchfieldcaleb: that sounds good, i'm mainly referring to the word individual

so perhaps setRateLimitPerSecond still? it'll be an overloaded function which seems fine

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: Disagree tbh. But it also adds gas as it's an extra function call (unless the compiler embeds this inline, which I don't think it does)

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: looks like you did this, cool!

contracts/utils/MultiRateLimited.sol
ElliotFriedman: since when have we been concerned about optimizing down to the last opcode? Our code isn't user facing so it's fine

contracts/utils/MultiRateLimited.sol
ElliotFriedman: fixed

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: You're positing a position I'm not taking; this is an easy and simple optimization. We should take these when we see them.

contracts/utils/GlobalRateLimitedMinter.sol
ElliotFriedman: I've updated the comments in the constructor args description. Please check them and let me know what you think.

contracts/utils/MultiRateLimited.sol
Joeysantoro: For multi minter I think we agreed yeah

Joeysantoro: having it be a global var across all minters also limits utility

contracts/utils/GlobalRateLimitedMinter.sol
```diff
@@ -0,0 +1,40 @@
...
+        )
+    {}
+
+    /// @notice mint fei to the target address and deplete the buffer
+    /// pausable and depletes the msg.sender's buffer
+    /// @param to the recipient address of the minted FEI
+    /// @param amount the amount of FEI to mint
+    function mintFei(address to, uint256 amount) external virtual override whenNotPaused {
+        _depleteBuffer(msg.sender, amount);
+        super._mintFei(to, amount);
```
Joeysantoro, line 38: why do you need super._mintFei()? Just _mintFei should work right?

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,203 @@
...
+/// @author Fei Protocol
+contract MultiRateLimited is RateLimited, IMultiRateLimited {
+    using SafeCast for *;
+
+    /// @notice the struct containing all information per rate limited address
+    struct RateLimitData {
+        uint32 lastBufferUsedTime;
+        uint144 bufferCap;
+        uint112 rateLimitPerSecond;
+        uint144 bufferStored;
```
Joeysantoro, line 22: Can give some breathing room to some of the other vars given this is packed into 2 slots anyway

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,203 @@
...
+    {}
+
+    // ----------- Governor and Admin only state changing api -----------
+
+    /// @notice add an authorized rateLimitedAddress contract
+    /// @param rateLimitedAddress the new address to add as a rateLimitedAddress
+    /// @param _rateLimitPerSecond the rate limit per second for this rateLimitedAddress
+    /// @param _bufferCap  the buffer cap for this rateLimitedAddress
+    function addAddress(address rateLimitedAddress, uint112 _rateLimitPerSecond, uint144 _bufferCap) external virtual override onlyGovernorOrAdmin {
+        require(_bufferCap <= bufferCap, "MultiRateLimited: new buffercap too high");
```
Joeysantoro, line 46: This isn't strict enough imo. There should be some tighter parameters we check against rather than the global buffer cap.

ElliotFriedman: We can have a max buffercap per address that is stored in the MultiRateLimited contract

contracts/utils/MultiRateLimited.sol
ElliotFriedman: Whatever we do here, it's cleaner to keep bufferCap and bufferStored the same type. I'm also not seeing of any scenario where we're going to replenish a buffer at a speed faster than 2^112 or have a bufferCap larger than 2^144

contracts/utils/MultiRateLimited.sol
ElliotFriedman: ok, removed the partial buffer depletion from MultiRateLimited

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,270 @@
...
+
+        _updateAddress(rateLimitedAddress, _rateLimitPerSecond, _bufferCap);
+    }
+
+    /// @notice add an authorized rateLimitedAddress contract
+    /// @param rateLimitedAddress the new address to add as a rateLimitedAddress
+    /// @param _rateLimitPerSecond the rate limit per second for this rateLimitedAddress
+    /// @param _bufferCap  the buffer cap for this rateLimitedAddress
+    /// TODO figure out what the modifier should be for this function
+    function addAddressSubGovernor(address rateLimitedAddress, uint112 _rateLimitPerSecond, uint144 _bufferCap) external virtual override onlyGovernorOrAdmin {
```
ElliotFriedman, line 97: remove the rate limit per second and buffer cap and automatically set all calls from sub gov to these global values

use the new roles

remove the admin modifiers from this contract

contracts/core/TribeRoles.sol
```diff
@@ -70,4 +70,7 @@ library TribeRoles {
 
     /// @notice capable of changing parameters within non-critical ranges
     bytes32 internal constant MINOR_PARAM_ROLE = keccak256("MINOR_PARAM_ROLE");
+
+    /// @notice capable of changing parameters within non-critical ranges
+    bytes32 internal constant MINOR_MINTER_ADD_ROLE = keccak256("MINOR_MINTER_ADD_ROLE");
```
ditchfieldcaleb, line 75: Should be "add_minter" not "minter_add"

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,272 @@
...
+            require(_bufferCap <= maxBufferCap, "MultiRateLimited: max buffer cap exceeds non governor allowable amount");
+        }
+        require(_bufferCap <= bufferCap, "MultiRateLimited: buffercap too high");
+
+        _updateAddress(rateLimitedAddress, _rateLimitPerSecond, _bufferCap);
+    }
+
+    /// @notice add an authorized rateLimitedAddress contract
+    /// @param rateLimitedAddress the new address to add as a rateLimitedAddress
+    /// TODO figure out what the modifier should be for this function
```
ditchfieldcaleb, line 107: TODO

test/unit/fei/minter/GlobalRateLimitedMinter.test.ts
```diff
@@ -0,0 +1,583 @@
...
+import { expect } from 'chai';
+import hre, { ethers } from 'hardhat';
+import { Contract, Signer, utils } from 'ethers';
+import { Core, Fei, GlobalRateLimitedMinter, MockMinter } from '@custom-types/contracts';
+import { keccak256 } from 'ethers/lib/utils';
+
+const toBN = ethers.BigNumber.from;
+const scale = ethers.constants.WeiPerEther;
+
+describe('GlobalglobalRateLimitedMinter', function () {
```
ditchfieldcaleb, line 11: typo

contracts/utils/GlobalRateLimitedMinter.sol
```diff
@@ -0,0 +1,44 @@
...
+    /// @param coreAddress address of the core contract
+    /// @param _globalMaxRateLimitPerSecond maximum amount of fei that can replenish per second ever, this amount cannot be changed by governance
+    /// @param _globalRateLimitPerSecond maximum rate limit per second per minter
+    /// @param _bufferCap maximum global buffer cap
+    /// @param _doPartialAction boolean to describe whether partial actions are allowed or not
+    constructor(
+        address coreAddress,
+        uint256 _globalMaxRateLimitPerSecond,
+        uint256 _globalRateLimitPerSecond,
+        uint256 _maxRateLimitPerSecondSubGovernor,
```
ditchfieldcaleb, line 20: Subgovernor is confusing, I thought we talked about removing It?

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,272 @@
...
+    /// @param rateLimitedAddress the address whose buffer and rate limit per second will be set
+    /// @param _rateLimitPerSecond the new rate limit per second for this rateLimitedAddress
+    /// @param _bufferCap  the new buffer cap for this rateLimitedAddress
+    function updateAddress(
+        address rateLimitedAddress,
+        uint112 _rateLimitPerSecond,
+        uint144 _bufferCap
+    ) external virtual override hasAnyOfTwoRoles(TribeRoles.MINOR_MINTER_ADD_ROLE, TribeRoles.GOVERNOR) {
+        if (core().hasRole(TribeRoles.MINOR_MINTER_ADD_ROLE, msg.sender)) {
+            require(_rateLimitPerSecond <= maxRateLimitPerSecond, "MultiRateLimited: rate limit per second exceeds non governor allowable amount");
```
ditchfieldcaleb, line 97: The concept of a "non governor allowable amount" is confusing. Might be better to just say "rate limit too high"

ditchfieldcaleb (changes requested):

thomas-waite (changes requested):
Believe you need to validate that `updateMaxBufferCap()` updates the buffer cap for an address, to a value less than the global max buffer cap.

Also as I understand things, this implementation won't globally limit how much Fei we can mint? It limits how much each individual address can mint. Without a limit on how many addresses we can add, then we can evade relatively easily (albeit only if you have the relevant permission) the global cap. Do we not need to track some cumulative global buffer that accounts for the buffers of all addresses, or something similar?

Also, I think the code readability could be improved with some variable renaming, moving around requires, seperating out the test file etc.

contracts/utils/GlobalRateLimitedMinter.sol
```diff
@@ -0,0 +1,44 @@
...
+
+import "./MultiRateLimited.sol";
+import "./IGlobalRateLimitedMinter.sol";
+
+/// @notice global contract to handle rate limited minting of FEI on a global level
+/// allows whitelisted minters to call in and specify the address to mint FEI to within
+/// that contract's limits
+contract GlobalRateLimitedMinter is MultiRateLimited, IGlobalRateLimitedMinter {
+    /// @param coreAddress address of the core contract
+    /// @param _globalMaxRateLimitPerSecond maximum amount of fei that can replenish per second ever, this amount cannot be changed by governance
```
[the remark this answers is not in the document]
ElliotFriedman: This is a concept from the RateLimited contract which is a pre-existing contract. Unless we want to edit that contract, we should just leave it as is.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,272 @@
...
+        uint256 _rateLimitPerSecond,
+        uint256 _maxRateLimitPerSecondMRL,
+        uint256 _maxBufferCap,
+        uint256 _bufferCap,
+        bool _doPartialAction
+    )
+        CoreRef(coreAddress)
+        RateLimited(_maxRateLimitPerSecond, _rateLimitPerSecond, _bufferCap, _doPartialAction)
+    {
+        require(_maxBufferCap < _bufferCap, "MultiRateLimited: max buffer cap invalid");
```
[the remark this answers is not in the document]
ElliotFriedman: Renamed to _globalBufferCap

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,272 @@
...
+
+        emit MultiMaxRateLimitPerSecondUpdate(oldMaxRateLimitPerSecond, newRateLimitPerSecond);
+    }
+
+    /// @notice update the sub gov max buffer cap
+    /// @param newSubGovBufferCap new buffer cap for sub governor added addresses
+    /// TODO determine the proper modifier for this function
+    function updateMaxBufferCap(uint256 newSubGovBufferCap) external virtual override onlyGovernor {
+        uint256 oldBufferCap = maxBufferCap;
+        maxBufferCap = newSubGovBufferCap;
```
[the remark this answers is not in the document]
ElliotFriedman: good catch, added checks.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,272 @@
...
+    /// @notice add an authorized rateLimitedAddress contract
+    /// @param rateLimitedAddress the new address to add as a rateLimitedAddress
+    /// @param _rateLimitPerSecond the rate limit per second for this rateLimitedAddress
+    /// @param _bufferCap  the buffer cap for this rateLimitedAddress
+    function addAddress(
+        address rateLimitedAddress,
+        uint112 _rateLimitPerSecond,
+        uint144 _bufferCap
+    ) external virtual override onlyGovernor {
+        _addAddress(rateLimitedAddress, _rateLimitPerSecond, _bufferCap);
```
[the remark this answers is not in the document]
ElliotFriedman: `_addAddress` emits an event

contracts/utils/MultiRateLimited.sol
ElliotFriedman: It makes perfect sense in this context. We are saying allow you to add up to a certain amount of rate limit per second as a non governor.

contracts/utils/GlobalRateLimitedMinter.sol
```diff
@@ -0,0 +1,44 @@
...
+    /// @param _globalRateLimitPerSecond maximum rate limit per second per minter
+    /// @param _bufferCap maximum global buffer cap
+    /// @param _doPartialAction boolean to describe whether partial actions are allowed or not
+    constructor(
+        address coreAddress,
+        uint256 _globalMaxRateLimitPerSecond,
+        uint256 _globalRateLimitPerSecond,
+        uint256 _maxRateLimitPerSecondSubGovernor,
+        uint256 _maxBufferCapSubGovernor,
+        uint256 _bufferCap,
```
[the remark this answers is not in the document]
ElliotFriedman: done

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,272 @@
...
+
+    /// @notice add an authorized rateLimitedAddress contract
+    /// @param rateLimitedAddress the new address to add as a rateLimitedAddress
+    /// @param _rateLimitPerSecond the rate limit per second for this rateLimitedAddress
+    /// @param _bufferCap  the buffer cap for this rateLimitedAddress
+    function addAddress(
+        address rateLimitedAddress,
+        uint112 _rateLimitPerSecond,
+        uint144 _bufferCap
+    ) external virtual override onlyGovernor {
```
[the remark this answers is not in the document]
ElliotFriedman: It's very clean to tuck all of this logic into a helper function. Because the `_addAddress` function is used in multiple places I think this is quite handy and reduces code size and duplication.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,272 @@
...
+
+    // ----------- Helper Methods -----------
+
+    function _updateAddress(address rateLimitedAddress, uint112 _rateLimitPerSecond, uint144 _bufferCap) internal {
+        uint112 oldRateLimitPerSecond = _rateLimitPerSecond;
+
+        RateLimitData storage rateLimitData = rateLimitPerAddress[rateLimitedAddress];
+
+        require(rateLimitData.lastBufferUsedTime != 0, "MultiRateLimited: rate limit address does not exist");
+        require(_rateLimitPerSecond <= MAX_RATE_LIMIT_PER_SECOND, "MultiRateLimited: rateLimitPerSecond too high");
```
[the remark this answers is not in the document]
ElliotFriedman: The reason we check against global max is because the other maximum is for the non governor, and that is checked in the function that calls this function. `updateAddress` already validates that we are under `maxRateLimitPerSecond` if a non governor is calling.

test/unit/fei/minter/GlobalRateLimitedMinter.test.ts
```diff
@@ -0,0 +1,583 @@
...
+    authorizedMinter = await (await ethers.getContractFactory('MockMinter')).deploy(globalRateLimitedMinter.address);
+
+    await core.connect(impersonatedSigners[governorAddress]).grantMinter(globalRateLimitedMinter.address);
+
+    await globalRateLimitedMinter
+      .connect(impersonatedSigners[governorAddress])
+      .addAddress(authorizedMinter.address, globalRateLimitPerSecond, bufferCap);
+  });
+
+  describe('Init', function () {
```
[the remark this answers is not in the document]
ElliotFriedman: Split into Buffer and Governor test files

contracts/utils/GlobalRateLimitedMinter.sol
```diff
@@ -0,0 +1,46 @@
...
+
+/// @notice global contract to handle rate limited minting of FEI on a global level
+/// allows whitelisted minters to call in and specify the address to mint FEI to within
+/// that contract's limits
+contract GlobalRateLimitedMinter is MultiRateLimited, IGlobalRateLimitedMinter {
+    /// @param coreAddress address of the core contract
+    /// @param _globalMaxRateLimitPerSecond maximum amount of fei that can replenish per second ever, this amount cannot be changed by governance
+    /// @param _globalRateLimitPerSecond maximum rate limit per second per address
+    /// @param _maxRateLimitPerSecondMRL maximum rate limit per second per address in multi rate limited
+    /// @param _maxBufferCap maximum buffer cap in multi rate limited
```
ditchfieldcaleb, line 15: What do you mean by "in multi rate limited"? Confusing, would suggest reword.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+/// @title contract for putting a rate limit on how fast an address can perform an action e.g. Minting
+/// there are two buffers, one buffer which is each individual addresses's current buffer,
+/// and then there is a global buffer which is the buffer that each individual address must respect as well
+/// @author Fei Protocol
+contract MultiRateLimited is RateLimited, IMultiRateLimited {
+    using SafeCast for *;
+
+    /// @notice the struct containing all information per rate limited address
+    struct RateLimitData {
+        uint32 lastBufferUsedTime;
```
ditchfieldcaleb, line 20: Nit: if we change this to uint112 we'll pack the two slots perfectly. Right now the first slot is only using 176/255 bits.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+    uint256 public maxRateLimitPerSecond;
+
+    /// @notice max buffer cap allowable by non governor
+    uint256 public maxBufferCap;
+
+    constructor(
+        address coreAddress,
+        uint256 _maxRateLimitPerSecond,
+        uint256 _rateLimitPerSecond,
+        uint256 _maxRateLimitPerSecondMRL,
```
ditchfieldcaleb, line 39: What is MRL here? Suggest renaming this variable.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+    /// @notice rate limited address information
+    mapping (address => RateLimitData) public rateLimitPerAddress;
+
+    /// @notice max rate limit per second allowable by non governor
+    uint256 public maxRateLimitPerSecond;
+
+    /// @notice max buffer cap allowable by non governor
+    uint256 public maxBufferCap;
+
+    constructor(
```
ditchfieldcaleb, line 35: Add doc for constructor params.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    /// @notice rate limited address information
+    mapping (address => RateLimitData) public rateLimitPerAddress;
+
+    /// @notice max rate limit per second allowable by non governor
+    uint256 public maxRateLimitPerSecond;
+
+    /// @notice max buffer cap allowable by non governor
+    uint256 public maxBufferCap;
```
ditchfieldcaleb, line 33: Sorry for the back and forth here, but I suggest changing this to be more descriptive. Even after spending a couple of minutes looking over the code, I can't tell if this maximum applies to the global buffer or the per-contract buffers.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+    {
+        require(_maxBufferCap < _globalBufferCap, "MultiRateLimited: max buffer cap invalid");
+
+        maxRateLimitPerSecond = _maxRateLimitPerSecondMRL;
+        maxBufferCap = _maxBufferCap;
+    }
+
+    // ----------- Governor and Admin only state changing api -----------
+
+    /// @notice update the ADD_MINTER_ROLE rate limit per second
```
ditchfieldcaleb, line 55: Is this supposed to be a directive "you need to update the ADD_MINTER_ROLE..." or did you just miss an s, ie it should be "updates"?

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+    function updateMaxRateLimitPerSecond(uint256 newRateLimitPerSecond) external virtual override onlyGovernor {
+        require(newRateLimitPerSecond <= MAX_RATE_LIMIT_PER_SECOND, "MultiRateLimited: exceeds global max rate limit per second");
+
+        uint256 oldMaxRateLimitPerSecond = maxRateLimitPerSecond;
+        maxRateLimitPerSecond = newRateLimitPerSecond;
+
+        emit MultiMaxRateLimitPerSecondUpdate(oldMaxRateLimitPerSecond, newRateLimitPerSecond);
+    }
+
+    /// @notice update the ADD_MINTER_ROLE max buffer cap
```
ditchfieldcaleb, line 66: (See above comment)

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+    function updateMaxBufferCap(uint256 newBufferCap) external virtual override onlyGovernor {
+        require(newBufferCap <= bufferCap, "MultiRateLimited: exceeds global buffer cap");
+
+        uint256 oldBufferCap = maxBufferCap;
+        maxBufferCap = newBufferCap;
+
+        emit MultiBufferCapUpdate(oldBufferCap, newBufferCap);
+    }
+
+    /// @notice add an authorized rateLimitedAddress contract
```
ditchfieldcaleb, line 77: *adds

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+            require(_bufferCap <= maxBufferCap, "MultiRateLimited: max buffer cap exceeds non governor allowable amount");
+        }
+        require(_bufferCap <= bufferCap, "MultiRateLimited: buffercap too high");
+
+        _updateAddress(rateLimitedAddress, _rateLimitPerSecond, _bufferCap);
+    }
+
+    /// @notice add an authorized rateLimitedAddress contract
+    /// @param rateLimitedAddress the new address to add as a rateLimitedAddress
+    function addAddressWithCaps(
```
ditchfieldcaleb, line 109: The name & docstring of this function don't represent the (powerful) feature of this - that the added address by default uses the maximum possible rate limit & buffer cap. I don't really see the need for this function, but even so, if you want to keep it, I'd suggest either using function overloading or change the name to "addAddressWithMaximumParams". Preferably the latter.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+        address rateLimitedAddress,
+        uint112 newRateLimitPerSecond
+    ) external virtual override onlyGovernor {
+        require(newRateLimitPerSecond <= rateLimitPerSecond, "MultiRateLimited: rateLimitPerSecond too high");
+
+        _updateIndividualBufferStored(rateLimitedAddress);
+        _setIndividualRateLimitPerSecond(rateLimitedAddress, newRateLimitPerSecond);
+    }
+
+    /// @notice set the buffer cap
```
ditchfieldcaleb, line 138: "set the buffer cap" --> "set a contract's individual buffer cap"

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+        return rateLimitPerAddress[limiter].rateLimitPerSecond;
+    }
+
+    /// @notice the last time the buffer was used by each address
+    function getLastBufferUsedTime(address limiter) external override view returns(uint256) {
+        return rateLimitPerAddress[limiter].lastBufferUsedTime;
+    }
+
+    /// @notice the cap of the buffer that can be used at once
+    function getBufferCap(address limiter) external override view returns(uint256) {
```
ditchfieldcaleb, line 170: The address in question isn't the limiter, it's the limited.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+    /// @param rateLimitedAddress the address whose buffer will be returned
+    function individualBuffer(address rateLimitedAddress) public view override returns(uint144) {
+        RateLimitData memory rateLimitData = rateLimitPerAddress[rateLimitedAddress];
+
+        uint256 elapsed = block.timestamp - rateLimitData.lastBufferUsedTime;
+        return uint144(Math.min(rateLimitData.bufferStored + (rateLimitData.rateLimitPerSecond * elapsed), rateLimitData.bufferCap));
+    }
+
+    /// @notice the rate per second for each address
+    function getRateLimitPerSecond(address limiter) external override view returns(uint256) {
```
ditchfieldcaleb, line 160: The address in question isn't the limiter, it's the limited.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+        return rateLimitPerAddress[limiter].rateLimitPerSecond;
+    }
+
+    /// @notice the last time the buffer was used by each address
+    function getLastBufferUsedTime(address limiter) external override view returns(uint256) {
```
ditchfieldcaleb, line 165: The address in question isn't the limiter, it's the limited.

contracts/utils/MultiRateLimited.sol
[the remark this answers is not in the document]
ditchfieldcaleb: I get it I think, but this is confusing and will be harder to debug/screen for security issues. TBH I'd suggest removing the governor-rate-limit; if the governor can change the minter contract anyways (which it can) then it can bypass the governor-minting-rate-limit. Then we'll only have a non-governor maximum, which is a lot easier to process. In my opinion. @Joeysantoro would love to get your thoughts.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+        rateLimitData.bufferStored = _bufferCap;
+
+        emit IndividualRateLimitPerSecondUpdate(rateLimitedAddress, oldRateLimitPerSecond, _rateLimitPerSecond);
+    }
+
+    /// @param rateLimitedAddress the new address to add as a rateLimitedAddress
+    /// @param _rateLimitPerSecond the rate limit per second for this rateLimitedAddress
+    /// @param _bufferCap  the buffer cap for this rateLimitedAddress
+    function _addAddress(address rateLimitedAddress, uint112 _rateLimitPerSecond, uint144 _bufferCap) internal {
+        require(_bufferCap <= bufferCap, "MultiRateLimited: new buffercap too high");
```
ditchfieldcaleb, line 196: Took me awhile to realize that `bufferCap` is coming from the base RateLimited contract. I wish there were a way to alias this to `globalBufferCap` or something. Can you add a comment explaining the origin of this variable, and that it's the global one?

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+        });
+
+        rateLimitPerAddress[rateLimitedAddress] = rateLimitData;
+
+        emit IndividualRateLimitPerSecondUpdate(rateLimitedAddress, 0, _rateLimitPerSecond);
+    }
+
+    /// @notice the method that enforces the rate limit. Decreases buffer by "amount". 
+    /// If buffer is <= amount either
+    /// 1. Does a partial depletion by the amount remaining in the buffer or
```
ditchfieldcaleb, line 214: I really think we should disable partial depletions here. It's not a widely used feature, and disabling it closes off a whole branch of security analysis we won't have to do.

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+    /// 2. Reverts
+    /// Depending on whether doPartialAction is true or false
+    /// @param rateLimitedAddress the address whose buffer will be depleted
+    /// @param amount the amount to remove from the rateLimitedAddress's buffer
```
ditchfieldcaleb, line 218: Missing return param docstring

contracts/utils/IMultiRateLimited.sol
```diff
@@ -0,0 +1,61 @@
...
+    /// @notice emitted when a buffer cap is updated
+    event IndividualBufferCapUpdate(address rateLimitedAddress, uint256 oldBufferCap, uint256 newBufferCap);
+    
+    /// @notice emitted when rate limit is updated
+    event IndividualRateLimitPerSecondUpdate(address rateLimitedAddress, uint256 oldRateLimitPerSecond, uint256 newRateLimitPerSecond);
+
+    /// @notice emitted when the sub gov buffer cap max is updated
+    event MultiBufferCapUpdate(uint256 oldBufferCap, uint256 newBufferCap);
+
+    /// @notice emitted when the sub gov buffer rate limit per second max is updated
```
ditchfieldcaleb, line 31: Remove stale reference to sub governor.

contracts/utils/IMultiRateLimited.sol
```diff
@@ -0,0 +1,61 @@
...
+    event MultiMaxRateLimitPerSecondUpdate(uint256 oldMaxRateLimitPerSecond, uint256 newMaxRateLimitPerSecond);
+
+    /// @notice update the sub gov max rate limit per second
```
ditchfieldcaleb, line 34: Remove stale reference to sub governor.

contracts/utils/IMultiRateLimited.sol
```diff
@@ -0,0 +1,61 @@
...
+    function updateMaxRateLimitPerSecond(uint256 newMaxRateLimitPerSecond) external;
+
+    /// @notice update the sub gov max buffer cap
```
ditchfieldcaleb, line 37: Remove stale reference to sub governor.

contracts/utils/MultiRateLimited.sol
ElliotFriedman: multi rate limited

ElliotFriedman:
> Concerns:
> 
> * Really one one big one. The architecture seems a bit convoluted here - why do we have GlobalRateLimitedMinter if it doesn't do anything? Shouldn't this be the contract that is RateLimited? Currently all of the global logic for rate limiting is in MultiRateLimitedMinter, which seems to make the GlobalRateLimitedMinter obsolete. I suggest either putting all of the global stuff in the GlobalRateLimitedMinter (decent amount of work, but cleaner), or just removing the GlobalRateLImitedMinter that you currently have and moving that one function override to the current MultiRateLimitedMinter, and then changin the name to GlobalMultiRateLimitedMinter. Hopefully that made sense

The whole reason for splitting out MultiRateLimited into a contract separate from GlobalRateLimitedMinter is that the next version of the Peg Stability Module is going to leverage this MultiRatedLimited contract to limit withdraws from a PCV deposit. We can use this contract for rate limiting many different things as a pattern in so many other places, but we don't get that benefit if we just squash all of the logic together into the GlobalRateLimitedMinter contract.

contracts/utils/MultiRateLimited.sol
ElliotFriedman: perfect is the enemy of the good

contracts/utils/MultiRateLimited.sol
ElliotFriedman: the comment literally says what the variable is.

contracts/utils/GlobalRateLimitedMinter.sol
```diff
@@ -0,0 +1,46 @@
...
+import "./IGlobalRateLimitedMinter.sol";
+
+/// @notice global contract to handle rate limited minting of FEI on a global level
+/// allows whitelisted minters to call in and specify the address to mint FEI to within
+/// that contract's limits
+contract GlobalRateLimitedMinter is MultiRateLimited, IGlobalRateLimitedMinter {
+    /// @param coreAddress address of the core contract
+    /// @param _globalMaxRateLimitPerSecond maximum amount of fei that can replenish per second ever, this amount cannot be changed by governance
+    /// @param _globalRateLimitPerSecond maximum rate limit per second per address
+    /// @param _maxRateLimitPerSecondMRL maximum rate limit per second per address in multi rate limited
```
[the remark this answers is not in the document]
ElliotFriedman: it means maximum rate limit per second that replenishes per address in multi rate limited

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: No it doesn't? Again: "I can't tell if this maximum applies to the global buffer or the per-contract buffers"

contracts/utils/MultiRateLimited.sol
```diff
@@ -32,6 +32,13 @@ contract MultiRateLimited is RateLimited, IMultiRateLimited {
     /// @notice max buffer cap allowable by non governor
     uint256 public maxBufferCap;
 
+    /// @param coreAddress address of the core contract
+    /// @param _maxRateLimitPerSecond maximum amount of fei that can replenish per second ever, this amount cannot be changed by governance
+    /// @param _rateLimitPerSecond maximum rate limit per second per address
+    /// @param _maxRateLimitPerSecondMRL maximum rate limit per second per address in multi rate limited
```
ditchfieldcaleb, line 38: This var name doesn't make much sense to me; instead of "the maximum in [some sub-contract that we don't care about]", the name should reflect the functionality of the var. Something like "_maxRateLimitPerSecondPerAddress" would clear this up quickly

contracts/utils/IMultiRateLimited.sol
ditchfieldcaleb: @ElliotFriedman why was this marked resolved?

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: catchphrases are the enemy of communication efficiency ;)

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: If it doesn't allow partial depletion, then the comment should say that.
Actually, if this functionality doesn't exist at all, why is it a parameter being passed to MRL?

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,273 @@
...
+
+import {CoreRef} from "../refs/CoreRef.sol";
+import {TribeRoles} from "./../core/TribeRoles.sol";
+import {RateLimited} from "./RateLimited.sol";
+import {IMultiRateLimited} from "./IMultiRateLimited.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
+
+/// @title contract for putting a rate limit on how fast an address can perform an action e.g. Minting
+/// there are two buffers, one buffer which is each individual addresses's current buffer,
```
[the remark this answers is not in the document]
ditchfieldcaleb: Sure, then the phrasing would be:

"There is one buffer per contract that tracks the contracts usage (and limits it), and one global buffer that tracks (and limits) usage globally."

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: (closing, will fix in my nitpicks pr)

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: (closing, will fix in my nitpicks pr)

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: (closing, will fix in my nitpicks pr)

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: (closing, will fix in my nitpicks pr)

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: (closing, will fix in my nitpicks pr)

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: (closing, will fix in my nitpicks pr)

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: (closing, will fix in my nitpicks pr)

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: (closing, will fix in my nitpicks pr)

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: (closing, will fix in my nitpicks pr)

contracts/utils/MultiRateLimited.sol
ditchfieldcaleb: (closing, will fix in my nitpicks pr)

contracts/utils/MultiRateLimited.sol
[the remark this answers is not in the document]
ditchfieldcaleb: (closing, will fix in my nitpicks pr)

contracts/utils/GlobalRateLimitedMinter.sol
```diff
@@ -0,0 +1,64 @@
...
+    {
+        _depleteBuffer(msg.sender, amount);
+        _mintFei(to, amount);
+    }
+
+    /// @notice mint fei to the target address and deplete the whole rate limited
+    ///  minter's buffer, pausable and completely depletes the msg.sender's buffer
+    /// @param to the recipient address of the minted FEI
+    /// mints all FEI that msg.sender has in the buffer
+    function mintMaxAllowableFei(address to)
```
Joeysantoro, line 53: cool

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,442 @@
...
+    /// @notice max rate limit per second allowable by non governor per contract
+    uint256 public maxRateLimitPerSecond;
+
+    /// @notice max buffer cap allowable by non governor per contract
+    uint256 public maxBufferCap;
+
+    /// @param coreAddress address of the core contract
+    /// @param _maxRateLimitPerSecond maximum amount of fei that can replenish per second ever, this amount cannot be changed by governance
+    /// @param _rateLimitPerSecond maximum rate limit per second per address
+    /// @param _maxRateLimitPerSecondMRL maximum rate limit per second per address in multi rate limited
```
Joeysantoro, line 38: `individualMaxRateLimitPerSecond` is clearer to me

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,442 @@
...
+    /// @param _maxBufferCap maximum buffer cap in multi rate limited
```
Joeysantoro, line 39: `individualMaxBufferCap` is clearer to me

contracts/utils/MultiRateLimited.sol
Joeysantoro: on global param too

contracts/utils/MultiRateLimited.sol
Joeysantoro: on global param too

contracts/utils/MultiRateLimited.sol
```diff
@@ -0,0 +1,442 @@
...
+        override
+        returns (uint144)
+    {
+        RateLimitData memory rateLimitData = rateLimitPerAddress[
+            rateLimitedAddress
+        ];
+
+        uint256 elapsed = block.timestamp - rateLimitData.lastBufferUsedTime;
+        return
+            uint144(
```
Joeysantoro, line 241: safecast?

ElliotFriedman: Safecast doesn't go to this data size

contracts/utils/MultiRateLimited.sol
ElliotFriedman: done

contracts/utils/MultiRateLimited.sol
ElliotFriedman: done

PR #516, opened 2022-02-01
Fix collateralization oracle tests & disable buybacks tests

ditchfieldcaleb:


test/integration/tests/buybacks.ts
```diff
@@ -17,7 +17,7 @@ import { BalancerLBPSwapper, CollateralizationOracle, IVault, IWeightedPool } fr
 import { forceEth } from '../setup/utils';
 const toBN = ethers.BigNumber.from;
 
-describe('e2e-buybacks', function () {
```
eswak, line 20: instead of skipping, you could add a before() hook that checks if the PCVEquityMinter is paused, and if so, unpause it

ditchfieldcaleb: good idea

eswak (approved):

PR #521, opened 2022-02-04
Tribe Solmate Auth

Joeysantoro:
https://docs.google.com/document/d/1AmtEmuJb2RGPHGOgbm2E9OJDjy4wCWrHZ0IjGbHO__M/edit#


contracts/core/TribeAuth.sol
```diff
@@ -0,0 +1,58 @@
...
+import "../external/solmate/src/auth/authorities/MultiRolesAuthority.sol";
+import "@openzeppelin/contracts/security/Pausable.sol";
+
+abstract contract TribeAuth is TribeRoles, Auth, Pausable {
+
+    IFei public immutable fei;
+
+    IERC20 public immutable tribe;
+
+    constructor(Authority _authority, IFei _fei, IERC20 _tribe)
```
ditchfieldcaleb, line 16: Should we go ahead and hardcode in fei & tribe as a constant? If they're going to immutable I don't see a reason to include them as params.

contracts/core/TribeAuth.sol
```diff
@@ -0,0 +1,58 @@
...
+import "@openzeppelin/contracts/security/Pausable.sol";
+
+abstract contract TribeAuth is TribeRoles, Auth, Pausable {
+
+    IFei public immutable fei;
+
+    IERC20 public immutable tribe;
+
+    constructor(Authority _authority, IFei _fei, IERC20 _tribe) 
+        Auth(address(0), _authority)
```
ditchfieldcaleb, line 17: Why not set the DAO as the owner?

contracts/core/TribeAuthority.sol
```diff
@@ -0,0 +1,37 @@
...
+    mapping(uint8 => bytes32) public getRoleAdmins;
+
+    // TODO add a way to assign admins and a constructor
+    
+    /*///////////////////////////////////////////////////////////////
+                      USER ROLE ASSIGNMENT LOGIC
+    //////////////////////////////////////////////////////////////*/
+
+    function isRoleAdmin(address admin, uint8 role) public view returns(bool) {
+        return (getRoleAdmins[role] & getUserRoles[admin]) != bytes32(0);
```
ditchfieldcaleb, line 18: Why bitwise AND here?

ditchfieldcaleb (commented):
Didn't have time to fully review but here are my comments so far.

contracts/core/TribeAuth.sol
ElliotFriedman: unit tests will break if you hardcode these vars

contracts/core/CoreAuth.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "./TribeAuth.sol";
+import "./IPermissionsRead.sol";
```
ElliotFriedman, line 5: This contract should be in the ref folder and should be named AuthRef

contracts/core/TribeAuthority.sol
ElliotFriedman: solmate uses bit flags and shifting for storing info about user roles, so it's checking to see if a user has an admin role by checking to see if they both have that same bit flipped

contracts/core/TribeAuthority.sol
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    function setUserRole(
+        address user,
+        uint8 role,
+        bool enabled
+    ) public override {
```
ElliotFriedman, line 25: why do we need to override this function?

Joeysantoro:
Closing in favor of #522

PR #522, opened 2022-02-04
tribe roles

Joeysantoro:


contracts/refs/CoreRef.sol
```diff
@@ -9,6 +9,8 @@ import "@openzeppelin/contracts/security/Pausable.sol";
 /// @notice defines some modifiers and utilities around interacting with Core
 abstract contract CoreRef is ICoreRef, Pausable {
     ICore private _core;
```
ElliotFriedman, line 11: While we're here, we should make core immutable as well to save on gas

Joeysantoro: I'm tentatively down if we are sure we won't want to do core upgrades in the future

ElliotFriedman (approved):

contracts/refs/CoreRef.sol
ditchfieldcaleb: I think core should be able to be set by governance

ditchfieldcaleb (changes requested):
Generally fine, two things:

1) I think core should be able to be set by governance
2) The roles library - I think it should be a contract where we can add and remove roles, otherwise we're stuck with a set of roles that is essentially immutable unless we switch out core

Joeysantoro:
> * I think core should be able to be set by governance

Lets discuss this as a team, I'm open for either.

> The roles library - I think it should be a contract where we can add and remove roles, otherwise we're stuck with a set of roles that is essentially immutable unless we switch out core

This is actually not an issue, all our contracts have immutable ACL so if we want to change the ACL we can simply deploy a new contract with an updated library. 

If we want to "remap" a role, we can use the admin functionality but I'm hoping to move away from that tbh. Remap is not an operation we should do often

ditchfieldcaleb (dismissed):

thomas-waite (commented):
Code looks good, but I'm wary of removing the ability to upgrade `core` in the future. 

Are we certain we won't need to adjust anything here in the future? If not, why not maintain optionality by keeping it upgradeable? I like the idea of discussing governance being able to set it

thomas-waite (approved):
As discussed, looks good!

PR #526, opened 2022-02-07
Add Forge

thomas-waite:
# Summary
Incorporates Forge - a DappTools based Solidity development framework - into the repo. This allows us to write unit tests in Solidity. Forge docs: https://onbjerg.github.io/foundry-book/ 

## Motivation
Forge allows for the following:
- Write smart contract unit tests in Solidity, so you never have to leave the language you're developing in
- Allows for easy function fuzzing (by passing arguments to the test function - these people the arguments that are fuzzed)
- It's faster to compile! Compile time comparison for Fei repo: Hardhat (~50s), Forge (~35s)

## Description
To get started you need to install Forge. This is done with `foundryup` which installs nightly releases. Use:

`curl -L https://foundry.paradigm.xyz | bash`

then install the latest nightly release: `foundryup`.

On a Mac, you might see this error: `dyld: Library not loaded: /usr/local/opt/libusb/lib/libusb-1.0.0.dylib`. In that case, install `libusb` with `brew install libusb`

### Structure
- Solidity unit tests are in a `test` directory in `contracts`
- A `utils` directory has been added to include a base `DSTest.sol` contract, a set of `Fixtures` to easily setup Fei tests and an interface to a special Forge `Vm` contract. The `Vm` contract allows you to manipulate VM state easily for tests, such as by advancing block time, changing `msg.sender` etc

### Quick start
- Compile contracts: `forge build` or `npm run compile`
- Run unit tests: `forge test` or `npm run test`

### Example test
I have ported the `Fei.test.ts` setup and one of the unit tests over to Forge to give an example of how to use this framework. It is `Fei.t.sol`


contracts/external/TransparentUpgradeableProxy.sol
```diff
@@ -1,2 +1,4 @@
+pragma solidity ^0.8.4;
```
thomas-waite, line 1: Forge will throw a compilation error if a `.sol` file doesn't include the `pragma` declaration. I think fine to add + probably best practice for all solidity files to specify their version

contracts/test/utils/Fixtures.sol
```diff
@@ -0,0 +1,59 @@
+import {Core} from "../../core/Core.sol";
```
ElliotFriedman, line 1: why does github not highlight this solidity?

thomas-waite: Sorted, missing `pragma` statement

contracts/external/TransparentUpgradeableProxy.sol
thomas-waite: Have looked and there isn't currently a way to do this with `forge`. 

Do you have an example of an external file that we are likely to import which doesn't have a `pragma solidity` statement? Surely all external contracts must have it in order for those contracts to be compilable in their home repos?

If this does become a strong need, we can ask the `forge` team about potentially putting it on the roadmap/see if others have the requirement

Joeysantoro (approved):

Joeysantoro (approved):

PR #527, opened 2022-02-07
Add e2e & config file for TribalChief

eswak:
I tried to make the error messages somewhat clear, to improve our debugging experience. Examples :
- Staked token "dai" has wrong address. Did you mean "fei3CrvStakingtokenWrapper" ? This contract is what is actually staked on the TribalChief pool # 14.
- AssertionError: TribalChief pool # 13 "d3StakingTokenWrapper" has 250 allocPoint on-chain, but 9250 in the config file.
- TribalChief pool # 7 "stakingTokenWrapperSYNLaaS" has unlocked=false on-chain, but true in the config file.


protocol-configuration/tribalchief.ts
```diff
@@ -0,0 +1,20 @@
+export default {
```
ditchfieldcaleb, line 1: Can you make an interface or type (preferably type) for this object? Will help devs to know what it is / how to use it.

test/integration/tests/tribalchief.ts
```diff
@@ -0,0 +1,100 @@
...
+
+      const poolInfo = await contracts.tribalChief.poolInfo(i);
+      const stakedToken = await contracts.tribalChief.stakedToken(i);
+      let stakedTokenName = null;
+      for (const key in mainnetAddresses) {
+        if (mainnetAddresses[key].address == stakedToken) {
+          stakedTokenName = key;
+        }
+      }
+      const errmsg2 =
```
ditchfieldcaleb, line 62: It'd be easier to use tilde-evaluation here instead of concatenation:

`Staked token ${documentedPoolsKeys[i]} has wrong address...etc`

ditchfieldcaleb (changes requested):

test/integration/tests/fip-38-tokemak.ts
```diff
@@ -67,7 +67,7 @@ describe.skip('e2e-fip-38-tokemak', function () {
 
     // Advance block by 1 tokemak cycle
     const currentBlock = await time.latestBlock();
-    await time.advanceBlockTo(currentBlock + 6400);
+    await time.advanceBlockTo(currentBlock + 6460);
```
thomas-waite, line 70: Why the extra 60 blocks?

test/integration/tests/fip-38-tokemak.ts
```diff
@@ -85,58 +85,4 @@ describe.skip('e2e-fip-38-tokemak', function () {
       ethers.utils.parseEther('10000')
     );
   });
-
-  it('should be able to deposit TOKE, and withdraw', async function () {
```
thomas-waite, line 89: Why remove this test?

test/integration/tests/pcv.ts
```diff
@@ -110,20 +110,26 @@ describe('e2e-pcv', function () {
 
   /// pause this test as it has been disabled for FIP-62
```
thomas-waite, line 111: Remove this comment?

test/integration/tests/tribalchief.ts
```diff
@@ -0,0 +1,100 @@
...
+import { TestEndtoEndCoordinator } from '@test/integration/setup';
+import mainnetAddresses from '@protocol/mainnetAddresses';
+import tribalchief from '@protocol/tribalchief';
+
+describe('e2e-tribalchief', function () {
+  let contracts: NamedContracts;
+  let e2eCoord: TestEndtoEndCoordinator;
+  const doLogging = Boolean(process.env.LOGGING);
+
+  before(async () => {
```
thomas-waite, line 16: I've seen this pattern elsewhere, not specific to this PR, but why do we set `chai.use()` in a `before()` block? Why not just at the top level beneath imports? It's test infrastructure setup, rather than unit test logic

thomas-waite (commented):
Looks good to me overall, some minor questions/points

test/integration/tests/fip-38-tokemak.ts
eswak: turns out the Tokemak cycles are 6460 blocks, and if the tests were running during these 60 blocks, the test could fail & we didn't understand why.

test/integration/tests/tribalchief.ts
eswak: ok

protocol-configuration/tribalchief.ts
eswak: ok

test/integration/tests/fip-38-tokemak.ts
eswak: we don't use a TOKE deposit, the contract has been deployed on mainnet but never used.

the metagovernance PR will use another deposit, so we don't need to keep trace of the existence of this TOKE contract that was never used for anything :p

test/integration/tests/pcv.ts
eswak: ok

test/integration/tests/tribalchief.ts
eswak: it may have something to do with tests running in //, idk

test/integration/tests/fip-38-tokemak.ts
thomas-waite: Makes sense!

test/integration/tests/fip-38-tokemak.ts
thomas-waite: Thanks for explaining

test/integration/tests/tribalchief.ts
```diff
@@ -0,0 +1,75 @@
...
+    expect(totalAllocPoint).to.be.equal(totalAllocPointDocumented);
+  });
+
+  it('should properly describe the current pools config', async function () {
+    const numPools = await contracts.tribalChief.numPools();
+    const documentedPoolsKeys = Object.keys(tribalchief);
+    const documentedPoolsValues = Object.values(tribalchief);
+    for (let i = 0; i < numPools && i < documentedPoolsKeys.length; i++) {
+      const documentedStakedContract = mainnetAddresses[documentedPoolsKeys[i]];
+      const errmsg1 = `Staked token "${documentedPoolsKeys[i]}" is not found in mainnetAddresses.ts`;
```
ElliotFriedman, line 51: is this the actual error when you dereference an object that doesn't exist?

Joeysantoro (approved):

test/integration/tests/tribalchief.ts
ElliotFriedman: I would think the error message would be something like 'property x is undefined'

ElliotFriedman (approved):
LGTM!

Joeysantoro (approved):

PR #528, opened 2022-02-07
TribeMetadata

ditchfieldcaleb:


contracts/utils/TribeMetadata.sol
```diff
@@ -0,0 +1,34 @@
...
+abstract contract TribeMetadataV1 {
+    struct TribeMetadata {
+        uint8 major;
+        uint8 minor;
+        uint8 patch;
+        string name;
+        string domain;
+    }
+
+    TribeMetadata public metadata;
```
ElliotFriedman, line 19: this should be immutable

ElliotFriedman:
Closing this out as we're going to use the ENS to store contract metadata

PR #535, opened 2022-02-10
Add backup oracles to PSMs

thomas-waite:
# Summary
Adds UniswapV3 TWAP backup oracles to the `PegStabilityModules` (PSMs) for `DAI` and `ETH`. These both have highly liquid `USDC` pools on Uniswap V3:
- `DAI-USDC`: ~$300m TVL
- `USDC-ETH`: ~$285m TVL

## Description
A `UniswapV3OracleWrapper` has been created which implements `IOracle.sol`. Each asset requires its own `UniswapV3OracleWrapper` contract to be deployed.

### Calculating oracle price and siloing out Uniswap logic
The Maths of calculating the oracle price is siloed out into `UniswapWrapper.sol`. This is mostly done because the UniswapV3 contracts are written in a version, `>=0.5.0 <0.8.0`, which is incompatible with the Fei codebase. They rely on  underflows and overflows happening in some of their accounting logic, which in >0.8.0 causes reverts.

### Oracle config and safety
The oracles settings are controlled by an `OracleConfig` that is injected in. It has the following form:

```
struct OracleConfig {
    uint32 twapPeriod;
    uint128 minPoolLiquidity;
    address uniswapPool;
    uint256 precision;
  }
```

The `minPoolLiquidity` config is a safety parameter that is used to enforce a minimum level of liquidity in the relevant pool in order to provide a valid reading.

### Testing 
Integration tests which fork `mainnet` have been added to the `contracts/test/integration` folder. Five oracles have been deployed against the corresponding pools:
- `daiUsdcPool`
- `usdcEthPool`
-  `wbtcUsdcPool`
- `usdtAvinocPool`
- `usdcFeiPool`

The prices returned by the relevant TWAP oracles have been validated manually against the relevant pool spot price. These 5 in particular were chosen because they have parameters (token0, token1, inputNumDecimals, outputNumDecimals) which allow all 4 paths of the Uniswap maths calculation logic to be tested. Details are outlined in the integration tests.

## Concern
I think there is a broken abstraction in `OracleRef.sol` where the decimal normalisation and inversion occurs globally and applies to both the primary and backup oracle price results. 

As in this PR, these considerations should be pushed down into the relevant oracle contract. These factors likely need to be taken into account in different ways for different oracles depending on how they calculate the price. For example in calculating the Uni TWAP price, we need to know whether we are inverting when we are calculating the `numerator` and `denominator`.

If this UniV3 backupOracle was to be set, and `setDecimalsNormalizer()` or `setDoInvert()` were later seperately called on `OracleRef.sol` - it would result in the backup oracle `OracleRef.sol` reporting the incorrect price due to this global post processing.

Before deploying this PR, I believe we should:
1. Remove `setDecimalsNormalizer()` or `setDoInvert()` from `OracleRef.sol` and push these considerations down into the primary Chainlink oracle contract


contracts/test/oracle/UniswapMathWrapper.t.sol
```diff
@@ -0,0 +1,24 @@
...
+
+  function setUp() public {
+    // Mock has an observe() method that returns predefined test values. Used to ensure price > 0
+    mockPool = new MockUniV3Pool();
+    uniswapMathWrapper = new UniswapMathWrapper();
+  }
+
+  function testPrice() public {
+    uint256 price = uniswapMathWrapper.calculatePrice(address(mockPool), secondsAgo);
+    assertGt(price, 0);
```
thomas-waite, line 22: Difficult to unit test the exact price calculated. This function itself is just calling three Uniswap library methods for oracles:

```
(int24 arithmeticMeanTick, ) = OracleLibrary.consult(pool, secondsAgo);
uint160 sqrtPrice = TickMath.getSqrtRatioAtTick(arithmeticMeanTick);
uint256 price = FullMath.mulDiv(sqrtPrice, sqrtPrice, FixedPoint96.Q96);
```

Adding an e2e test to compare oracle performance with our primary oracle, Chainlink, as a means of validation. Also potentially worthwhile getting the implementation reviewed by a Uniswap team member

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
```diff
@@ -0,0 +1,71 @@
...
+  /// Does this check belong here? Maybe ideally in whatever executes the trade?
+  function validatePrice(uint256 rawPrice) internal view {}
+
+  function validateNumTWAPSlots() internal view {}
+
+
+  // ----------- Governor only state changing api -----------
+
+  /// @notice Change the time period over which the TWAP price is calculated
+  function setSecondsAgo(uint32 _secondsAgo) external onlyGuardianOrGovernor {
```
thomas-waite, line 67: May be necessary as a safety mechanism to change the duration over which the TWAP price is calculated

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
```diff
@@ -0,0 +1,84 @@
...
+  /// @notice Validate that the slippage caused by a proposed trade is acceptable
+  /// Does this check belong here? Maybe ideally in whatever executes the trade?
+  function validatePrice(uint256 rawPrice) internal view returns (bool) {
+    return true;
+  }
+
+  /// @notice Validate that the UniswapV3 pool has sufficient observation slots
+  /// to support the requested TWAP period. If not revert and 
+  //  `pool.increaseObservationCardinalityNext()` should be called
+  function validateNumOberservationSlots(uint32 _secondsAgo) internal view {
```
Joeysantoro, line 64: does this need to be checked every single time? Or just once at construction

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
```diff
@@ -0,0 +1,84 @@
...
+  function update() external view override {}
+
+  // ----------- Getters -----------
+
+  function read() external view override returns (Decimal.D256 memory, bool) {
+    validateNumOberservationSlots(secondsAgo);
+
+    uint256 rawPrice = uniswapWrapper.calculatePrice(pool, secondsAgo);
+
+    bool valid = !paused() && validatePrice(rawPrice);
```
Joeysantoro, line 45: validatePrice doesn't do anything, is there a reason to keep it?

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
```diff
@@ -0,0 +1,84 @@
...
+  /// @notice updates the oracle price
+  /// @dev no-op, Uniswap is updated automatically
+  function update() external view override {}
+
+  // ----------- Getters -----------
+
+  function read() external view override returns (Decimal.D256 memory, bool) {
+    validateNumOberservationSlots(secondsAgo);
+
+    uint256 rawPrice = uniswapWrapper.calculatePrice(pool, secondsAgo);
```
Joeysantoro, line 43: how are decimals handled in this calculation?

contracts/oracle/uniswap/UniswapWrapper.sol
```diff
@@ -0,0 +1,26 @@
...
+  function calculatePrice(address pool, uint32 secondsAgo) external view returns (uint256) {
+    (int24 arithmeticMeanTick, ) = OracleLibrary.consult(pool, secondsAgo);
+    uint160 sqrtPrice = TickMath.getSqrtRatioAtTick(arithmeticMeanTick);
+    uint256 price = FullMath.mulDiv(sqrtPrice, sqrtPrice, FixedPoint96.Q96);
+    return price;
+  }
+
+  function getObservationCardinality(address pool) external view returns (uint16) {
+    (, , , uint16 observationCardinality, , , ) = IUniswapV3Pool(pool).slot0();
+    return observationCardinality;
```
Joeysantoro, line 23: this logic should probably just be in the UniswapV3OracleWrapper to save gas

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
thomas-waite: In progress, I'm going to flesh it out with something - ideally as a way to check slippage

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
thomas-waite: On the todo - adding in

contracts/oracle/uniswap/UniswapWrapper.sol
thomas-waite: As discussed agreed, but the Uniswap contracts are compiled with an incompatible version of Solidity. Cleanest solution imo was to seperate out

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
thomas-waite: Have changed to immediately add a sufficient number of observation slots on oracle deployment. That way we don't need to make a seperate off-chain transaction to add slots

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
```diff
@@ -0,0 +1,189 @@
...
+  /// @param _twapPeriod Time period over which the mean price is calculated
+  function validateTwapPeriod(uint32 _twapPeriod) internal pure {
+    require(_twapPeriod > 0, "Twap period must be greater than 0");
+  }
+
+  /// @notice Increase pool observation cardinality to support requested TWAP period
+  function addSupportForPool(address _pool, uint32 _twapPeriod, uint32 _meanBlockTime) internal {
+    uint16 requiredCardinality = uint16(_twapPeriod / _meanBlockTime) + uint16(10); // Add additional number of slots to ensure available
+
+    IUniswapV3Pool(_pool).increaseObservationCardinalityNext(requiredCardinality);
```
thomas-waite, line 164: What happens if cardinality is already sufficient?

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
```diff
@@ -0,0 +1,189 @@
...
+  function sortTokensAccordingToUniswap(address tokenA, address tokenB) internal pure returns (address, address) {
+    (address token0, address token1) = tokenA < tokenB ? (tokenA, tokenB) : (tokenB, tokenA);
+    return (token0, token1);
+  }
+
+  // ----------- Governor only state changing api -----------
+
+  /// @notice Change the time period over which the TWAP price is calculated
+  /// @param _twapPeriod Period of time over which time weighted average price is calculated
+  function setTwapPeriod(uint32 _twapPeriod) external onlyGuardianOrGovernor {
```
thomas-waite, line 182: Change to new TRIBE roles: governor, guardian and a new one oracle admin

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
```diff
@@ -0,0 +1,189 @@
...
+/// @title UniswapV3 TWAP Oracle wrapper
+/// @notice Reads a UniswapV3 TWAP oracle, based on a single Uniswap pool, and wraps it under 
+/// the standard Fei interface
+contract UniswapV3OracleWrapper is IOracle, CoreRef {
+  using Decimal for Decimal.D256;
+  using SafeCast for uint256;
+
+  /// @notice Uniswap V3 pool which the oracle is built on
+  address public pool;
+
```
thomas-waite, line 23: Make state variables immutable

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
```diff
@@ -0,0 +1,189 @@
...
+import {CoreRef} from "../../refs/CoreRef.sol";
+import {Decimal} from "../../external/Decimal.sol";
+import {IOracle} from "../IOracle.sol";
+import {IUniswapWrapper} from "./IUniswapWrapper.sol";
+
+/// @title UniswapV3 TWAP Oracle wrapper
+/// @notice Reads a UniswapV3 TWAP oracle, based on a single Uniswap pool, and wraps it under 
+/// the standard Fei interface
+contract UniswapV3OracleWrapper is IOracle, CoreRef {
```
thomas-waite, line 17: Add interface for wrapper

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
```diff
@@ -0,0 +1,189 @@
...
+
+    bool valid = !paused();
+    
+    // Divide out the extra decimals that decimal library adds
+    Decimal.D256 memory value = Decimal.from(rawPrice).div(10**18);
+    return (value, valid);
+  }
+
+  /// @notice no-op, Uniswap V3 constantly updates the price
+  function isOutdated() external view override returns (bool) {}
```
thomas-waite, line 126: Consider returning explicitly returning `false`

contracts/oracle/uniswap/UniswapWrapper.sol
```diff
@@ -0,0 +1,110 @@
...
+    address pool,
+    uint32 twapPeriod,
+    address oracleInputToken,
+    address oracleOutputToken,
+    uint8 inputTokenDecimals,
+    uint8 outputTokenDecimals,
+    uint256 precision
+  ) external view returns (uint256) {
+    (address token0, address token1) = sortTokensAccordingToUniswap(oracleInputToken, oracleOutputToken);
+    bool invertTick = token0 == oracleInputToken ? false : true;
```
thomas-waite, line 41: Change to `invertPrice` rather than `invertTick`

contracts/test/integration/backupOracles.sol
```diff
@@ -0,0 +1,208 @@
...
+    });  
+
+    wbtcOracle = new UniswapV3OracleWrapper(
+      address(core),
+      wbtc, // Note, the order here matters. First token should be the input (i.e. PSM asset), not dollar tracker
+      usdc,
+      uniswapMathWrapper,
+      wbtcUsdcOracleConfig
+    );
+
```
thomas-waite, line 103: Test path where not inverting tick, but are inverting decimal normaliser

test/integration/proposals_config.ts
```diff
@@ -11,6 +13,15 @@ const proposals: ProposalsConfigMap = {
         proposal: fip_xx_proposal // full proposal file, imported from '@proposals/description/fip_xx.ts'
     }
     */
+  fip_78: {
+    deploy: true,
+    proposalId: '',
+    affectedContractSignoff: ['uniswapV3OracleWrapper', 'uniswapWrapper'],
```
thomas-waite, line 19: Include DAI PSM, EthPSM etc

test/integration/tests/backupOracles.ts
```diff
@@ -0,0 +1,101 @@
...
+  });
+
+  it('should read reasonable Uniswap V3 TWAP oracle price', async () => {
+    const [daiPrice, isValid] = await daiUsdcBackupOracle.read();
+    expect(isValid).to.be.true;
+    expectApprox(daiPrice.toString(), BNe18(1).toString());
+  });
+
+  it('should read Uniswap V3 TWAP oracle price comparable to primary oracle price', async () => {
+    const primaryChainlinkOracle = contracts.chainlinkDaiUsdOracleWrapper;
```
thomas-waite, line 95: Check price of eth perhaps against primary

test/integration/tests/backupOracles.ts
```diff
@@ -0,0 +1,101 @@
...
+    const primaryOraclePrice = await primaryChainlinkOracle.read();
+    const backupOraclePrice = await daiUsdcBackupOracle.read();
+
```
thomas-waite, line 98: `pause()` chainlink contract, to get main system to fall back to backupOracle. The check all works

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
thomas-waite: Method is a no-op if the pool already has an observationCardinalityNext greater than the increase

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
```diff
@@ -0,0 +1,193 @@
...
+  /// @param _pool Uniswap pool address
+  /// @param _inputToken Input token for which a price is being quoted
+  /// @param _outputToken Output token
+  function validateTokensInPool(address _pool, address _inputToken, address _outputToken) internal view {
+    address uniswapToken0 = IUniswapV3Pool(_pool).token0();
+    address uniswapToken1 = IUniswapV3Pool(_pool).token1();
+    (address token0, address token1) = sortTokensAccordingToUniswap(_inputToken, _outputToken);
+  
+    require(
+      uniswapToken0 == token0 || uniswapToken1 == token1,
```
eswak, line 153: should be && ?

contracts/refs/OracleRef.sol
```diff
@@ -92,6 +92,8 @@ abstract contract OracleRef is IOracleRef, CoreRef {
 
         // Scale the oracle price by token decimals delta if necessary
         uint256 scalingFactor;
+
+        // TODO: Broken abstraction, this needs pushing down
```
eswak, line 96: I like being able to invert & normalize decimals at this level, but the fact that they can be set after deployment (setDoInvert and setDecimalsNormalizer) is weird, because for a given deployment that should be fixed. These functions are governor-only, so I guess that's fine.

eswak: Let's discuss this oracle topic at the next dev meeting

contracts/test/unit/oracle/UniswapV3OracleWrapper.t.sol
```diff
@@ -0,0 +1,157 @@
...
+  FeiTestAddresses addresses = getAddresses();
+  Vm public constant vm = Vm(HEVM_ADDRESS);
+
+  MockERC20 private tokenA = new MockERC20();
+  MockERC20 private tokenB = new MockERC20();
+  
+  
+  function setUp() public {
+    core = getCore();
+    mockUniswapPool = IMockUniswapV3Pool(deployCode("./out/MockUniV3Pool.sol/MockUniV3Pool.json"));
```
eswak, line 38: why this weird syntax ?

contracts/test/unit/oracle/UniswapV3OracleWrapper.t.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    UniswapV3OracleWrapper.OracleConfig memory oracleConfig = UniswapV3OracleWrapper.OracleConfig({
+      twapPeriod: twapPeriod,
+      uniswapPool: address(mockUniswapPool),
+      minPoolLiquidity: mockUniswapPool.liquidity(),
+      precision: precision
+    });    
+    
+    
+    oracle = new UniswapV3OracleWrapper(
```
eswak, line 50: Maybe add more tests about setting the wrong pool tokens etc, seems like there are a lot of revert conditions that are not covered by these tests

contracts/test/utils/StdLib.sol
```diff
@@ -0,0 +1,306 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "./Vm.sol";
+
+// Wrappers around Cheatcodes to avoid footguns
+abstract contract StdLib {
```
eswak, line 7: where is this lib coming from ?

package.json
```diff
@@ -12,13 +12,14 @@
     "devnet": "ENABLE_MAINNET_FORKING=true NO_RESET=true npx hardhat node",
     "console:rinkeby": "npx hardhat console --network rinkeby",
     "console:ropsten": "npx hardhat console --network ropsten",
-    "clean": "rm -rf build && rm -rf out",
-    "test": "forge test",
+    "clean": "forge clean",
```
eswak, line 15: does that also clean the hardhat builds files?

package.json
```diff
@@ -12,13 +12,14 @@
...
+    "test": "forge test --no-match-contract IntegrationTest",
+    "test:integration": "forge test --fork-url https://eth-mainnet.alchemyapi.io/v2/$MAINNET_ALCHEMY_API_KEY --fork-block-number 14207980 --match-contract UniswapV3OracleIntegrationTest",
```
eswak, line 17: maybe don't hardcode the block number in the package.json, it's faster to do so but it's going to be a pain to debug later if we have an issue because a contract is deployed after this block & we try to use it

also --match-contract is too specific

proposals/dao/fip_85.ts
```diff
@@ -0,0 +1,101 @@
...
+  expect(daiBackupOracle).to.equal(daiUsdcTwapOracle.address);
+
+  const ethPSM = contracts.ethPSM;
+  const ethBackupOracle = await ethPSM.backupOracle();
+  expect(ethBackupOracle).to.equal(ethUsdcTwapOracle.address);
+
+  // 2. Validate that a valid `read()` can be fetched from the oracles
+  const [daiPrice, daiPriceValid] = await daiUsdcTwapOracle.read();
+  expect(daiPriceValid).to.equal(true);
+  expect(daiPrice.value).to.be.bignumber.greaterThan(ethers.BigNumber.from(0));
```
eswak, line 94: check some range like > 0.5 < 1.5, and same for ETH, > 500 < 10000, at least to test sanity & the order of magnitude and that the decimals are not screwed somewhere

eswak: if I remember correctly they are Decimal so it's not just an uint256

contracts/mock/IMockUniswapV3Pool.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import {IUniswapV3Pool} from "@uniswap/v3-core/contracts/interfaces/IUniswapV3Pool.sol";
+
+interface IMockUniswapV3Pool is IUniswapV3Pool {
+  
+  // NOTE: Not part of the UnisV3Pool interface. Test method only
+  function setTokens(address _token0, address _token1) external;
```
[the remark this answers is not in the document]
thomas-waite: nice idea, added

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
thomas-waite: Very good spot. Fixed and will add tests to cover this

contracts/test/unit/oracle/UniswapV3OracleWrapper.t.sol
thomas-waite: The `Uniswap` contracts are compiled with a different version of Solidity to what we use. This test file is compiled with Solidity > 0.8, whereas the Uniswap contracts use < 0.8. 

That means you can't import the Uniswap related contracts into this test Solidity file and compile. Forge has a nice way around this where you can pull the pre-compiled artifact and deploy it with `deployCode()`

contracts/test/unit/oracle/UniswapV3OracleWrapper.t.sol
thomas-waite: Sure, will definitely add one for the wrong pool tokens

contracts/test/utils/StdLib.sol
thomas-waite: Was in a `forge` example repo. Contains lots of useful wrappers such as `deployCode()`, which makes deploying contracts compiled with a different version of straightforward

package.json
thomas-waite: No good point, have added a `rm -rf artifacts`

package.json
thomas-waite: Yeah good point, I shouldn't have hard coded - will move to the `.env.example`. 

In terms of `--match-contract` I need a way to distinguish the integration tests from the unit tests. It matches based on regex, so best way is likely to use `--match-contract` but make it more general than what I have now. Something like:

`--match-contract IntegrateTest`

contracts/mock/MockUniV3Pool.sol
```diff
@@ -0,0 +1,59 @@
...
+    uint16 observationCardinalityNext;
+    // the current protocol fee as a percentage of the swap fee taken on withdrawal
+    // represented as an integer denominator (1/x)%
+    uint8 feeProtocol;
+    // whether the pool is locked
+    bool unlocked;
+  }
+
+  Slot0 public slot0 = Slot0({
+     sqrtPriceX96: uint160(5),
```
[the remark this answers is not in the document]
thomas-waite: Thanks, I use prettier to autoformat and I think it's reading from the single `.prettierrc` we have in the repo (which is setup for `js`). Will sort

proposals/dao/fip_85.ts
thomas-waite: I did this in the corresponding dedicated `backupOracle` `e2e` tests, but agreed - good idea to add here too

thomas-waite: And yep, `read()` returns a `Decimal` as the price. You get access to the raw value through the `.value` property

contracts/mock/MockUniV3Pool.sol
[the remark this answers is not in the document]
thomas-waite: Added an auto-formatter and linter in this PR: https://github.com/fei-protocol/fei-protocol-core/pull/554 . Once that's merged, will then apply to this code

contracts/refs/OracleRef.sol
thomas-waite: Why do you want to be able to invert and normalise at this level? That's an oracle specific implementation detail - how those factors are taken into account differs for each oracle. 

For example, I believe with the Chainlink oracle it's a simple inversion and multiplication once the price is retrieved. However, for the Uniswap TWAP oracle it requires mixing these factors into the intermediate numerator and denominator that you calculate

thomas-waite: Refactoring `OracleRef.sol` is a significant effort and it's unlikely we will call `setDoInvert()` and `setDecimalsNormalizer()`, as you say governor only. But it feels very dangerous to set a `backupOracle` which will report a wildly incorrect price if those methods are ever accidentally called.

eswak (dismissed):

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
```diff
@@ -0,0 +1,251 @@
...
+    // ----------- Governor only state changing api -----------
+
+    /// @notice Change the time period over which the TWAP price is calculated
+    /// @param _twapPeriod Period of time over which time weighted average price is calculated
+    function setTwapPeriod(uint32 _twapPeriod)
+        external
+        override
+        hasAnyOfThreeRoles(
+            TribeRoles.GOVERNOR,
+            TribeRoles.GUARDIAN,
```
ElliotFriedman, line 241: Is this a power we want to give to a guardian?

contracts/oracle/uniswap/UniswapWrapper.sol
```diff
@@ -0,0 +1,147 @@
...
+        (
+            uint256 decimalNormaliser,
+            bool invertDecNormaliser
+        ) = calculateDecimalNormaliser(inputTokenDecimals, outputTokenDecimals);
+
+        uint256 numerator;
+        uint256 denominator;
+        if (!invertPrice) {
+            // price = ((sqrtPrice*2^96) * (sqrtPrice*2^96) * decimalNormaliser) / (2^96)^2
+            numerator = !invertDecNormaliser
```
ElliotFriedman, line 56: If statement would be more legible here.

contracts/oracle/uniswap/UniswapWrapper.sol
```diff
@@ -0,0 +1,147 @@
...
+            denominator = 2**192;
+        } else {
+            // Invert the calculation in the case where Uniswap stores the tokens in the inverse
+            // ratio to that which the oracle requires the price.
+            // i.e. uniswap may quote the price as outputToken/inputToken, whereas we may want inputToken/outputToken
+
+            //price = ((2^96)^2 * decimalNormaliser) / (sqrtPrice*2^96) * (sqrtPrice*2^96)
+            numerator = !invertDecNormaliser
+                ? FullMath.mulDiv(2**192, decimalNormaliser, uint256(1))
+                : FullMath.mulDiv(2**192, uint256(1), decimalNormaliser);
```
ElliotFriedman, line 76: If statement would be more legible here.

contracts/test/integration/backupOracles.sol
```diff
@@ -0,0 +1,293 @@
...
+    /// outputTokenDecimals (usdc): 6
+    /// ----------  Path  --------------
+    /// Invert price ratio: True, Invert decimals: False
+    function testPriceForEthUsdc() public {
+        (Decimal.D256 memory price, bool valid) = ethOracle.read();
+        console.log("eth price:", price.value / precision);
+        assertTrue(valid);
+
+        assertGt(price.value, 2000e18);
+        assertLt(price.value, 5000e18);
```
ElliotFriedman, line 229: This test is pretty fragile, I would recommend either widening the range to 1k-10k, or just testing that spot price isn't more than 10 percent away from the TWAP price.

contracts/test/integration/backupOracles.sol
```diff
@@ -0,0 +1,293 @@
...
+    /// outputTokenDecimals (avinoc): 18
+    /// ----------  Path  --------------
+    /// Invert price ratio: False, Invert decimals: True
+    function testPriceForUsdtAvinoc() public {
+        (Decimal.D256 memory price, bool valid) = usdtAvinocOracle.read();
+        console.log("usdt/avinoc price:", price.value);
+
+        assertTrue(valid);
+        assertGt(price.value, 2e18 + 5e17);
+        assertLt(price.value, 3e18 + 5e17);
```
ElliotFriedman, line 276: Same comment about test being fragile.

contracts/test/integration/backupOracles.sol
```diff
@@ -0,0 +1,293 @@
...
+    /// outputTokenDecimals (usdc): 6
+    /// ----------  Path  --------------
+    /// Invert price ratio: False, Invert decimals: False
+    function testPriceForWbtcUsdc() public {
+        (Decimal.D256 memory price, bool valid) = wbtcOracle.read();
+        console.log("wbtc price:", price.value / precision);
+        assertTrue(valid);
+
+        assertGt(price.value, 40000e18);
+        assertLt(price.value, 50000e18);
```
ElliotFriedman, line 291: This test is fragile and needs to be reworked.

protocol-configuration/backupOracleConfig.ts
```diff
@@ -0,0 +1,26 @@
...
+export const daiUsdcBackupOracleConfig: BackupOracleConfig = {
+  twapPeriod: 600, // 10 minutes
+  minPoolLiquidity: 50e6, // 50m
+  uniswapPool: '0x5777d92f208679DB4b9778590Fa3CAB3aC9e2168',
+  precision: BNe18(1)
+};
+
+export const ethUsdcBackupOracleConfig: BackupOracleConfig = {
+  twapPeriod: 600, // 10 minutes
+  minPoolLiquidity: 50e6, // 50m
```
ElliotFriedman, line 16: Please make sure to run these params by Storm and or Delia for all backup oracle parameters.

ElliotFriedman (changes requested):

contracts/oracle/uniswap/UniswapV3OracleWrapper.sol
```diff
@@ -0,0 +1,251 @@
...
+        uint32 oldTwapPeriod,
+        uint32 newTwapPeriod
+    );
+
+    /// @notice An event emitted when a Uniswap pool has oracle support added
+    event AddPoolSupport(
+        address indexed pool,
+        uint32 twapPeriod,
+        uint16 cardinality
+    );
```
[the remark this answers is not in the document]
thomas-waite: done

protocol-configuration/backupOracleConfig.ts
thomas-waite: I drafted these values with Storm's input + they're what he's been running the data gathering scripts on. But will confirm before deploy

PR #545, opened 2022-02-17
Create PSM Without Reserves Threshold

ElliotFriedman:
Peg Stability Module that holds no funds. On a mint, it transfers all proceeds to a PCV Deposit and then deposits them into the deposit's target When funds are needed for a redemption, they are simply pulled from the PCV Deposit.

TODO:

- Integration tests
- Integrate with the MultiRateLimitedMinter to prevent all funds in the PCV Deposit from being drained in one go.


contracts/peg/PegStabilityModuleNonCustodial.sol
```diff
@@ -0,0 +1,90 @@
...
+
+import {IPCVDeposit} from "./../pcv/IPCVDeposit.sol";
+import {FixedPricePSM} from "./FixedPricePSM.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+/// @notice Peg Stability Module that holds no funds.
+/// On a mint, it transfers all proceeds to a PCV Deposit and then deposits them into the deposit's target
+/// When funds are needed for a redemption, they are simply pulled from the PCV Deposit
+contract PegStabilityModuleNonCustodial is FixedPricePSM {
```
ditchfieldcaleb, line 12: Why is this is a fixed price psm?

contracts/peg/PegStabilityModuleNonCustodial.sol
```diff
@@ -0,0 +1,90 @@
...
+        _floor,
+        _ceiling,
+        params,
+        _mintFeeBasisPoints,
+        _redeemFeeBasisPoints,
+        0, /// hardcode reserve threshold to 0
+        _feiLimitPerSecond,
+        _mintingBufferCap,
+        _underlyingToken,
+        _pcvDeposit
```
ditchfieldcaleb, line 45: Are we going to have one psm per pcv deposit? This doesn't make much sense to me.

ditchfieldcaleb (commented):
(see thoughts in discord)

contracts/peg/IPegStabilityModuleNonCustodial.sol
```diff
@@ -0,0 +1,100 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "../pcv/IPCVDeposit.sol";
+
+/**
+ * @title Fei Peg Stability Module
```
thomas-waite, line 8: Early days of developing this, but the comments here will need adjusting to the `IPegStabilityModuleNonCustodial`

contracts/peg/PegStabilityModuleNonCustodial.sol
thomas-waite: Hmm, I don't agree with this layering of abstractions. 

It feels like a `NonCustodial` PSM is a much more general abstraction than a `FixedPricePSM`. The `NonCustodial` can be used, and if we add more PSMs likely will be, across asset type; whereas the `FixedPricePSM` is an asset specific abstraction. 

So, it makes more sense imo for `NonCustodial` to be the 'base' and have the `FixedPricePSM` inherit from it. 

I'm not a fan of how nested our PSM contracts are becoming and how long the inheritance chain is becoming, but there's not much we can do about that here

contracts/peg/PegStabilityModuleNonCustodial.sol
thomas-waite: I don't believe this design requires that we have one PSM per PCV deposit. 

Each `NonCustodial` PSM will deposit it's funds into a designated initial PCV Deposit. However, we can later freely withdraw those funds and transfer them to a different PCVDeposit using the `ratioPCVController`.

contracts/test/psm/PegStabilityModuleNonCustodial.t.sol
```diff
@@ -0,0 +1,235 @@
...
+import {ICore} from "../../core/ICore.sol";
+import {Core} from "../../core/Core.sol";
+import {IFei, Fei} from "../../fei/Fei.sol";
+import {PegStabilityModuleNonCustodial} from "./../../peg/PegStabilityModuleNonCustodial.sol";
+import {PegStabilityModule} from "./../../peg/PegStabilityModule.sol";
+import {Vm} from "./../utils/Vm.sol";
+import {DSTest} from "./../utils/DSTest.sol";
+import {getCore, getAddresses, FeiTestAddresses} from "./../utils/Fixtures.sol";
+
+contract PegStabilityModuleNonCustodialTest is DSTest {
```
thomas-waite, line 18: Loving the use of `Forge`!

contracts/test/psm/PegStabilityModuleNonCustodial.t.sol
```diff
@@ -0,0 +1,235 @@
...
+
+    modifier prank(address toPrank) {
+      vm.startPrank(toPrank);
+
+      _;
+
+      vm.stopPrank();
+    }
+
+    uint256 immutable public mintAmount = 10_000_000e18;
```
thomas-waite, line 32: Nitpick, but these `immutable` variables can become `constant` variables. You won't be changing the variables of these variables in the constructor. 

Not important though, given these are just tests...

contracts/test/psm/PegStabilityModuleNonCustodial.t.sol
```diff
@@ -0,0 +1,235 @@
...
+
+      PegStabilityModule.OracleParams memory oracleParams = PegStabilityModule.OracleParams({
+          coreAddress: address(core),
+          oracleAddress: address(oracle),
+          backupOracle: address(0),
+          decimalsNormalizer: 0,
+          doInvert: false
+      });
+
+      psm = new PegStabilityModuleNonCustodial(
```
thomas-waite, line 69: There's quite a bit of setup going on here. I'd maybe add some comments to make it easier to follow or break some of it out into a fixture or other function. For example, the granting of permissions or minting could be some form of `setupPSM()` fixture perhaps

contracts/test/psm/PegStabilityModuleNonCustodial.t.sol
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    /// @notice PSM is set up correctly, all state variables and balances are correct 
+    function testPSMSetup() public {
+      uint256 startingPSMUnderlyingBalance = psm.balance();
+      uint256 startingUserFEIBalance = fei.balanceOf(userAddress);
+      uint256 reserveThreshold = psm.reservesThreshold();
+
+      assertEq(reserveThreshold, 0);
+      assertEq(startingPSMUnderlyingBalance, 0);
```
thomas-waite, line 103: Nice! Really like the idea of adding a natspace comment to make the test easier to understand

contracts/test/psm/PegStabilityModuleNonCustodial.t.sol
```diff
@@ -0,0 +1,235 @@
...
+          abi.encodePacked("PegStabilityModule: price out of bounds")
+      );
+
+      psm.redeem(userAddress, mintAmount, mintAmount);
+    }
+
+    /// @notice redeem fails without approval
+    function testSwapFeiForUnderlyingFailsWithoutApproval() public prank(userAddress) {
+      vm.expectRevert(
+          abi.encodePacked("ERC20: transfer amount exceeds allowance")
```
thomas-waite, line 183: You don't need to use `abi.encodePacked` in the `vm.expectRevert`. 

Can just do `bytes("ERC20: transfer amount exceeds allowance")`

contracts/test/psm/PegStabilityModuleNonCustodial.t.sol
```diff
@@ -0,0 +1,235 @@
...
+      );
+
+      psm.redeem(userAddress, mintAmount, mintAmount);
+    }
+
+    /// @notice mint fails without approval
+    function testSwapUnderlyingForFeiFailsWithoutApproval() public prank(userAddress) {
+      vm.expectRevert(
+          abi.encodePacked("ERC20: transfer amount exceeds allowance")
```
thomas-waite, line 192: Surely you're expecting a different error message to this if a mint fails without approval?

contracts/test/psm/PegStabilityModuleNonCustodial.t.sol
```diff
@@ -0,0 +1,235 @@
...
+import {DSTest} from "./../utils/DSTest.sol";
+import {getCore, getAddresses, FeiTestAddresses} from "./../utils/Fixtures.sol";
+
+contract PegStabilityModuleNonCustodialTest is DSTest {
+
+    PegStabilityModuleNonCustodial private psm;
+    ICore private core;
+    IFei private fei;
+
+    modifier prank(address toPrank) {
```
thomas-waite, line 24: Wondering if we can make the tests cleaner than having to add this modifier to each test. Why not just set the `userAddress` to be this contract (`address(this)`) and then mint Fei to this contract. Then we can remove the modifier from each unit test

thomas-waite (commented):
One comment on the layering of abstractions between `FixedPricePSM` and `NonCustodialPSM` which I feel makes more sense to have the other way round. 

Then a few comments on unit test style

contracts/test/psm/PegStabilityModuleNonCustodial.t.sol
ElliotFriedman: if you don't approve a token, then the error should be about not having enough allowance.

contracts/test/psm/PegStabilityModuleNonCustodial.t.sol
ElliotFriedman: added comments documenting this setup.

contracts/test/psm/PegStabilityModuleNonCustodial.t.sol
ElliotFriedman: good idea. accepted!

contracts/peg/PegStabilityModuleNonCustodial.sol
ElliotFriedman: Just inherit from the regular PSM

ElliotFriedman: Done

PR #552, opened 2022-02-25
Houskeeping: proposals

eswak:
Just archive past proposals (that already executed on-chain) in old folders


thomas-waite (approved):
Thanks for keeping things organised!

PR #564, opened 2022-03-01
Improve checkProposal script

eswak:
- Add warning if setup() function is not empty
- Add optional env variable READ_CR_ORACLE that reads CR oracle before & after, and prints the difference

Example of warning:
image

Example of CR oracle reading:
image


thomas-waite (dismissed):
Really like this + the colours are great

xklob (dismissed):
See comment about moving the func to a different file. Also not a huge fan of directly embedding color/terminal format strings in console logs, but should be fine as long as its not something we start doing everywhere.

PR #568, opened 2022-03-02
Optimistic governance pods 

thomas-waite:
# Summary
Adds a new type of governance unit: optimistic pods. They are an Orca pod together with an optimistic timelock. 

This PR only adds the relevant Solidity, a seperate follow up PR will be submitted to add the relevant FIP scripts to deploy, setup and add e2e tests. That will be fairly involved as it involves transitioning over to our new system.

## Description
An optimistic governance pod consists of an Orca Pod together with an optimistic timelock. Deploying and linking these together requires a fair number of function calls, so to aid that this PR adds a `PodManager.sol` as a factory contract to do that. It also includes some helper getter functions for convenience.

Our pods will be arranged into two layers based on permission level:
- `Tribal Council` (admin) 
- `protocol` tier

Runtime contract interaction flow: `Gnosis safe` -> `OptimisticTimelock` -> `Fei protocol contracts`
where the `OptimisticTimelock` has the various permissions and roles granted.

## Expected deployment process
1. Have the relevant podAdmin (DAO or TribalCouncil) deploy a `PodManager`
2. `podAdmin` deploys the required number of pods using the `PodManager`
3. A governor grants the timelock of the OptimisticPod the relevant roles over the parts of the Pod relevant to the protocol, using `grantRole()`


contracts/pods/PodManager.sol
```diff
@@ -0,0 +1,136 @@
...
+        address _core,
+        address proposer,
+        address executor,
+        uint256 minDelay
+    ) internal returns (address) {
+        address[] memory proposers = new address[](1);
+        proposers[0] = proposer;
+
+        address[] memory executors = new address[](1);
+        executors[0] = executor;
```
Joeysantoro, line 125: execution can be basically public or abstracted away. I think it could be cool to create a single "executor" contract which has the access granted upon timelock creation and exposing execution publicly.

Maybe even worth considering a public proposer and vetoer contract. Essentially porting the access control to other fei contracts and logic.

contracts/pods/PodManager.sol
```diff
@@ -0,0 +1,136 @@
...
+        string memory _ensString,
+        string memory _imageUrl,
+        uint256 minDelay
+    ) public onlyPodAdmin returns (uint256, address) {
+        uint256 podId = memberToken.getNextAvailablePodId();
+
+        podController.createPod(
+            _members,
+            _threshold,
+            podAdmin,
```
Joeysantoro, line 89: should the podAdmin be a contract rather than an EOA? Then we don't need to choose between the DAO and Tribal Council and it can allow more of a heirarchy

contracts/pods/PodManager.sol
```diff
@@ -0,0 +1,136 @@
...
+    /// @param _imageUrl Metadata, URL to a image to represent the pod in frontends
+    /// @param minDelay Delay on the timelock
+    function createChildOptimisticPod(
+        address[] memory _members,
+        uint256 _threshold,
+        bytes32 _podLabel,
+        string memory _ensString,
+        string memory _imageUrl,
+        uint256 minDelay
+    ) public onlyPodAdmin returns (uint256, address) {
```
ElliotFriedman, line 83: Make this function external and then have the memory items be calldata

contracts/pods/PodManager.sol
ElliotFriedman: I think the podAdmin will likely be a timelock per the comment on it.

contracts/pods/PodManager.sol
```diff
@@ -0,0 +1,136 @@
...
+
+/// @notice Contract used by an Admin pod to manage child pods.
+
+/// @dev This contract is primarily a factory contract which an admin
+/// can use to deploy more optimistic governance pods. It will create an
+/// Orca pod and deploy an optimistic timelock alongside it.
+///
+/// The timelock and Orca pod are then linked up so that the Orca pod is
+/// the only proposer and executor.
+contract PodManager {
```
ElliotFriedman, line 18: This contract is not named correctly. It should be something like "PodFactory", not PodManager as it does not actually manage the pods after they are created.

package.json
```diff
@@ -12,10 +12,12 @@
     "devnet": "ENABLE_MAINNET_FORKING=true NO_RESET=true npx hardhat node",
     "console:rinkeby": "npx hardhat console --network rinkeby",
     "console:ropsten": "npx hardhat console --network ropsten",
-    "clean": "rm -rf build && rm -rf out",
-    "test": "forge test",
+    "clean": "forge clean && rm -rf artifacts",
+    "test": "forge test --no-match-contract IntegrationTest",
+    "test:integration": "forge test --fork-url https://eth-mainnet.alchemyapi.io/v2/$MAINNET_ALCHEMY_API_KEY --fork-block-number `cat ./contracts/test/integration/_block.txt` --match-contract IntegrationTest",
```
ElliotFriedman, line 17: we should really change this and just leverage an env var instead of catting a file.

ditchfieldcaleb: Agreed, we absolutely shouldn't be using operating-specific commands here either

ditchfieldcaleb (changes requested):

package.json
thomas-waite: Pushed changes and latest feedback regarding integration tests to the PR: https://github.com/fei-protocol/fei-protocol-core/pull/565 . Forked off from that as I need it here

contracts/pods/PodManager.sol
thomas-waite: Good points, overlooked. Thanks

contracts/pods/PodManager.sol
thomas-waite: Yeah that's an artifact of broader plans I originally had for this contract - I was planning for it to be a "gateway" into the pods and managing activities like adding signers, removing signers, granting roles etc. Decided against it in the end + missed renaming. Done now

contracts/pods/PodManager.sol
thomas-waite: Currently it's setup so that each admin pod (i.e. Tribe DAO and Tribal Council) will deploy it's own `PodFactory.sol`, and then use that to create it's child pods. So the `podAdmin` is a contract (the timelock of the DAO or Tribal Council).

I thought that was the simplest/quickest way to get started + the `podAdmin` is what has the powers to add, remove signers etc. 

Will have a think about some kind of orchestrator contract that could create the hierarchy, rather than pushing that to the deploy scripts to create

thomas-waite: As discussed, leaving as is for this MVP

PR #570, opened 2022-03-03
Non Custodial Rate Limited PSM

ElliotFriedman:
This PR builds on work done for the Global Rate Limited Minter by modifying the MultiRateLimited contract to be abstract so that inheriting it in the PSM doesn't cause a double initialization of the CoreRef contract.

This PR modifies the base PSM implementation to make redemptions be rate limited in addition to minting.

Additionally, the PSM and all of its variants had to be refactored to support the additional constructor arguments as there were numerous "Stack Too Deep" errors.


ElliotFriedman:
remove reserve threshold

ElliotFriedman:
remove allocate surplus and add a sweep function

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,460 @@
...
+import {ReentrancyGuard} from "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
+/// @notice Peg Stability Module that holds no funds.
+/// On a mint, it transfers all proceeds to a PCV Deposit and then deposits them into the deposit's target
+/// When funds are needed for a redemption, they are simply pulled from the PCV Deposit
+contract NonCustodialPSM is
+    INonCustodialPSM,
+    RateLimited,
+    OracleRef,
+    PCVDeposit,
```
Joeysantoro, line 25: imo it shouldn't be a PCVDeposit. Assume it holds no assets

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,460 @@
...
+    {
+        updateOracle();
+
+        amountFeiOut = _getMintAmountOut(amountIn);
+        require(
+            amountFeiOut >= minAmountOut,
+            "PegStabilityModule: Mint not enough out"
+        );
+
+        _transferFrom(msg.sender, address(this), amountIn);
```
Joeysantoro, line 298: transfer to the PCVDeposit? Maybe don't call deposit to save gas

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,460 @@
...
+
+        uint256 amountFeiToTransfer = Math.min(
+            fei().balanceOf(address(this)),
+            amountFeiOut
+        );
+        uint256 amountFeiToMint = amountFeiOut - amountFeiToTransfer;
+
+        IERC20(fei()).safeTransfer(to, amountFeiToTransfer);
```
Joeysantoro, line 306: Check if this is 0

contracts/peg/NonCustodialPSM.sol
ElliotFriedman: that would be a pretty massive gas saver if we didn't call deposit

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,460 @@
...
+        address from,
+        address,
+        uint256 amount
+    ) internal {
+        underlyingToken.safeTransferFrom(from, address(pcvDeposit), amount);
+        pcvDeposit.deposit();
+    }
+
+    /// @notice function to move non FEI ERC20 tokens to the PCV Deposit
+    function sweep() external override {
```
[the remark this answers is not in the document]
ElliotFriedman: I can add withdrawERC20 in, but that should be separate functionality from sweeping.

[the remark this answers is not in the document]
ElliotFriedman: Added withdrawERC20

contracts/peg/NonCustodialPSM.sol
ElliotFriedman: done

contracts/peg/NonCustodialPSM.sol
ElliotFriedman: done

contracts/peg/NonCustodialPSM.sol
ElliotFriedman: done

contracts/peg/PegStabilityModule.sol
```diff
@@ -1,13 +1,18 @@
...
-import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
-import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import {Decimal} from "../external/Decimal.sol";
+import {OracleRef} from "./../refs/OracleRef.sol";
+import {CoreRef} from "./../refs/CoreRef.sol";
+import {Constants} from "../Constants.sol";
+import {RateLimitedMinter} from "./../fei/minter/RateLimitedMinter.sol";
+import {IPegStabilityModule} from "./IPegStabilityModule.sol";
+import {IPCVDeposit, PCVDeposit} from "./../pcv/PCVDeposit.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
```
ditchfieldcaleb, line 11: Looks like there are a lot of extra imports here that aren't used

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,456 @@
...
+
+    /// @notice the PCV deposit target to deposit and withdraw from
+    IPCVDeposit public override pcvDeposit;
+
+    /// @notice the token this PSM will exchange for FEI
+    /// This token will be set to WETH9 if the bonding curve accepts eth
+    IERC20 public immutable override underlyingToken;
+
+    /// @notice Rate Limited Minter contract that will be called when FEI needs to be minted
+    GlobalRateLimitedMinter public immutable rateLimitedMinter;
```
ditchfieldcaleb, line 44: I don't see why this shouldn't be changeable by governance - it'll save us several contract deploys if/when we upgrade the minter contract.

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,456 @@
...
+        uint112 feiRateLimitPerSecond;
+        uint144 feiBufferCap;
+        uint112 underlyingTokenRateLimitPerSecond;
+        uint144 underlyingTokenBufferCap;
+    }
+
+    /// @notice constructor
+    /// @param params PSM constructor parameter struct
+    constructor(
+        OracleParams memory params,
```
ditchfieldcaleb, line 101: Using calldata here should save a lot of gas

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,456 @@
...
+            newRedeemFeeBasisPoints <= MAX_FEE,
+            "PegStabilityModule: Redeem fee exceeds max fee"
+        );
+        uint256 _oldRedeemFee = redeemFeeBasisPoints;
+        redeemFeeBasisPoints = newRedeemFeeBasisPoints;
+
+        emit RedeemFeeUpdate(_oldRedeemFee, newRedeemFeeBasisPoints);
+    }
+
+    /// @notice helper function to set the surplus target
```
ditchfieldcaleb, line 230: "surplus target" doesn't make sense here anymore right? this is just the pcv deposit

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,456 @@
...
+    function _setPCVDeposit(IPCVDeposit newSurplusTarget) internal {
```
ditchfieldcaleb, line 231: We should add a check here that ensures the underlying token of the new pcv deposit matches the immutable token in this contract.

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,456 @@
...
+        amountTokenOut = _getRedeemAmountOut(amountFeiIn);
+    }
+
+    /// @notice the maximum mint amount out
+    function getMaxMintAmountOut() external view override returns (uint256) {
+        return fei().balanceOf(address(this)) + buffer();
+    }
+
+    /// @notice function from PCVDeposit that usually must be overriden, but because
+    /// this contract is not a PCV deposit is not overriden
```
ditchfieldcaleb, line 361: Why do we need this method?

contracts/peg/NonCustodialPSM.sol
ditchfieldcaleb: Is this done? Still looks like it's transferring to the psm?

ditchfieldcaleb: Also, if we do this, we need to have some kind of threshold for calling deposit() on the pcv deposit automatically, else it would be relatively cheap for an actor to mint/redeem cycle and pull all of our assets out of the underlying yield-earning strategy in the pcv deposit

ditchfieldcaleb (changes requested):

contracts/utils/RateLimitedReplenishable.sol
```diff
@@ -0,0 +1,35 @@
...
+            _maxRateLimitPerSecond,
+            _rateLimitPerSecond,
+            _bufferCap,
+            _doPartialAction
+        )
+    {}
+
+    /// @notice function to replenish buffer
+    /// @param amount to increase buffer by if under buffer cap
+    function _replenishBuffer(uint256 amount) internal {
```
ditchfieldcaleb, line 25: Can you override redeem and add a call to this inside of it as well?

ditchfieldcaleb (changes requested):

contracts/peg/PegStabilityModule.sol
ElliotFriedman: all of those imports are needed

contracts/peg/NonCustodialPSM.sol
ElliotFriedman: the minter contract should be immutable as changing it everywhere it is used is a massive pain

contracts/peg/NonCustodialPSM.sol
ElliotFriedman: makes testing easier

contracts/utils/RateLimitedReplenishable.sol
ElliotFriedman: why would I override redeem to add this functionality in? I'll just add this logic straight into the redeem function itself.

contracts/peg/NonCustodialPSM.sol
ElliotFriedman: can't make these constructor params calldata, compiler yells when you try

contracts/peg/PegStabilityModule.sol
ElliotFriedman: try removing any of them them and see what happens

contracts/peg/NonCustodialPSM.sol
ElliotFriedman: We don't need to add that check, because we can trust governance to act non-maliciously

contracts/peg/NonCustodialPSM.sol
ElliotFriedman: removed

contracts/peg/NonCustodialPSM.sol
ditchfieldcaleb: That doesn't make any sense; redeploying every contract where the minter is used, because the minter immutable, is much more of a pain than calling a function on every contract where the minter is used to change it. 

Also, please don't resolve comments like this until we come to some kind of consensus.

contracts/peg/NonCustodialPSM.sol
ditchfieldcaleb: @ElliotFriedman why did you resolve this comment?

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,452 @@
...
+        pcvDeposit.withdraw(to, amount);
+    }
+
+    /// @notice transfer assets from user to the PCV Deposit
+    /// Does not call deposit to save gas and make mints cheaper
+    /// @param from sending address
+    /// @param amount number of tokens sent to PCV Deposit
+    function _transferFrom(
+        address from,
+        address,
```
ditchfieldcaleb, line 417: Instead of hardcoding in the pcvDeposit here and ignoring the second parameter (???), it'd be much better to use the pcv deposit address in the function that *calls* this - ie, inside of mint()

contracts/utils/RateLimitedReplenishable.sol
ditchfieldcaleb: Right, that's actually what I meant, thank you!

ditchfieldcaleb (changes requested):
See comments

contracts/peg/NonCustodialPSM.sol
ditchfieldcaleb: It's not about maliciousness, but about preventing error. It's an easy check, I think we should add it.

Joeysantoro: Agree with caleb, assume we are dumb and add sanity checks. Been saved by at least a few of them

contracts/peg/NonCustodialPSM.sol
Joeysantoro: absolutely. This is weird as-is

contracts/peg/NonCustodialPSM.sol
ElliotFriedman: ok, now non immutable and changeable by governance

contracts/peg/NonCustodialPSM.sol
ElliotFriedman: fixed

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,479 @@
...
+    function deposit() external {
+        uint256 currentBalance = underlyingToken.balanceOf(address(this));
+        require(
+            currentBalance != 0,
+            "PegStabilityModule: No balance to deposit"
+        );
+
+        underlyingToken.safeTransfer(address(pcvDeposit), currentBalance);
+        pcvDeposit.deposit();
+    }
```
Joeysantoro, line 469: This function is also useless now imo

Joeysantoro (changes requested):

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,479 @@
...
+        returns (uint256 amountOut)
+    {
+        updateOracle();
+
+        amountOut = _getRedeemAmountOut(amountFeiIn);
+        require(
+            amountOut >= minAmountOut,
+            "PegStabilityModule: Redeem not enough out"
+        );
+        _depleteBuffer(amountFeiIn);
```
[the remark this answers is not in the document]
ditchfieldcaleb: I feel like we should rate limit in the asset that has the oracle, since this will prevent oracle manipulation attacks. Not a trivial question to answer though!

contracts/peg/NonCustodialPSM.sol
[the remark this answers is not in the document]
ElliotFriedman: will delete

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,479 @@
...
+
+        if (amountFeiToTransfer != 0) {
+            IERC20(fei()).safeTransfer(to, amountFeiToTransfer);
+        }
+
+        if (amountFeiToMint != 0) {
+            rateLimitedMinter.mintFei(to, amountFeiToMint);
+        }
+
+        _replenishBuffer(amountFeiToTransfer + amountFeiToMint);
```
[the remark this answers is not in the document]
ElliotFriedman: Yes, this was added to fix the cycling issue. Will move the _replenishBuffer function to the base RateLimited class and remove the intermediary contract.

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,479 @@
...
+        onlyGovernorOrAdmin
+    {
+        _setGlobalRateLimitedMinter(newMinter);
+    }
+
+    /// @notice withdraw ERC20 from the contract
+    /// @param token address of the ERC20 to send
+    /// @param to address destination of the ERC20
+    /// @param amount quantity of ERC20 to send
+    function withdrawERC20(
```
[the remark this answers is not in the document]
ElliotFriedman: fixed

contracts/peg/NonCustodialPSM.sol
ElliotFriedman: removed

contracts/utils/RateLimited.sol
```diff
@@ -85,6 +85,11 @@ abstract contract RateLimited is CoreRef {
             Math.min(_bufferStored + (rateLimitPerSecond * elapsed), bufferCap);
     }
 
+    /// @notice return buffer stored to validate state transitions
```
ditchfieldcaleb, line 88: Why not just make the _bufferStored var public?

ElliotFriedman: that's a really good idea

contracts/peg/NonCustodialPSM.sol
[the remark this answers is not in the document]
ElliotFriedman: the functions are identical

contracts/peg/NonCustodialPSM.sol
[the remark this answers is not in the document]
ElliotFriedman: use `amountFeiOut` now

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,479 @@
...
+    /// Does not call deposit to save gas and make mints cheaper
+    /// @param from sending address
+    /// @param amount number of tokens sent to PCV Deposit
+    function _transferFrom(
+        address from,
+        address pcvDepositAddress,
+        uint256 amount
+    ) internal {
+        underlyingToken.safeTransferFrom(from, pcvDepositAddress, amount);
+    }
```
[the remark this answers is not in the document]
ElliotFriedman: both have been inlined

contracts/peg/NonCustodialPSM.sol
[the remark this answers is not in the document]
ElliotFriedman: Good catch, will remove the buffer depletion before the oracle update so that if things fail, they waste less gas.

contracts/peg/NonCustodialPSM.sol
```diff
@@ -190,6 +190,16 @@ contract NonCustodialPSM is
         _setPCVDeposit(newTarget);
     }
 
+    /// @notice set the target to call for FEI minting
+    /// @param newMinter new Global Rate Limited Minter for this PSM
+    function setGlobalRateLimitedMinter(GlobalRateLimitedMinter newMinter)
+        external
+        override
+        onlyGovernorOrAdmin
```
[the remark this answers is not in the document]
ElliotFriedman: Done

contracts/peg/NonCustodialPSM.sol
```diff
@@ -0,0 +1,476 @@
...
+    bool public mintPaused;
+
+    /// @notice event that is emitted when redemptions are paused
+    event RedemptionsPaused(address account);
+
+    /// @notice event that is emitted when redemptions are unpaused
+    event RedemptionsUnpaused(address account);
+
+    /// @notice event that is emitted when minting is paused
+    event MintingPaused(address account);
```
ditchfieldcaleb, line 64: Event declarations should be in interface

contracts/utils/RateLimited.sol
```diff
@@ -103,15 +103,28 @@ abstract contract RateLimited is CoreRef {
...
+    /// @param amount to increase buffer by if under buffer cap
+    function _replenishBuffer(uint256 amount) internal {
+        uint256 newBuffer = buffer();
+
+        /// cannot replenish any further if already at buffer cap
+        if (newBuffer == bufferCap) {
+            return;
+        }
+
+        bufferStored = newBuffer + amount;
```
ditchfieldcaleb, line 125: Shouldn't this be:

```
bufferStored = Math.min(newBuffer + amount, bufferCap)
```

ditchfieldcaleb (changes requested):

contracts/utils/RateLimited.sol
ElliotFriedman: Good catch.

PR #578, opened 2022-03-07
Fix Tokemak e2e (new rollover address)

eswak:
Just update the rollover address in Tokemak e2e


thomas-waite (approved):
Thanks for fixing!

PR #582, opened 2022-03-09
Add Tribal Council and Optimistic Governance pods

thomas-waite:
# Summary
Introduces `OptimisticGovernancePods` to the protocol. Specifically adds the `TribalCouncil`, a candidate base pod, factories to deploy both and make the process more scaleable/manageable and an FIP script together with e2e tests. 

There is further functionality that will be added in seperate PRs:
- **TribalCouncil council role based authority**: TribalCouncil should be able to delegate certain roles to protocol tier pods. It should also be able to revoke these and cancel transactions on child pod timelocks. 
- **TribalCouncil membership**: Members for the Tribal Council will be voted on through snapshot and then added here. 9 Ethereum addresses will be added
- **Details about the nature of the first pod**: A future PR will add exact details and roles to the first protocol pod. It will also update ENS names
- (possible) Support controller upgrades from Orca

## Timelock details
In general, the more executive and powerful a contract or role, then the longer the associated timelock should be. As pods get more and more specific and isolated timelock requirements should decrease.

Specifically:
- TribalCouncil timelock: 4 days
- Protocol/product tier timelock: 2 days


contracts/pods/PodExecutor.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.0;
+
+import {ITimelock} from "../dao/timelock/ITimelock.sol";
+
+/// @title PodExecutor
+/// @notice Executor gateway contract that allows any address to execute prepared timelock transactions
+/// @dev Access is granted to this contract to execute transactions on a timelock
+contract PodExecutor {
```
Joeysantoro, line 9: add executeBatch?

contracts/pods/PodFactory.sol
```diff
@@ -0,0 +1,225 @@
...
+    /// @param _podAdmin The admin of the pod - able to add and remove pod members
+    function createChildOptimisticPod(
+        address[] calldata _members,
+        uint256 _threshold,
+        bytes32 _podLabel,
+        string calldata _ensString,
+        string calldata _imageUrl,
+        uint256 _minDelay,
+        address _podAdmin
+    ) external override onlyOwner returns (uint256, address) {
```
Joeysantoro, line 149: who is the owner of this contract? imo this should use TribeRoles right?

contracts/pods/PodFactory.sol
```diff
@@ -0,0 +1,225 @@
...
+    /// @notice Create an Orca pod - a Gnosis Safe with a membership wrapper
+    function createPod(
+        address[] calldata _members,
+        uint256 _threshold,
+        bytes32 _podLabel,
+        string calldata _ensString,
+        string calldata _imageUrl,
+        address _podAdmin,
+        uint256 podId
+    ) internal returns (address) {
```
Joeysantoro, line 186: should any of these be standardized? For example we own tribedao and feiprotocol.eth and could use those in an automated way for ens?

Joeysantoro: Probably fine but just an idea

contracts/pods/PodFactory.sol
```diff
@@ -0,0 +1,225 @@
...
+        );
+        return podController.podIdToSafe(podId);
+    }
+
+    /// @notice Create an optimistic timelock, linking to an Orca pod
+    /// @param safeAddress Address of the Gnosis Safe
+    /// @param minDelay Delay on the timelock
+    /// @param publicExecutor Non-permissioned smart contract that
+    ///        allows any address to execute a ready transaction
+    function createOptimisticTimelock(
```
Joeysantoro, line 204: what about vetoing? The "cancel" functionality should be exposed by some kind of veto controller and veto tribe role imo.

Joeysantoro: Imagining a contract like the PublicExecutor called the VetoController which stores a mapping of which addresses can veto which pods or which roles can veto which pods and exposes the "cancel" functionality through them. I think this is required for the mvp

Joeysantoro: Probably not required for this MVP

contracts/test/integration/optimisticPods/OptimisticPodTest.t.sol
```diff
@@ -0,0 +1,139 @@
...
+            "",
+            bytes32("testing"),
+            bytes32("random"),
+            0
+        );
+    }
+
+    /// @notice Validate that a member can be removed from a pod by the admin
+    /// Note: Orca sets the memberTokenID to be the podId
+    function testRemovePodMember() public {
```
Joeysantoro, line 95: how would the DAO remove a pod member?

proposals/dao/fip_82.ts
```diff
@@ -0,0 +1,217 @@
...
+    addresses.feiDAOTimelock // podAdmin
+  );
+  const tribalCouncilPodId = await tribalCouncilPodFactory.latestPodId();
+  const councilTimelockAddress = await tribalCouncilPodFactory.getPodTimelock(tribalCouncilPodId);
+
+  logging && console.log('TribalCouncil pod Id: ', tribalCouncilPodId.toString());
+  logging && console.log('Tribal council timelock deployed to: ', councilTimelockAddress);
+
+  // 4. Deploy protocolTierPodFactory. Set podAdmin to deploy address, to pods can be created
+  const protocolTierPodFactory = await podFactoryEthersFactory.deploy(
```
Joeysantoro, line 79: am I understanding this right? Each pod requires a new factory? That feels wrong

proposals/description/fip_82.ts
```diff
@@ -0,0 +1,144 @@
...
+        '0x899bd46557473cb80307a9dabc297131ced39608330a2d29b2d52b660c03923e' // GOVERN_ROLE
+      ],
+      description: 'Create ORACLE_ADMIN role'
+    },
+    {
+      target: 'core',
+      values: '0',
+      method: 'grantRole(bytes32,address)',
+      arguments: ['0x2172861495e7b85edac73e3cd5fbb42dd675baadf627720e687bcfdaca025096', '{tribalCouncilTimelock}'],
+      description: 'Grant Tribal Council ROLE_ADMIN'
```
Joeysantoro, line 32: what does ROLE_ADMIN do

contracts/pods/PodExecutor.sol
thomas-waite: Yeah good point!

contracts/pods/PodFactory.sol
thomas-waite: Yep! Want to add standardised ENS names in a follow up PR that adjusts the deployment names. 

Maybe all pods are subdomains of the `tribedao` ENS name. `ops.tribedao.eth`, `fuse.tribedao.eth`, `feitribedao.eth` etc

contracts/test/integration/optimisticPods/OptimisticPodTest.t.sol
thomas-waite: PodAdmins can remove pod members. 

The DAO can remove members from the `TribalCouncil` by calling `memberToken.burn(memerToRemove, podId)`. 

The `TribalCouncil` in return is the `podAdmin` of child pods. It in turn can remove any member from a childPod by calling the burn function. 

There isn't a direct way at the moment for the DAO to remove members from child pods - it's devolved in this implementation to the `TribalCouncil`. Is that something we want to add?

proposals/description/fip_82.ts
thomas-waite: I've polluted this PR slightly with the roles PR I'm submitting. `ROLE_ADMIN` is the role for the `TribalCouncil`. 

I want to be able to allow this role to grant a subset of other lesser roles so that the TribalCouncil can authorise child pods

contracts/pods/PodFactory.sol
thomas-waite: I was looking for a way to be able to call this function, with null members, via the deploy address in the first step of deploying the contract and not require a DAO action to do it (makes the DAO script much simpler). 

Have transitioned to TribeRoles, but have a concept of a `deployer` that is able to call the createPod fn. Can discuss

contracts/pods/PodFactory.sol
```diff
@@ -0,0 +1,242 @@
...
+/// can use to deploy more optimistic governance pods. It will create an
+/// Orca pod and deploy an optimistic timelock alongside it.
+///
+/// The timelock and Orca pod are then linked up so that the Orca pod is
+/// the only proposer and executor.
+contract PodFactory is CoreRef, IPodFactory {
+    /// @notice Tribe roles governing deployment of pods
+    bytes32 internal constant GOVERN_ROLE = keccak256("GOVERN_ROLE");
+    bytes32 internal constant POD_DEPLOYER_ROLE =
+        keccak256("POD_DEPLOYER_ROLE");
```
ElliotFriedman, line 26: These two roles should be imported from TribeRoles instead of being redefined.

contracts/pods/PodFactory.sol
thomas-waite: Update from Orca call: Orca doesn't currently support subdomains. Need to pass individual ENS name for each pod

contracts/pods/PodFactory.sol
thomas-waite: Love it, was breaking this out into a seperate PR to ease reviewing load

contracts/pods/PodFactory.sol
thomas-waite: Yeah great point

proposals/dao/fip_82.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  // 4. Create protocol tier pod
+  await podFactory.createChildOptimisticPod(
+    protocolPodConfig.placeHolderMembers,
+    protocolPodConfig.threshold,
+    protocolPodConfig.podLabel,
+    protocolPodConfig.ensString,
+    protocolPodConfig.imageUrl,
+    protocolPodConfig.minDelay,
+    addresses.feiDAOTimelock // Temporarily set to FeiDAOTimelock for deployment. Later transferred to TribalCouncil pod timelock
```
ElliotFriedman, line 86: Why can't we spin up a new timelock just for this initial pod instead of using the feiDAOTimelock?

protocol-configuration/optimisticGovernance.ts
```diff
@@ -0,0 +1,72 @@
...
+  '0x000000000000000000000000000000000000000c'
+];
+
+export const tribeCouncilPodConfig: PodConfig = {
+  members: tribalCouncilMembers,
+  threshold: 5,
+  podLabel: '0x54726962616c436f756e63696c00000000000000000000000000000000000000', // TribalCouncil
+  ensString: 'tribalCouncil.eth',
+  imageUrl: 'tribalCouncil.com',
+  minDelay: 345600,
```
ElliotFriedman, line 42: a 4 day delay for the tribal council timelock sounds good

protocol-configuration/optimisticGovernance.ts
```diff
@@ -0,0 +1,72 @@
...
+  '0x0000000000000000000000000000000000000008'
+];
+
+export const protocolPodConfig: PodConfig = {
+  members: protocolPodMembers,
+  threshold: 3,
+  podLabel: '0x50726f746f636f6c506f64000000000000000000000000000000000000000000', // ProtocolPod
+  ensString: 'protocolPod.eth',
+  imageUrl: 'protocolPod.com',
+  minDelay: 0,
```
ElliotFriedman, line 69: Is this delay of 0 intentional? Seems like the timelock should have some period before becoming executable unless the protocolpod doesn't have any real powers in the system to move funds or grant important roles.

contracts/test/integration/optimisticPods/OptimisticPodTest.t.sol
ditchfieldcaleb: The DAO absolutely should have the ability to remove members of child pods imo

proposals/dao/fip_82.ts
```diff
@@ -0,0 +1,191 @@
...
+// 3. DAO deploys a protocol tier pod factory to deploy protocol pods
+// 4. DAO uses protocol tier pod factory to deploy the first protocol pod
+
+// Validation
+// 1. Validate admins
+// 2. Validate all pod members, proposers and executors
+// 3. Validate correct roles set on contracts
+
+// TODO: Remove once have SHIP tokens on Mainnet
+const mintOrcaToken = async (address: string) => {
```
ditchfieldcaleb, line 26: Haven't seen anything else referencing this Orca token, is this intentional? If so can you put a blurb in the description about it? Feels weird for us to be minting Orca tokens (?)

proposals/dao/fip_82.ts
```diff
@@ -0,0 +1,191 @@
...
+    addresses.feiDAOTimelock // Temporarily set to FeiDAOTimelock for deployment. Later transferred to TribalCouncil pod timelock
+  );
+  const protocolPodId = await podFactory.latestPodId();
+  logging && console.log('Protocol tier pod Id: ', protocolPodId.toString());
+
+  const protocolPodTimelockAddress = await podFactory.getPodTimelock(protocolPodId);
+  logging && console.log('Protocol pod timelock deployed to: ', protocolPodTimelockAddress);
+
+  // 7. Create contract artifacts for timelocks, so address is available to DAO script
+  const timelockABI = [
```
ditchfieldcaleb, line 95: Please import this instead of inlining

test/helpers.ts
```diff
@@ -300,6 +300,16 @@ async function performDAOAction(
   );
 }
 
+const validateArraysEqual = async (arrayA: string[], arrayB: string[]) => {
```
ditchfieldcaleb, line 303: One-liner:

`a.every((a) => expect(b.map((b) => b.toLowerCase()).includes(a.toLowerCase()))
`

Don't think we should add this function in.

ditchfieldcaleb (changes requested):
Overall looks great! Are we importing a bunch of stuff from Orca by copying-pasting? If so, can we use git submodules instead?

contracts/pods/PodFactory.sol
thomas-waite: Update: adding a burner function for initial deploy

proposals/dao/fip_82.ts
thomas-waite: This argument is the admin of the pod being created - which has capabilities such as adding and removing members from the pod. It should be whichever contract is 'above' the contract in the hierarchy. 

In this case, I set it to the `feiDAOTimelock` so that the DAO proposal can add members to the pod on deploy. I then transition the admin at the end of the script to the true `podAdmin`, the `TribalCouncil` timelock

protocol-configuration/optimisticGovernance.ts
thomas-waite: Nope - was going to fine tune the final pod parameters such as ENS names etc. in a follow up PR, but should do this here. Have set to 2 days

proposals/dao/fip_82.ts
thomas-waite: Sure. It's a gating mechanism Orca uses for a slow beta rollout. You can only deploy pods successfully if you have some Orca tokens

proposals/dao/fip_82.ts
thomas-waite: done

ditchfieldcaleb (approved):

contracts/test/integration/optimisticPods/OptimisticPodTest.t.sol
thomas-waite: This is implemented here: https://github.com/fei-protocol/fei-protocol-core/pull/587

PR #586, opened 2022-03-14
Add MetadataRegistry contract

thomas-waite:
# Summary
Adds a `MetadataRegistry` contract with a single non-permissioned state changing function: `registerProposal(uint256 podId,uint256 proposalId,string memory metadata`. 

Used in order to emit `metadata` for pod proposals, which the frontend can listen to and display. This is required as Gnosis Safe's do not support adding arbitrary metadata. A spec is available if useful.

It is intended that the first function call in a pod transaction - before the transaction goes to the appropriate timelock - is sent to this contract. The contract can emit the metadata and allow the frontend to show the proposal as it is in the timelock, where it can the be vetoed if appropriate.

## TODO
Once permissions PR is in, make the `registerProposal()` method permissioned to pods only. Will prevent an unlikely DDOS weakness


contracts/pods/MetadataRegistry.sol
```diff
@@ -0,0 +1,62 @@
...
+        );
+        return registration[proposalHash];
+    }
+
+    /// @notice Register a pod proposal. Specifically used as a layer ontop of a Gnosis safe to emit
+    ///         proposal metadata
+    /// @param podId Unique identifier of the pod for which metadata is being registered
+    /// @param proposalId Unique identifier of the proposal for which metadata is being registered
+    /// @param metadata Proposal metadata
+    // TODO: Restrict to only being callable by pods, to limit DDOS potential. Add new role
```
ElliotFriedman, line 43: This PR should extend or leverage TribeRoles for this.

contracts/pods/MetadataRegistry.sol
```diff
@@ -0,0 +1,62 @@
...
+    function registerProposal(
+        uint256 podId,
+        uint256 proposalId,
+        string memory metadata
+    ) external {
```
ElliotFriedman, line 48: Agree that this function needs to be permissioned, otherwise we are going to get spammed with garbage data.

Joeysantoro: I'd take it a step further and possibly enforce that this is done atomically with pod creation. What are the pros and cons of this?

contracts/pods/MetadataRegistry.sol
thomas-waite: Yep! Once I've got the roles + veto PR in for the TribalCouncil, will be making use of the TribeRoles to permission this

ditchfieldcaleb:
@thomas-waite can you change the naming here to be more specific? Ie it would be really easy to confuse this, governance-proposal specific contract with the other, generalized metadata push that we're thinking about.

Something like "GovernanceMetadataRegistry" would suffice imo

thomas-waite:
> @thomas-waite can you change the naming here to be more specific? Ie it would be really easy to confuse this, governance-proposal specific contract with the other, generalized metadata push that we're thinking about.
> 
> Something like "GovernanceMetadataRegistry" would suffice imo

Totally, have changed

contracts/pods/MetadataRegistry.sol
thomas-waite: Not sure I follow? My plan is to grant a new role `POD_ROLE`, who's admin is the `ROLE_ADMIN`, when the pod is deployed. 

I will then permission this `registerProposal()` function to that `POD_ROLE`

PR #590, opened 2022-03-17
1) Add VetoController and Role hierarchy

thomas-waite:
# Summary
This PR introduces two elements:
1. VetoController
2. Role hierarchy between the TribalCouncil and Pods

The VetoController is being introduced principally to allow the TribalCouncil to veto lower ranking pods.

### Role Hierarchy
Two new roles have been introduced:
- `ROLE_ADMIN`, who's purpose is to manage all roles within `TribeRoles` from `Admin` level down
- `POD_VETO_ADMIN`, who's purpose is to manage the granting and revoking of pod timelock permissions to other TribeRoles

The hierarchy of roles is:
- `GOVERNOR`
   - `ROLE_ADMIN`
        - `Admin Roles`
              - `Minor Roles`     

### VetoController
The `VetoController` allows TribeRoles to be given the ability to veto and cancel proposals that are in the timelock of specific pods. These veto powers can be granted by the `GOVERNOR`, `GUARDIAN` and a newly created role `POD_VETO_ADMIN`.

This contract itself is given the `Proposer` role on the pod timelocks, which in turn allows it to `cancel` and so veto proposals.

## Other
The PR also does some small refactoring, to absorb additional parameters onto the config of the relevant pod.

### DIFF
--- a/contracts/pods/VetoController.sol
+++ b/contracts/pods/VetoController.sol
@@ -10,8 +10,7 @@ import {TribeRoles} from "../core/TribeRoles.sol";
 /// @title Pod Veto Controller
 /// @notice Exposes veto functionality over the timelock of pods to a configurable set of TribeRoles
 /// @dev This contract is intended to be set as a proposer on the timelock of a pod. This gives it the ability
-///      to cancel a proposal. The contract introduced to allow TribeRoles to be added and revoked from
-///      vetoing pods
+///      to cancel a proposal. TribeRoles are added which have the ability to veto proposals.
 contract VetoController is CoreRef {
     using EnumerableSet for EnumerableSet.Bytes32Set;
 
@@ -53,13 +52,13 @@ contract VetoController is CoreRef {
     }
 
     /// @notice Grant veto permission over a particular pod to a TribeRole
-    /// @dev Permissioned to the GOVERNOR, ROLE_ADMIN and GUARDIAN. Used to allow particular TribeRoles
+    /// @dev Permissioned to the GOVERNOR, POD_VETO_ADMIN and GUARDIAN. Used to allow particular TribeRoles
     ///      to be granted veto ability over pods
     function grantVetoPermission(uint256 _podId, bytes32 role)
         public
         hasAnyOfThreeRoles(
             TribeRoles.GOVERNOR,
-            TribeRoles.ROLE_ADMIN,
+            TribeRoles.POD_VETO_ADMIN,
             TribeRoles.GUARDIAN
         )
     {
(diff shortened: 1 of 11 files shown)


contracts/pods/VetoController.sol
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /// @notice Batch revoke veto permission
+    function batchRevokeVetoPermission(
+        uint256[] memory _podIds,
+        bytes32[] memory roles
+    ) external {
+        require(_podIds.length == roles.length, "Mismatched input lengths");
+        for (uint256 i = 0; i < roles.length; ) {
+            revokeVetoPermission(_podIds[i], roles[i]);
```
ElliotFriedman, line 107: I would recommend the same pattern here of removing the check that does multiple calls to core to check permissions, put logic in an internal helper function and call that function instead and permission this function explicitly.

contracts/pods/VetoController.sol
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+
+    /// @notice Batch revoke veto permission
+    function batchRevokeVetoPermission(
+        uint256[] memory _podIds,
+        bytes32[] memory roles
+    ) external {
+        require(_podIds.length == roles.length, "Mismatched input lengths");
+        for (uint256 i = 0; i < roles.length; ) {
```
ElliotFriedman, line 106: gas opti: cache the length in a variable and compare against that

contracts/pods/VetoController.sol
```diff
@@ -0,0 +1,143 @@
...
+        emit RevokeVetoPermission(_podId, timelock, role);
+    }
+
+    /// @notice Batch grant veto permission
+    function batchGrantVetoPermission(
+        uint256[] memory _podIds,
+        bytes32[] memory roles
+    ) external {
+        require(_podIds.length == roles.length, "Mismatched input lengths");
+        for (uint256 i = 0; i < roles.length; ) {
```
ElliotFriedman, line 90: gas opti: cache the length in a variable and compare against that

contracts/pods/VetoController.sol
```diff
@@ -0,0 +1,143 @@
...
+            grantVetoPermission(_podIds[i], roles[i]);
```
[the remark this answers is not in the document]
thomas-waite: Great idea. The modifier currently on `grantVetoPermission`: 

```
hasAnyOfThreeRoles(
            TribeRoles.GOVERNOR,
            TribeRoles.POD_VETO_ADMIN,
            TribeRoles.GUARDIAN
        )
```
is making calls to core to validate roles. That's unnecessary here to do many times

PR #610, opened 2022-03-24
RoleBastion for TribalCouncil

thomas-waite:
# Summary
Introduces a `RoleBastion` to allow the TribalCouncil to create new roles. `Core` will be used to grant and revoke roles.

`RoleBastion` is granted the `GOVERNOR` role to allow it to create new roles. It has a very simple API with one external state changing function to minimise the attack surface.


contracts/pods/RoleBastion.sol
```diff
@@ -0,0 +1,83 @@
...
+    // function createRole(bytes32 role)
+    //     external
+    //     onlyTribeRole(TribeRoles.ROLE_ADMIN)
+    // {
+    //     // onlyGovernor can call create role
+    //     core().createRole(role, TribeRoles.ROLE_ADMIN);
+    // }
+
+    /// @notice Grant a role to an address. Only the ROLE_ADMIN can call this
+    function grantRole(bytes32 role, address to)
```
ElliotFriedman, line 62: Don't you need to check here that the `role` isn't PCVController, Governor, etc, or is this implicitly done in Core?

contracts/test/RoleBastion.t.sol
```diff
@@ -0,0 +1,90 @@
...
+import {DSTest} from "./utils/DSTest.sol";
+import {Vm} from "./utils/Vm.sol";
+import {getCore, getAddresses, FeiTestAddresses} from "./utils/Fixtures.sol";
+import {Core} from "../core/Core.sol";
+import {RoleBastion} from "../pods/RoleBastion.sol";
+import {TribeRoles} from "../core/TribeRoles.sol";
+
+// import "hardhat/console.sol";
+
+contract RoleBastionTest is DSTest {
```
ElliotFriedman, line 13: I'd like to see a test here that shows RoleBastion trying to grant privileged roles and failing

contracts/pods/RoleBastion.sol
```diff
@@ -0,0 +1,98 @@
...
+    {
+        // Add explicit guard that major roles cannot be granted.
+        // It is a redundant check as Core access control will also protect
+        //  - ROLE_ADMIN is not the admin of major roles
+        require(
+            role != TribeRoles.GOVERNOR &&
+                role != TribeRoles.GUARDIAN &&
+                role != TribeRoles.PCV_CONTROLLER &&
+                role != TribeRoles.MINTER,
+            "Only non-major roles can be granted"
```
thomas-waite, line 70: getRoleAdmin - to check admin is not 0

contracts/pods/RoleBastion.sol
```diff
@@ -0,0 +1,29 @@
...
+    /// @param role Role to be created
+    /// @dev Function is used by an address with ROLE_ADMIN, e.g. the TribalCouncil, to create
+    ///      non-major roles. A check is made to ensure that the role to be created does not
+    //       have an admin i.e. that it is a new admin
+    function createRole(bytes32 role)
+        external
+        onlyTribeRole(TribeRoles.ROLE_ADMIN)
+    {
+        bytes32 roleAdmin = core().getRoleAdmin(role);
+        require(roleAdmin == bytes32(0), "Role already exists");
```
ElliotFriedman, line 25: We should not allow a role that is bytes32(0) to be created using this `createRole` method as that is the CONTRACT_ADMIN_ROLE in core.

Screen Shot 2022-03-25 at 11 45 33 AM

thomas-waite: It is possible to create and grant a role of `bytes32(0)` - it has not yet been created in core. I added an integration test that validates that it can be created. As I understand it, that would allow the TribalCouncil to create a powerful role, which it could then grant. 

Have added a check to prevent creating the `bytes32(0)` role + a test to validate. I'm not sure I follow what you mean by granting another contract admin of 0x0?

Joeysantoro: `    bytes32 public constant DEFAULT_ADMIN_ROLE = 0x00; ` in AccessControl.sol

The 0 role is the default admin of all non-created roles and is super powerful. Good to make sure this can't be granted out

PR #620, opened 2022-03-27
Add NopeDAO

thomas-waite:
# Summary
Introduces the NopeDAO - a DAO governed by Tribe who's only functionality is to veto pods


contracts/dao/NopeDAO.sol
```diff
@@ -0,0 +1,122 @@
...
+        GovernorVotesComp(_tribe)
+    {}
+
+    function quorum(uint256 blockNumber)
+        public
+        pure
+        override
+        returns (uint256)
+    {
+        return 10000000e18;
```
Joeysantoro, line 32: nit use underscore notation
```suggestion
        return 10_000_000e18;
```

contracts/test/integration/fixtures/MainnetAddresses.sol
```diff
@@ -0,0 +1,21 @@
...
+
+struct MainnetAddresses {
+    address feiDAOTimelock;
+    address core;
+    address tribe;
+    address memberToken;
+    address podController;
+}
+
+function getMainnetAddresses() pure returns (MainnetAddresses memory) {
```
Joeysantoro, line 12: maybe this should be a constants lib?

contracts/dao/NopeDAO.sol
```diff
@@ -0,0 +1,122 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import {Governor, IGovernor} from "@openzeppelin/contracts/governance/Governor.sol";
+import {GovernorSettings} from "@openzeppelin/contracts/governance/extensions/GovernorSettings.sol";
+import {GovernorVotesComp, IERC165} from "@openzeppelin/contracts/governance/extensions/GovernorVotesComp.sol";
+import {GovernorCountingSimple} from "@openzeppelin/contracts/governance/extensions/GovernorCountingSimple.sol";
+import {ERC20VotesComp} from "@openzeppelin/contracts/token/ERC20/extensions/ERC20VotesComp.sol";
+
+contract NopeDAO is
```
Joeysantoro, line 10: Two major comments, I *believe*  the only change required to achieve early execution is overriding state():
```
    function state(uint256 proposalId) public view virtual override returns (ProposalState) {
        ProposalCore storage proposal = _proposals[proposalId];

        if (proposal.executed) {
            return ProposalState.Executed;
        }

        if (proposal.canceled) {
            return ProposalState.Canceled;
        }

        uint256 snapshot = proposalSnapshot(proposalId);

        if (snapshot == 0) {
            revert("Governor: unknown proposal id");
        }

        if (snapshot >= block.number) {
            return ProposalState.Pending;
        }

        uint256 deadline = proposalDeadline(proposalId);

        if (deadline >= block.number) {
            return ProposalState.Active;
        }

        if (_quorumReached(proposalId) && _voteSucceeded(proposalId)) {
            return ProposalState.Succeeded;
        } else {
            return ProposalState.Defeated;
        }
    }
   ``` 
   
   And reordering the Succeeded state to come before Active

Comment (Joeysantoro) [reply contracts/dao/NopeDAO.sol:10]: Also and this is important, I think the NopeDAO should wait for xTRIBE before launching possibly

File: contracts/test/integration/fixtures/MainnetAddresses.sol (hunk 2 above)
Comment (thomas-waite) [reply contracts/test/integration/fixtures/MainnetAddresses.sol:12]: Yeah I like it! Converted

File: contracts/dao/NopeDAO.sol (hunk 3 above)
Comment (thomas-waite) [reply contracts/dao/NopeDAO.sol]: I also think this is the case. `execute()` checks that a proposal has either `Succeeded` or  is `Queued` in order to allow it to execute. 

`_proposals` is private in `Governor`, so the `GovernorQuickReaction()` overrides `state()` by calling super and modifying the state based on if quorum is met etc

File: contracts/dao/nopeDAO/GovernorQuickReaction.sol [hunk 4]
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import {GovernorCountingSimple} from "@openzeppelin/contracts/governance/extensions/GovernorCountingSimple.sol";
+
+/**
+ * @dev Extension of {Governor} for quick reaction, whereby a proposal
+ *      is executed immediately once Quorum is reached.
+ */
+abstract contract GovernorQuickReaction is GovernorCountingSimple {
```
thomas-waite, line 10: Inherits from `GovernorCountingSimple` as need `_quorumReached` etc and some vote counting logic to be implemented. Would prefer to be inheriting from a general standard interface like `GovernorCounting` that is vote method agnostic

Joeysantoro: This should just inherit from Governor which defines _quorumReachedAbstractly.

I have a slight preference for explicitly enumerating the state rather than using this inheritance supercall because I think that is cleaner and more auditable/readable.

contracts/dao/nopeDAO/NopeDAO.sol
```diff
@@ -0,0 +1,172 @@
...
+        override
+        returns (uint256)
+    {
+        return _quorum;
+    }
+
+    ////////////     GOVERNOR ONLY FUNCTIONS     //////////////
+
+    /// @notice Override of a Governor Settings function, to restrict to Tribe GOVERNOR
+    function setVotingDelay(uint256 newVotingDelay)
```
Joeysantoro, line 49: delay can be hardcoded to zero

contracts/dao/nopeDAO/NopeDAO.sol
```diff
@@ -0,0 +1,172 @@
...
+        override(IGovernor, GovernorSettings)
+        returns (uint256)
+    {
+        return super.votingPeriod();
+    }
+
+    function getVotes(address account, uint256 blockNumber)
+        public
+        view
+        override(IGovernor, GovernorVotesComp)
```
ElliotFriedman, line 108: Why override, couldn't you just inherit?

contracts/dao/nopeDAO/NopeDAO.sol
```diff
@@ -0,0 +1,172 @@
...
+        return super.propose(targets, values, calldatas, description);
+    }
+
+    function proposalThreshold()
+        public
+        view
+        override(Governor, GovernorSettings)
+        returns (uint256)
+    {
+        return super.proposalThreshold();
```
ElliotFriedman, line 138: Why override?

contracts/dao/nopeDAO/NopeDAO.sol
```diff
@@ -0,0 +1,172 @@
...
+import {CoreRef} from "../../refs/CoreRef.sol";
+import {TribeRoles} from "../../core/TribeRoles.sol";
+import {GovernorQuickReaction} from "./GovernorQuickReaction.sol";
+
+contract NopeDAO is
+    Governor,
+    GovernorSettings,
+    GovernorVotesComp,
+    GovernorQuickReaction,
+    CoreRef
```
ElliotFriedman, line 17: This architecture is really interesting. I thought the NopeDAO would have been created as an opinionated DAO that could only call the cancel function of another DAO, but this is a way better architecture and keeps backwards compatibility with all the existing DAO infrastructure.

contracts/dao/nopeDAO/NopeDAO.sol
thomas-waite: It's a Solidity requirement. This contract inherits from two base classes that define the same function with the same arguments - `getVotes()` is on the interface of `IGovernor` and implemented in `GovernorVotesComp()`. 

Solidity requires in that case that you use this pattern and override. This error trace is what you get when you just try to inherit instead:

image

contracts/dao/nopeDAO/GovernorQuickReaction.sol
thomas-waite: Yep great point on the inheritance from Governor, and then rely on any end contract implementing/inheriting from something that implements `_quorumReached`. Have updated

PR #644, opened 2022-04-04
Add pod membership lock

thomas-waite:
# Summary
Integrates a new pod controller version which adds a pod membership lock toggle. Pod admins are now able to lock and unlock pod membership transfers

### DIFF
--- a/contracts/pods/PodAdminGateway.sol
+++ b/contracts/pods/PodAdminGateway.sol
@@ -13,8 +13,11 @@ import {IPodFactory} from "./interfaces/IPodFactory.sol";
 
 /// @title PodAdminGateway for TRIBE Governance pods
 /// @notice Acts as a gateway for admin functionality and vetos in the TRIBE governance pod system
-/// @dev Contract is intended to be set as the podAdmin for all deployed Orca pods. It then allows
-///
+/// @dev Contract is intended to be set as the podAdmin for all deployed Orca pods. Specifically enables:
+///     1. Adding a member to a pod
+///     2. Removing a member from a pod
+///     3. Transferring a pod member
+///     4. Toggling a pod membership transfer switch
 contract PodAdminGateway is CoreRef, IPodAdminGateway {
     /// @notice Orca membership token for the pods. Handles permissioning pod members
     MemberToken private immutable memberToken;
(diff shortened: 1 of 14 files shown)


contracts/pods/PodAdminGateway.sol
```diff
@@ -190,6 +209,41 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
         }
     }
 
+    /// @notice Admin functionality to turn off pod membership transfer
+    /// @dev Permissioned to GOVERNOR, POD_ADMIN, GUARDIAN and the specific role
+    function lockMembershipTransfers(uint256 _podId)
```
ElliotFriedman, line 214: Are all membership transfers locked by default?

thomas-waite: Not by default, this switch needs to be toggled on by the admin and at the moment requires it the pod operator creating another pod to disable transfers. Will look at disabling by default

contracts/pods/PodFactory.sol
```diff
@@ -116,7 +119,17 @@ contract PodFactory is CoreRef, IPodFactory {
         override
         returns (address)
     {
-        return IControllerV1(podController).podAdmin(podId);
+        return ControllerV1(podController).podAdmin(podId);
```
Joeysantoro, line 122: nit: no need to cast it is already casted
```suggestion
        return podController.podAdmin(podId);
```

contracts/pods/PodFactory.sol
```diff
@@ -116,7 +119,17 @@ contract PodFactory is CoreRef, IPodFactory {
...
+    }
+
+    /// @notice Get whether membership transfers are enabled for a pod
+    function getIsMembershipTransferLocked(uint256 podId)
+        external
+        view
+        override
+        returns (bool)
+    {
+        return ControllerV1(podController).isTransferLocked(podId);
```
Joeysantoro, line 132:
```suggestion
        return podController.isTransferLocked(podId);
```

contracts/pods/PodAdminGateway.sol
```diff
@@ -65,6 +68,22 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
             );
     }
 
+    /// @notice Calculate the specific pod membership transfer lock role
```
Joeysantoro, line 71: what does this do

contracts/pods/PodAdminGateway.sol
Joeysantoro: agree it should be disabled by default. The factory can have the role and atomically disable

Joeysantoro (commented):
LGTM overall

contracts/pods/PodFactory.sol
thomas-waite: Ah, slipped through. Corrected

contracts/pods/PodAdminGateway.sol
thomas-waite: Yep agreed

PR #674, opened 2022-04-08
Governance TODOs

thomas-waite:
# Summary
Implements governance upgrade todos:

**Main points**
- `admin` of a pod is set on a per pod basis via the `PodConfig` that is passed to `factory.createOptimisticPod()`. This is done for two reasons:
    - Allow the `PodFactory` to be be passed to `PodAdminGateway`, so that in `veto()` we can validate that `podId` corresponds to the passed `timelock` (there won't be a cyclical dependency)
    - Allow us to more easily upgrade the admin contract. An individual pod/safe can migrate itself, and then we will create all new pods in the future by passing a new admin. Rather than requiring to redeploy the factory
- Consolidated specific pod admin roles into: `admin` and `guardian`. Map onto existing global role abilities
- DAO vote will not be bricked by additional Orca pod deploys. We deploy the TribalCouncil first and get a unique podId. The DAO vote script will need updating with our TribalCouncil Id 
- Replaced used of `OptimisticTimelock.sol` with `TimelockController.sol`
- Removed some unnecessary global roles and consolidated
- Have a simplified `deployGenesisPod()` function to deploy the first pod. It is used to deploy the TribalCouncil before the DAO vote FIP. This means we have the podId. It is not placed in the constructor of the factory as it relies on args passed to the constructor which I want to be immutable for future deploys

### DIFF
--- a/contracts/dao/timelock/OptimisticTimelock.sol
+++ b/contracts/dao/timelock/OptimisticTimelock.sol
@@ -26,7 +26,7 @@ contract OptimisticTimelock is TimelockController, CoreRef {
 
         In addition it allows for much more granular and flexible access for multisig operators
     */
-    // Guardian can become admin on every Optimistic timelock and effetively become 
+    // Guardian can become admin on every Optimistic timelock and effetively become
     function becomeAdmin() public onlyGuardianOrGovernor {
         this.grantRole(TIMELOCK_ADMIN_ROLE, msg.sender);
     }
--- a/contracts/dao/timelock/OptimisticTimelock.sol
+++ b/contracts/dao/timelock/OptimisticTimelock.sol
@@ -26,7 +26,6 @@ contract OptimisticTimelock is TimelockController, CoreRef {
 
         In addition it allows for much more granular and flexible access for multisig operators
     */
-    // Guardian can become admin on every Optimistic timelock and effetively become
     function becomeAdmin() public onlyGuardianOrGovernor {
         this.grantRole(TIMELOCK_ADMIN_ROLE, msg.sender);
     }
--- a/contracts/pods/PodAdminGateway.sol
+++ b/contracts/pods/PodAdminGateway.sol
@@ -26,13 +26,18 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
     /// @notice Pod controller for the pods
     ControllerV1 public immutable podController;
 
+    /// @notice Pod factory
+    IPodFactory public immutable podFactory;
+
     constructor(
         address _core,
         address _memberToken,
-        address _podController
+        address _podController,
+        address _podFactory
     ) CoreRef(_core) {
         memberToken = MemberToken(_memberToken);
         podController = ControllerV1(_podController);
+        podFactory = IPodFactory(_podFactory);
     }
 
     ////////////////////////   GETTERS   ////////////////////////////////
@@ -211,6 +216,10 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
             getSpecificPodAdminRole(_podId)
         )
     {
+        require(
+            _podTimelock == podFactory.getPodTimelock(_podId),
+            "PodId and timelock mismatch"
+        );
         emit VetoTimelock(_podId, _podTimelock, _proposalId);
         TimelockController(payable(_podTimelock)).cancel(_proposalId);
     }
--- a/contracts/pods/PodAdminGateway.sol
+++ b/contracts/pods/PodAdminGateway.sol
@@ -50,7 +50,7 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
         override
         returns (bytes32)
     {
-        return keccak256(abi.encode(_podId, "ORCA_POD", "ADMIN"));
+        return keccak256(abi.encode(_podId, "_ORCA_POD", "_ADMIN"));
     }
 
     /// @notice Calculate the specific pod guardian role identifier
@@ -61,7 +61,7 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
         override
         returns (bytes32)
     {
-        return keccak256(abi.encode(_podId, "ORCA_POD", "GUARDIAN"));
+        return keccak256(abi.encode(_podId, "_ORCA_POD", "_GUARDIAN"));
     }
 
     /////////////////////////    ADMIN PRIVILEDGES       ////////////////////////////
--- a/contracts/pods/PodAdminGateway.sol
+++ b/contracts/pods/PodAdminGateway.sol
@@ -64,20 +64,6 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
         return keccak256(abi.encode(_podId, "ORCA_POD", "POD_VETO_ROLE"));
     }
 
-    /// @notice Calculate the specific pod transfer admin role
-    // TODO: Remove this - not needed
-    function getPodTransferAdminRole(uint256 _podId)
-        public
-        pure
-        override
-        returns (bytes32)
-    {
-        return
-            keccak256(
-                abi.encode(_podId, "ORCA_POD", "POD_TRANSFER_ADMIN_ROLE")
-            );
-    }
-
     /// @notice Calculate the specific pod membership transfer lock role
     function getSetMembershipTransferLockRole(uint256 _podId)
         public
@@ -96,53 +82,6 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
     }
 
     /////////////////////////    ADMIN PRIVILEDGES       ////////////////////////////
-
-    /// @notice Transfer the pod admin address for a pod to another address
-    function transferPodAdmin(uint256 _podId, address newPodAdmin)
-        external
-        override
-        hasAnyOfThreeRoles(
-            TribeRoles.GOVERNOR,
-            TribeRoles.POD_ADMIN,
-            getPodTransferAdminRole(_podId)
-        )
-    {
-        _transferPodAdmin(_podId, newPodAdmin);
-    }
-
-    /// @notice Batch transfer the pod admin address for several pods
-    /// @dev Mass transfer of podAdmins only expected to be performed by GOVERNOR or POD_ADMIN
-    function batchTransferPodAdmins(
-        uint256[] calldata _podIds,
-        address[] calldata newPodAdmins
-    )
-        external
-        override
-        hasAnyOfTwoRoles(TribeRoles.GOVERNOR, TribeRoles.POD_ADMIN)
-    {
-        require(
-            _podIds.length == newPodAdmins.length,
-            "MISMATCHED_ARG_LENGTHS"
-        );
-        uint256 numPodsToTransfer = _podIds.length;
-        for (uint256 i = 0; i < numPodsToTransfer; ) {
-            _transferPodAdmin(_podIds[i], newPodAdmins[i]);
-
-            // i is bounded by numPodsToTransfer
-            unchecked {
-                i += 1;
-            }
-        }
-    }
-
-    /// @notice Transfer a pod admin from this gateway to another address
-    function _transferPodAdmin(uint256 _podId, address newPodAdmin) internal {
-        ControllerV1 podController = podFactory.podController();
-
-        address oldPodAdmin = address(this);
-        emit UpdatePodAdmin(_podId, oldPodAdmin, newPodAdmin);
-        podController.updatePodAdmin(_podId, newPodAdmin);
-    }
 
     /// @notice Admin functionality to add a member to a pod
     /// @dev Permissioned to GOVERNOR, POD_ADMIN and POD_ADD_MEMBER_ROLE
--- a/contracts/pods/PodAdminGateway.sol
+++ b/contracts/pods/PodAdminGateway.sol
@@ -200,11 +200,7 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
     /// @notice Allow a proposal to be vetoed in a pod timelock
     /// @dev Permissioned to GOVERNOR, POD_VETO_ADMIN, GUARDIAN, POD_ADMIN and the specific
     ///      pod admin and guardian roles
-    function veto(
-        uint256 _podId,
-        address _podTimelock,
-        bytes32 _proposalId
-    )
+    function veto(uint256 _podId, bytes32 _proposalId)
         external
         override
         hasAnyOfSixRoles(
@@ -216,10 +212,7 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
             getSpecificPodAdminRole(_podId)
         )
     {
-        require(
-            _podTimelock == podFactory.getPodTimelock(_podId),
-            "PodId and timelock mismatch"
-        );
+        address _podTimelock = podFactory.getPodTimelock(_podId);
         emit VetoTimelock(_podId, _podTimelock, _proposalId);
         TimelockController(payable(_podTimelock)).cancel(_proposalId);
     }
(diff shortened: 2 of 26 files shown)


contracts/pods/PodAdminGateway.sol
```diff
@@ -236,58 +179,61 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
     }
 
     /// @notice Admin functionality to turn off pod membership transfer
-    /// @dev Permissioned to GOVERNOR, POD_ADMIN, GUARDIAN and the specific role
+    /// @dev Permissioned to GOVERNOR, POD_ADMIN and the specific pod admin role
     function lockMembershipTransfers(uint256 _podId)
         external
         override
-        hasAnyOfFourRoles(
+        hasAnyOfThreeRoles(
```
thomas-waite, line 186: Locked by default when this factory deploys pods. GUARDIAN should not need unlock/lock powers. GUARDIAN scoped to remove members + veto

contracts/pods/PodAdminGateway.sol
```diff
@@ -236,58 +179,61 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
...
+        uint256 _podId,
+        address _podTimelock,
+        bytes32 _proposalId
+    )
         external
         override
-        hasAnyOfFourRoles(
+        hasAnyOfSixRoles(
             TribeRoles.GOVERNOR,
             TribeRoles.POD_VETO_ADMIN,
```
thomas-waite, line 229: POD_VETO_ADMIN is granted to the NopeDAO, this is how the NopeDAO vetoes

contracts/pods/PodFactory.sol
```diff
@@ -1,64 +1,77 @@
...
+        podExecutor = PodExecutor(_podExecutor);
         podController = ControllerV1(_podController);
         memberToken = MemberToken(_memberToken);
+        podAdminGateway = PodAdminGateway(_podAdminGateway);
     }
 
     ///////////////////// GETTERS ///////////////////////
 
+    /// @notice Get the number of pods this factory has created
+    function getNumberOfPods() external view override returns (uint256) {
```
thomas-waite, line 71: Utility, to easily track how many pods this factory is creating

contracts/pods/PodFactory.sol
```diff
@@ -135,85 +148,49 @@ contract PodFactory is CoreRef, IPodFactory {
...
 
-    //////////////////// STATE-CHANGING API ////////////////////
-
-    /// @notice Create a child Orca pod with optimistic timelock. Callable by the DAO and the Tribal Council
-    ///         Returns podId, optimistic pod timelock address and the Pod Gnosis Safe address
-    ///         This will lock membership transfers by default
-    function createChildOptimisticPod(PodConfig calldata _config)
-        public
+    /// @notice Deploy the genesis pod, one time use method
+    function deployGenesisPod(PodConfig calldata _config)
```
thomas-waite, line 152: Want a simple way to deploy the first pod. Could do this in the constructor but it references constructor params that I want to be immutable, and would prefer to keep them immutable. 

This method is intended to be used to deploy the TribalCouncil, before the DAO vote happens, so we already know the podId and can't get bricked. A seperate call in the DAO vote must be made to lock transfers, as at deploy time the factory doesn't have the necessary role to call into the `PodAdminGateway`.

contracts/pods/PodFactory.sol
```diff
@@ -270,34 +258,36 @@ contract PodFactory is CoreRef, IPodFactory {
...
-        OptimisticTimelock timelock = new OptimisticTimelock(
-            address(core()),
+        TimelockController timelock = new TimelockController(
             minDelay,
             proposers,
             executors
         );
-        emit CreateOptimisticTimelock(address(timelock));
+        // Revoke TIMELOCK_ADMIN_ROLE priviledges from deployer factory
+        timelock.revokeRole(timelock.TIMELOCK_ADMIN_ROLE(), address(this));
```
thomas-waite, line 288: Safety/standard practice, even though this factory has no method to take advantage of this role

contracts/dao/timelock/OptimisticTimelock.sol
```diff
@@ -26,7 +26,6 @@ contract OptimisticTimelock is TimelockController, CoreRef {
 
         In addition it allows for much more granular and flexible access for multisig operators
     */
-    // Guardian can become admin on every Optimistic timelock and effetively become
```
Joeysantoro, line 29: why was this deleted

contracts/pods/PodAdminGateway.sol
```diff
@@ -18,132 +18,71 @@ import {IPodFactory} from "./interfaces/IPodFactory.sol";
...
-    {
-        return
-            keccak256(abi.encode(_podId, "ORCA_POD", "POD_REMOVE_MEMBER_ROLE"));
+    MemberToken public immutable memberToken;
+
+    /// @notice Pod controller for the pods
+    ControllerV1 public immutable podController;
+
+    /// @notice Access control modifier to extend CoreRef ACL to check for 6 roles
+    modifier hasAnyOfSixRoles(
```
Joeysantoro, line 30: should be put in coreref

contracts/pods/PodAdminGateway.sol
```diff
@@ -18,132 +18,71 @@ import {IPodFactory} from "./interfaces/IPodFactory.sol";
...
         public
         pure
         override
         returns (bytes32)
     {
-        return
-            keccak256(
-                abi.encode(_podId, "ORCA_POD", "POD_TRANSFER_ADMIN_ROLE")
-            );
+        return keccak256(abi.encode(_podId, "ORCA_POD", "ADMIN"));
```
Joeysantoro, line 70: nit but might be better to include a delimiter

contracts/pods/PodAdminGateway.sol
```diff
@@ -242,60 +179,61 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
...
     /// @notice Allow a proposal to be vetoed in a pod timelock
-    /// @dev Permissioned to GOVERNOR, POD_ADMIN, GUARDIAN and specific POD_VETO_ROLE
-    function veto(uint256 _podId, bytes32 proposalId)
+    /// @dev Permissioned to GOVERNOR, POD_VETO_ADMIN, GUARDIAN, POD_ADMIN and the specific
+    ///      pod admin and guardian roles
+    function veto(
+        uint256 _podId,
+        address _podTimelock,
+        bytes32 _proposalId
+    )
```
Joeysantoro, line 224: this change introduced a vulnerability. There is no check that the timelock input is the one associated with the pod

contracts/pods/PodFactory.sol
Joeysantoro: ok thats fine but can you just call it council instead of genesis. Don't want it to be confused with the protocol launch

contracts/pods/PodFactory.sol
```diff
@@ -1,64 +1,77 @@
...
-    MemberToken private immutable memberToken;
+    MemberToken public immutable memberToken;
 
     /// @notice Public contract that will be granted to execute all timelocks created
-    address public immutable podExecutor;
+    PodExecutor public immutable podExecutor;
 
-    /// @notice Mapping between podId and it's optimistic timelock
+    /// @notice Pod admin gateway, through which admin functionality on pods is accessed
+    PodAdminGateway public immutable podAdminGateway;
```
Joeysantoro, line 33: We may want this to be upgradeable, would we just upgrade the factory too in that case?

contracts/pods/PodFactory.sol
Joeysantoro: Also there is no access control on this method but I think thats ok

Joeysantoro: membership transfer lock is not enabled by default on the first pod

contracts/pods/PodFactory.sol
```diff
@@ -212,40 +201,43 @@ contract PodFactory is CoreRef, IPodFactory {
...
             );
-            // Set mapping from podId to timelock for reference
-            getPodTimelock[podId] = timelock;
-            getPodId[timelock] = podId;
-            latestPodId = podId;
-            }
-            
+        } else {
+            timelock = address(0);
         }
```
Joeysantoro, line 230: not necessary else condition, empty variables are default initialized to null value 0x0;

contracts/pods/PodFactory.sol
```diff
@@ -212,40 +201,43 @@ contract PodFactory is CoreRef, IPodFactory {
...
-        emit CreatePod(podId, safeAddress);
+
+        // Set mapping from podId to timelock for reference
+        getPodTimelock[podId] = timelock;
+        getPodId[timelock] = podId;
```
Joeysantoro, line 234: these two should be moved inside the if condition. Don't want to set the address 0 podId and also theres no need to set timelock of podId to 0 given its default 0x0

contracts/pods/PodFactory.sol
```diff
@@ -212,40 +201,43 @@ contract PodFactory is CoreRef, IPodFactory {
...
+        numPods += 1;
```
Joeysantoro, line 235: If we're going to spend an sstore to add this, I'd rather push to an append-only array so we have the addresses on-chain too. numPods = array.length

contracts/pods/PodFactory.sol
```diff
@@ -267,43 +259,36 @@ contract PodFactory is CoreRef, IPodFactory {
...
         proposers[1] = podAdmin;
 
         address[] memory executors = new address[](2);
         executors[0] = safeAddress;
         executors[1] = publicExecutor;
 
-        // TODODODODODO: Use TimelockController from OZ
-        OptimisticTimelock timelock = new TimelockController(
-            address(core()),
+        TimelockController timelock = new TimelockController(
```
Joeysantoro, line 283: we should use v4.6 when it releases https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.6/contracts/governance/TimelockController.sol and be more specific with the roles

contracts/pods/RoleBastion.sol
```diff
@@ -30,9 +30,5 @@ contract RoleBastion is CoreRef {
         require(roleAdmin == bytes32(0), "Role already exists");
         emit BastionRoleCreate(role, TribeRoles.ROLE_ADMIN);
         core().createRole(role, TribeRoles.ROLE_ADMIN);
-
-        // TODO: DAO vote to transfer all roles which have admins of GOVERNOR
-        // VOTIUM, admin GOVERNOR
-        //  transfer
```
Joeysantoro, line 36: what are your plans here? Is there a github issue?

Joeysantoro (changes requested):
A lot better overall but still some important unresolved issues

contracts/pods/RoleBastion.sol
thomas-waite: Implemented in this PR: https://github.com/fei-protocol/fei-protocol-core/pull/678

contracts/pods/PodFactory.sol
thomas-waite: Good point, removed. Will add a comment to make it explicit

contracts/pods/PodFactory.sol
thomas-waite: Will rename to council. It will be called as part of the deployment script so haven't included access control. 

Membership transfer lock is not called in the deploy function as the podFactory will not yet have the relevant roles granted to it by the DAO, to be able to call into the `podAdminGateway`.

Flow is:
**Deploy script**
1. Deploy `podFactory`. It will not yet have any roles
2. Call `deployCouncilPod()` - to deploy the initial TribalCouncil. It will be unlocked

**DAO vote**
1. Grant `podFactory` the relevant roles, e.g. `POD_ADMIN`
2. DAO calls the locking function to lock transfers on the TribalCouncil. It is included in `fip_82.ts` and tested for

contracts/pods/PodFactory.sol
thomas-waite: Agreed. Didn't want to pull the contract direct from Github pre-release. Reaching out to ask about v4.6.0 release timelines

contracts/pods/PodFactory.sol
thomas-waite: Done

contracts/pods/PodFactory.sol
thomas-waite: Yep, should have put inside. Done

contracts/pods/PodAdminGateway.sol
thomas-waite: I added here as wasn't certain about how we handle applying changes to `CoreRef` and then deploying those. We can add extra functionality and then deploy new contracts with a new `CoreRef` - it's additive only so no need to be concerned about existing contracts having the new functionality available

contracts/pods/PodFactory.sol
```diff
@@ -69,7 +69,17 @@ contract PodFactory is CoreRef, IPodFactory {
 
     /// @notice Get the number of pods this factory has created
     function getNumberOfPods() external view override returns (uint256) {
-        return numPods;
+        return podSafeAddresses.length;
+    }
+
+    /// @notice Get the safe addresses of all pods created by this factory
+    function getPodSafeAddresses()
```
thomas-waite, line 70: Should never be so many pods created that this runs out of gas fetching

contracts/pods/PodAdminGateway.sol
thomas-waite: Have added the check, the mistake came about when I tried to decouple the factory and this gateway. 

The `PodAdminGateway` is now a dependency on the factory rather than the other way around so it can have this check:

```
require(
            _podTimelock == podFactory.getPodTimelock(_podId),
            "PodId and timelock mismatch"
        );
```

The `admin` is made available to the factory when it seeks to deploy a pod, again via the `PodConfig` that is passed to `factory.createOptimisticPod()`. This also makes it easier to create new pods with an updated admin

contracts/pods/PodFactory.sol
thomas-waite: Have updated. Have removed setting the `podAdminGateway` in the `PodFactory` constructor, and instead the `PodAdminGateway` is now a dependency of the factory - it's necessary for `podAdminGateway.veto()`.

The `admin` is set on a per pod basis through the `PodConfig` passed to `factory.createOptimisticPod()`. To upgrade the admin:

1. New PodAdmin contract is deployed
2. TribalCouncil deploys a new pod by calling `podFactory.createOptimisticPod()`, passing this new podAdmin
3. TribalCouncil authorises new pod
4. TribalCouncil revokes and deprecates the old pod

Only the podAdmin is able to migrate a pod's admin, and even if we were to add that functionality we would need a a contract with the `TIMELOCK_ADMIN_ROLE`(maybe the `podFactory`)? For this first version of governance propose running with above flow and iterating if need be. Admin functionality is expected to be fairly stable.

contracts/pods/PodFactory.sol
```diff
@@ -39,7 +39,7 @@ contract PodFactory is CoreRef, IPodFactory {
     mapping(address => uint256) public getPodId;
 
     /// @notice Number of pods created
-    uint256 private numPods;
+    address[] private podSafeAddresses;
 
     /// @notice Track whether the one time use initial pod deploy has been used
     bool public genesisDeployed;
```
Joeysantoro, line 42: this param name should change too

contracts/pods/PodAdminGateway.sol
```diff
@@ -211,6 +216,10 @@ contract PodAdminGateway is CoreRef, IPodAdminGateway {
             getSpecificPodAdminRole(_podId)
         )
     {
+        require(
+            _podTimelock == podFactory.getPodTimelock(_podId),
```
Joeysantoro, line 220: might as well just remove the _podTimelock input param and use the canonical one

contracts/pods/PodFactory.sol
Joeysantoro: Still looks to be 4.5. Can we import the more recent commits via npm override?

Joeysantoro (approved):

contracts/pods/PodFactory.sol
thomas-waite: Going to update now in a seperate PR, merging this down

thomas-waite: Upgraded here: https://github.com/fei-protocol/fei-protocol-core/pull/739

PR #735, opened 2022-04-26
Multiratelimited and Ratelimited updates

ditchfieldcaleb:
Adapted from @ElliotFriedman's pr, took a pass over it, added IRateLimited interface, and standardized var and function names across both Ratelimited and Multiratelimited.


ElliotFriedman:
RateLimited looks a lot cleaner now that it doesn't have the the doPartialAction boolean.

PR #736, opened 2022-04-26
Updates .sol line width to 120

ditchfieldcaleb:


Joeysantoro (approved):
LGTM, I reran prettier on another branch off develop and got the same diff

thomas-waite (approved):
Like the increase of line width to 120

PR #739, opened 2022-04-26
Upgrade OZ contracts, make use of CANCELLER role

thomas-waite:
# Summary
Upgrades the OZ contracts to the next release. The next release contains a version of the `TimelockController.sol` that has a more granular `CANCELLOR_ROLE` we can make use of. 

We are technically using the 'pre-release' here - the pre-release period ended on the 22nd April and it is now going to main release. Have confirmed with the OZ team that there are no planned differences between this pre-release and the main release - it is being released as is in a few days.

## Code changes
Previously, all `proposers` passed to the `new TimelockController()` created for a pod in `podFactory` were given the `PROPOSER_ROLE` which entitled the proposers to both proposal creation and cancellation abilities. 

This upgrade restricts the proposer to just the pod safeAddress and not the pod admin contract. Both the pod safe and podAdmin have cancellation rights.


contracts/pods/PodFactory.sol
```diff
@@ -268,7 +268,10 @@ contract PodFactory is CoreRef, IPodFactory {
     /// @param publicExecutor Non-permissioned smart contract that
     ///        allows any address to execute a ready transaction
     /// @param podAdmin Address which is the admin of the Orca pods
-    ///        Will also be able to propose
+    /// @dev Roles that individual addresses are granted on the relevant timelock:
+    //         safeAddress - PROPOSER_ROLE, CANCELLER_ROLE, EXECUTOR_ROLE
+    //         podAdmin - CANCELLER_ROLE
+    //         publicExecutor - EXECUTOR_ROLE
```
Joeysantoro, line 274: just want to double check, is the publicExecutor pausable/revokable?

Joeysantoro (approved):

PR #763, opened 2022-04-29
Add proposal DEBUG tool

eswak:


thomas-waite (approved):
Looks great. This is a really useful tool for measuring proposal gas costs and debugging them when they fail locally

PR #783, opened 2022-05-07
Add sanity check tests to Gov upgrade

thomas-waite:
An e2e test to make sure the TribalCouncil timelock works, as well as a batch execute test


eswak (approved):

package.json
```diff
@@ -14,7 +14,7 @@
     "console:ropsten": "npx hardhat console --network ropsten",
     "clean": "forge clean && rm -rf artifacts",
     "test": "forge test --no-match-contract IntegrationTest",
-    "test:integration": "FORK_BLOCK=14516592; forge test --fork-url https://eth-mainnet.alchemyapi.io/v2/$MAINNET_ALCHEMY_API_KEY --fork-block-number $FORK_BLOCK --match-contract IntegrationTest",
+    "test:integration": "FORK_BLOCK=14726759; forge test --fork-url https://eth-mainnet.alchemyapi.io/v2/$MAINNET_ALCHEMY_API_KEY --fork-block-number $FORK_BLOCK --match-contract IntegrationTest",
```
ditchfieldcaleb, line 17: It's weird to me that we're setting the fork block in two places - here and in the hardhat config. Should we use a commited .env? (for example - .forksettings.env or something)

test/integration/tests/podOperation.ts
```diff
@@ -68,6 +68,8 @@ describe('Pod operation and veto', function () {
     const podAdminGateway = contracts.podAdminGateway as PodAdminGateway;
     tribalCouncilTimelockSigner = await getImpersonatedSigner(contractAddresses.tribalCouncilTimelock);
 
+    await forceEth(contractAddresses.tribalCouncilTimelock);
+    await forceEth(contractAddresses.feiDAOTimelock);
```
ditchfieldcaleb, line 72: Why is this duplicated?

thomas-waite: Needed more than 1 eth, and the `forceSpecificEth` didn't seem to be working as expected

thomas-waite: so just did the quick and nasty for this test

package.json
thomas-waite: That was the earlier purpose behind my committed `block.txt` file :)

thomas-waite: As discussed, added the `block.txt` and we'll source it. Bad practice imo to commit `.env` files + I use a global gitignore for them

ditchfieldcaleb (approved):

PR #786, opened 2022-05-09
Add TribalCouncil proposal simulation option

thomas-waite:
# Summary
Adds a TribalCouncil simulation option for developing and testing TC proposals. 

Refactors the existing simulate OA tool slightly to re-use.


eswak (approved):
yay

PR #789, opened 2022-05-10
contract admin cleanup

Joeysantoro:


contracts/pcv/convex/VotiumBriber.sol
```diff
@@ -43,7 +43,7 @@ contract VotiumBriber is CoreRef {
     /// the _proposal ID, if _choiceIndex is out of range, or of block.timestamp
     /// is after the deadline for bribing (usually 6 hours before Convex snapshot
     /// vote ends).
-    function bribe(bytes32 _proposal, uint256 _choiceIndex) public onlyGovernorOrAdmin whenNotPaused {
+    function bribe(bytes32 _proposal, uint256 _choiceIndex) public onlyTribeRole(TribeRoles.VOTIUM_ROLE) whenNotPaused {
```
ditchfieldcaleb, line 46: should this be `oneOfTwoRoles(TribeRoles.GOVERNOR, TribeRoles.VOTIUM_ROLE)` or was removing the gov intentional?

Joeysantoro: removing gov is fine, its only ever been called by the ops timelock

ditchfieldcaleb (approved):

thomas-waite (commented):
Looks good to me. Do we also want to deploy this contract and deprecate the old one? We could make a TC proposal to do it

Joeysantoro:
> Looks good to me. Do we also want to deploy this contract and deprecate the old one? We could make a TC proposal to do it

I don't think this is necessary. It doesn't change anything practically

PR #791, opened 2022-05-10
Withdraw 30M LP tokens from Convex and 10M Fei from D3 Pool

thomas-waite:
# Summary
Withdraws 30M Fei from the Curve D3 pool.

Uses the PCVGuardian to first pull the LP tokens from the Convex PCVDeposit and send them to the Curve PCVDeposit. Then withdraws 10M Fei from the Curve PCVDeposit and sends it to the DAI PSM. Two further transactions will be made in the future to withdraw additional Fei. Makes use of the TribalCouncil to do this.

A Fei Skimmer is also deployed for the DAI PSM to enable Fei to be skimmed off the DAI PSM. A DAO vote is required to do this, as the `PCV_CONTROLLER_ROLE` needs granting.


protocol-configuration/mainnetAddresses.ts
```diff
@@ -1676,6 +1676,11 @@ const MainnetAddresses: MainnetAddresses = {
     address: '0xFf419Bc27483edb94b7Ad5c97b7FaB5DB323c7E0',
     category: AddressCategory.Deprecated
   },
+  daiFixedPriceFeiSkimmer: {
+    artifactName: 'FeiSkimmer',
+    address: '',
```
thomas-waite, line 1681: To be filled in on deployment

proposals/dao/fip_105.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/*
+
+DAO Proposal #105
+
+Description: Deploy a Fei skimmer for the DAI PSM
+*/
+
+const fipNumber = '105';
+const skimThreshold = ethers.constants.WeiPerEther.mul(10_000_000);
```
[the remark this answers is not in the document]
thomas-waite: Updated to 20M - the `bufferCap` for the DAI PSM

ditchfieldcaleb (dismissed):

test/unit/pcv/utils/FeiSkimmer.test.ts
```diff
@@ -81,4 +81,21 @@ describe('FeiSkimmer', function () {
...
+  describe('Set Source', function () {
+    it('from governor succeeds', async function () {
+      expect(await skimmer.source()).to.be.equal(source.address);
+
+      await skimmer.connect(impersonatedSigners[governorAddress]).setSource(userAddress);
+
+      expect(await skimmer.source()).to.be.equal(userAddress);
+    });
+
+    it('not from governor succeeds', async function () {
```
eswak, line 94: nit: "fails"

protocol-configuration/mainnetAddresses.ts
```diff
@@ -1676,6 +1676,12 @@ const MainnetAddresses: MainnetAddresses = {
     address: '0xFf419Bc27483edb94b7Ad5c97b7FaB5DB323c7E0',
     category: AddressCategory.Deprecated
   },
+  // TODO: Uncomment this FeiSkimmer and use once DAO vote has granted roles
```
thomas-waite, line 1679: Commented until a DAO vote gives the FeiSkimmer the relevant roles (currently fails e2e tests)

contracts/pcv/utils/FeiSkimmer.sol
```diff
@@ -53,4 +54,11 @@ contract FeiSkimmer is CoreRef {
         threshold = newThreshold;
         emit ThresholdUpdate(newThreshold);
     }
+
+    /// @notice Set the target to skim from. Only Governor or Admin
+    /// @param newSource the new source to skim from
+    function setSource(address newSource) external onlyGovernorOrAdmin {
```
ElliotFriedman, line 60: if you're going to use the onlyGovernorOrAdmin modifier, you have to set the admin of this contract in the constructor

ElliotFriedman (changes requested):

contracts/pcv/utils/FeiSkimmer.sol
thomas-waite: It's set by default in `CoreRef`, which this contract inherits from, to `GOVERNOR`. It's a bit redundant as the modifier allows GOVERNOR to interact with this method anyway but the contract admin could be changed in the future to something else. 

Should stay as is imo

thomas-waite: It's set by default in `CoreRef`, which this contract inherits from, to `GOVERNOR`. That's a bit redundant at the moment as the GOVERNOR can call this method anyway, but the contract admin could be changed in the future to something else. 

Should stay as is imo

thomas-waite: Have thought about this some more. In the recent governance upgrade, we changed the admin of the `feiSkimmer` contracts to the `PCV_MINOR_PARAM_ROLE`. That role is intended to only manage minor PCV related parameters and should not have the scope to change the source in my opinion.

Going to change `setSource` to be `onlyGovernor` and then `setContractAdmin` internally to `PCV_MINOR_PARAM_ROLE` to avoid having to do later in a DAO vote

ElliotFriedman (approved):

ditchfieldcaleb (dismissed):

ditchfieldcaleb (approved):

PR #796, opened 2022-05-11
Withdraw Fei from major deposits and add underlying compound address

thomas-waite:


protocol-configuration/mainnetAddresses.ts
```diff
@@ -1981,6 +1981,11 @@ const MainnetAddresses: MainnetAddresses = {
     address: '0xB80B3dc4F8B30589477b2bA0e4EF2b8224bDf0a5',
     category: AddressCategory.PCV
   },
+  compoundUnderlyingPCVDepost: {
```
eswak, line 1984: to be consistent, can you name it `compoundFeiPCVDeposit` ? and rename `compoundPCVDepositWrapper` to `compoundFeiPCVDepositWrapper`

thomas-waite: yep

eswak (approved):

PR #97, opened 2021-04-30
Feature : PCVSwapperUniswap

eswak:
## Description
This pull request introduces a new `PCVSwapperUniswap` contract.

It is designed to receive drips from an `EthPCVDripper` or any future `ERC20PCVDripper` and perform series of small swaps to sell an ERC20 token over time and acquire other ERC20 tokens on Uniswap. The received ETH is instantly converted to WETH and the contract can work with any two ERC20 that have a pair on Uniswap to perform conversions of the tokens or ETH held in the PCV.

An example of first deployment could be : 
- A dripper drips 15000 ETH per day to the `PCVSwapperUniswap`
- The `swap()` function can be called every 10 minutes and swaps 100 ETH for ~270k DAI every time

A trade of this size already incurs a ~0.3% slippage in the ETH-DAI pool, I don't think we should do much larger trades. After we have some DAI (or other stablecoins), Curve could be used for larger (>10M$) stablecoin swaps at low slippage.

The interval of ~10min between swaps should allow the market to absorb the price movements created by the PCV.

At this pace, it would take ~2.5 days to convert 100M$ of the PCV to another ERC20.

The swap size & interval is configurable by governor & the contract can be paused by guardian if we observe some market manipulations.

## Steps
- [x] Implement smart contracts
- [x] Mocha unit tests
- [x] Test on a local mainnet fork
- [x] More formal review of the completed code
- [ ] Safety audit
- [ ] Snapshot vote
- [ ] DAO vote
- [ ] Deployed

PR #802, opened 2022-05-13
Slowly liquidate CREAM 

thomas-waite:
# Summary
Slowly liquidate the Cream position by selling it over 1 month into the Sushiswap CREAM-WETH market.

## Sale parameters
The Sushiswap CREAM-WETH market is the most liquid DEX market for CREAM. There is ~$570k liquidity, with 24hr volume that fluctuates daily between usually ~$50k to sometimes up to $700k. More info: https://analytics.sushi.com/pairs/0xf169cea51eb51774cf107c88309717dda20be167

We will sell with the following parameters:
- ~$8k of CREAM every 8hrs (400 CREAM tokens per sale, according to current prices of $20 per CREAM)
- Each $8k sale is expected to incur ~3% slippage. This will get progressively worse
- The protocol has 31,780 CREAM tokens to liquidate, ~$600k
- At this rate of selling (~$24k per day) it will take ~25 days to liquidate our position
- Max slippage allowed per individual trade set to 10%

## Implementation
Makes use of the `PCVSwapperUniswap` contract. Available here: https://github.com/fei-protocol/fei-protocol-core/pull/97 

This was audited but never shipped. The contract has been kept as close to the original as possible, with a few minimal upgrades/compatibility updates:
- Upgrade Solidity version from 0.6.0 to 0.8.0
- Remove the `FeiRouter.sol` contract and associated dependency. It was significant extra surface area and was only used to source the `WETH` address. The `WETH` address is now passed as a parameter at deploy time.
- Make the interface of `tokenSpent`, `tokenReceived` and `tokenReceivingAddress` consistent with the latest version of `IPCVSwapper.sol` (previously this contract used `getTokenReceived()` etc)
- Updated the arguments passed to the UniRef constructor
- Updated various events and moved ones that were previously on `IPCVSwapper` to this contract
- Modernises the access control to use TribeRoles where appropriate and reduces the level required for non-major setters from `onlyGovernor` to `hasTribeRole(TribeRoles.PCV_MINOR_PARAM_ROLE)`. This is held by the DAO timelock, TribalCouncil timelock and optimisticTimelock

Diff between the original contract and the minimally updated/simplified contract: https://www.diffchecker.com/RZyO9dYt

### Oracle
Uses the `ChainlinkCREAMEthOracle`. Price has been checked, it reports the CREAM price in units of Eth. 

### Receiving address
This is the Aave Eth PCV Deposit, which wraps Eth.


ElliotFriedman (changes requested):

contracts/pcv/uniswap/FeiRouter.sol
```diff
@@ -0,0 +1,125 @@
...
+        uint256 amountOutMin,
+        address to,
+        uint256 deadline
+    ) external payable override ensure(deadline) returns (uint256 amountOut) {
+        (uint256 reservesETH, uint256 reservesOther, bool isWETHPairToken0) = _getReserves();
+
+        uint256 amountIn = msg.value;
+        amountOut = UniswapV2Library.getAmountOut(amountIn, reservesETH, reservesOther);
+
+        require(amountOut >= amountOutMin, "FeiRouter: Insufficient output amount");
```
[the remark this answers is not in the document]
thomas-waite: This PR is very much WIP and not ready for review. I open prematurely in general and attach WIP to show what I'm working on. 

This contract was copied over from a historic branch that I'm using. It's being removed and isn't needed

contracts/pcv/uniswap/PCVSwapperUniswap.sol
```diff
@@ -0,0 +1,320 @@
...
+        (uint256 amount0Out, uint256 amount1Out) = pair.token0() == address(tokenSpent)
+            ? (uint256(0), amountOut)
+            : (amountOut, uint256(0));
+        pair.swap(amount0Out, amount1Out, tokenReceivingAddress, new bytes(0));
+
+        // Emit event
+        emit Swap(msg.sender, tokenSpent, tokenReceived, amountIn, amountOut);
+
+        // Incentivize call with FEI rewards
+        _incentivize();
```
thomas-waite, line 253: Note: This won't be used, not granting this contract minting priviledges

contracts/pcv/uniswap/PCVSwapperUniswap.sol
```diff
@@ -0,0 +1,320 @@
...
+        // Prepare swap
+        uint256 amountOut = UniswapV2Library.getAmountOut(amountIn, tokenSpentReserves, tokenReceivedReserves);
+
+        return amountOut;
+    }
+
+    /// @notice see external function getNextAmountReceivedThreshold()
+    function _getMinimumAcceptableAmountOut(uint256 amountIn) internal view returns (uint256) {
+        Decimal.D256 memory twap = readOracle();
+        Decimal.D256 memory oracleAmountOut = twap.mul(amountIn);
```
eswak, line 289: be careful about this, most oracles report an USD value, so the "token out" has to be a USD token. Or if oracle reports in ETH value, "token out" has to be in ETH. Easy to misconfigure things around oracles there, we'll need to be very careful in proposal validation scripts when using swappers

PR #819, opened 2022-05-18
DAO vote for PCV reinforcement

thomas-waite:
# Summary
Consolidates various PCV reinforcement strands into one proposal for execution, and performs a couple of needed technical maintenance tasks. DAO vote is necessary as several of the funds are on the DAO timelock (Cream and DPI) and we need to give out the PCV_CONTROLLER role.

1. Set `uniswapPCVDeposit` and `dpiToDaiLBPSwapper` as safe addresses
2. Deploy Balancer LBP and initialise auction of DPI for DAI
3. Fix NopeDAO voting period
4. Transfer CREAM to TribalCouncil multisig, where it will then be sold to ETH
5. Fund the TribalCouncil multisig with 10 ETH


contracts/test/unit/governance/NopeDAO.t.sol
```diff
@@ -77,7 +77,7 @@ contract NopeDAOTest is DSTest {
         assertEq(votingDelay, 0);
 
         uint256 votingPeriod = nopeDAO.votingPeriod();
-        uint256 fourDays = 86400 * 4;
+        uint256 fourDays = 26585;  // (86400 * 4) / 13
```
thomas-waite, line 80: OZ governor has this parameter in units of blocks rather than seconds. Make a mistake in the original config, currently reports 22 months

proposals/description/fip_105.ts
```diff
@@ -1,17 +1,115 @@
...
+      description: 'Set the voting period for the NopeDAO to 4 days'
+    },
+    ////////  Transfer CREAM to TribalCouncil Multisig  /////////
+    {
+      target: 'cream',
+      values: '0',
+      method: 'transfer(address,uint256)',
+      arguments: ['{tribalCouncilSafe}', '31780370000000000000000'],
+      description: 'Transfer CREAM to TribalCouncil multisig where it will then be swapped'
+    },
```
eswak, line 85: Could you also move the `15,000` that are on the `feiDAOTimelock` to the `aaveEthPCVDeposit` ? they are unreachable by the guardian and it's a good occasion to move them back to a PCVDeposit

protocol-configuration/mainnetAddresses.ts
```diff
@@ -176,6 +176,26 @@ const MainnetAddresses: MainnetAddresses = {
...
+  dpiToDaiLensDpi: {
+    artifactName: 'BPTLens',
+    address: '0xaDdB7eBdCA3fa3b72D2e57c8e660C90ec00af7Cc',
+    category: AddressCategory.PCV
+  },
+  dpiToDaiLBPPool: {
+    artifactName: 'ILiquidityBootstrappingPoolFactory',
+    address: '0xd10386804959a121a8a487e49f45aa9f5a2eb2a0',
+    category: AddressCategory.External
+  },
```
eswak, line 198: don't forget to mark `aaveRaiPCVDepositWrapper`, `aaveRaiPCVDeposit`, etc as `AddressCategory.Deprecated`

eswak: all these + the contracts under the wrappers :

 '{rariPool31FeiPCVDepositWrapper}',
          '{rariPool25FeiPCVDepositWrapper}',
          '{rariPool9RaiPCVDepositWrapper}',
          '{aaveRaiPCVDepositWrapper}',
          '{rariPool19DpiPCVDepositWrapper}',
          '{liquityFusePoolLusdPCVDeposit}',
          '{rariPool72FeiPCVDepositWrapper}',
          '{raiDepositWrapper}',
          '{dpiDepositWrapper}'

proposals/description/fip_105.ts
eswak: and keep WETH lens on the feiDAOTimelock because WETH might end up there some time again in the future. We remove the RAI and DPI ones from CR because these assets are liquidated out of the PCV, but it's not the case for WETH

thomas-waite: Makes sense. Have transferred and then deposited it into the PCV deposit. Kept the lens

protocol-configuration/mainnetAddresses.ts
thomas-waite: Already done in this PR :)

proposals/description/fip_105.ts
```diff
@@ -1,17 +1,133 @@
...
+  title: 'FIP-105: Reinforce PCV by consolidating assets and perform technical maintenance',
   commands: [
+    /////////  DAI PSM Skimmer   //////////////
     {
       target: 'core',
       values: '0',
       method: 'grantRole(bytes32,address)',
       arguments: ['0x0866eae1216ed05a11636a648003f3f62921eb97ccb05acc30636f62958a8bd6', '{daiFixedPricePSMFeiSkimmer}'],
       description: 'Grant the new DAI PSM Skimmer the PCV_CONTROLLER_ROLE'
+    },
```
Joeysantoro, line 13: just flagging that this is unscalable and we should try to revitalize the sink efforts when things settle down

proposals/description/fip_105.ts
```diff
@@ -1,17 +1,133 @@
...
+      arguments: ['{dpiToDaiLBPSwapper}', '37888449801955370645659'],
+      description: 'Transfer DPI from DAO timelock to the LBP swapper'
+    },
+    {
+      target: 'compoundDaiPCVDeposit',
+      values: '0',
+      method: 'withdraw(address,uint256)',
+      arguments: ['{dpiToDaiLBPSwapper}', '187947000000000000000000'],
+      description: 'Withdraw DAI from the CompoundPCVDeposit and transfer to the LBP pool'
+    },
```
Joeysantoro, line 29: what happens if DPI price changes? make sure to overfund DAI, or whitelist the swapper as a safe address

contracts/dao/nopeDAO/NopeDAO.sol
```diff
@@ -28,7 +28,7 @@ contract NopeDAO is
         Governor("NopeDAO")
         GovernorSettings(
             0, /* 0 blocks */
-            4 days,
+            26585, /* 4 days measured in blocks. Assumed 13s block time */
```
kryptoklob, line 31: Using blocks to measure time is an antipattern. I haven't looked at the GovernorSettings, but is there a way we can use block.timestamp instead of # of blocks?

proposals/description/fip_105.ts
thomas-waite: As discussed, have removed this in favour of setting an appropriate PCV deposit to be a safe address, pulling funds there and then burning

contracts/dao/nopeDAO/NopeDAO.sol
thomas-waite: I agree it's an antipattern and is what caught me out originally. The `GovernorSettings` module measures time in units of blocks unfortunately

proposals/description/fip_105.ts
thomas-waite: Good point, have whitelisted as a safe address. Can then fine-tune as necessary

Joeysantoro (approved):

eswak (dismissed):

Joeysantoro (approved):

PR #821, opened 2022-05-18
Prepare LBP withdraw

thomas-waite:
# Summary
Withdraws liquidity from the DPI LBP once the auction has finished. End state is the DAI is on the `compoundPCVDaiDeposit` and the `DPI` is on the TribalCouncil Safe. The DPI is intended to be sold via 1inch once on the TC safe. 

To `exitPool()` on the LBP, the `PCV_CONTROLLER` role is required - so this is a DAO proposal, which will likely be consolidated with other actions.


proposals/description/fip_104b.ts
```diff
@@ -0,0 +1,55 @@
+import { ProposalDescription } from '@custom-types/types';
+
+const fip_104b: ProposalDescription = {
+  title: 'FIP_104b: Withdraw LBP liquidity',
+  commands: [
+    {
+      target: 'pcvGuardianNew',
+      values: '0',
+      method: 'setSafeAddresses(address[])',
+      arguments: [['{tribalCouncilSafe}', '{tribalCouncilTimelock}']],
```
eswak: do not list the multisig as a safe address, the guardian could send funds there. listing the timelock is fine, as any movement is gated behind a 4 day timelock that can be veto'd

proposals/description/fip_104b.ts
```diff
@@ -0,0 +1,55 @@
...
+      description: 'Withdraw all DAI and DPI from LBP pool to the compoundDAIPCVDeposit'
+    },
+    {
+      target: 'ratioPCVControllerV2',
+      values: '0',
+      method: 'withdrawRatioERC20(address,address,address,uint256)',
+      arguments: [
+        '{compoundDaiPCVDeposit}', // pcvDeposit
+        '{dpi}', // token
+        '{tribalCouncilSafe}', // to
```
eswak, line 20: this is executed with PCV controller privilege and directly calls `withdrawERC20`, this does not need to be listed as a safe address

proposals/description/fip_104b.ts
```diff
@@ -0,0 +1,55 @@
...
+  ],
+  description: `
+  FIP_104b: Withdraw all liquidity from the DPI LBP. 
+
+  This FIP cleans up and finishes elements of the PCV reinforcement process snapshotted here:
+  https://snapshot.fei.money/#/proposal/0x2fd5bdda0067098f6c0520fe309dfe90ca403758f0ce98c1854a00bf38999674  
+  and discussed here: https://tribe.fei.money/t/fip-104-fei-pcv-reinforcement-proposal/4162 .
+
+  Specifically it:
+  - Sets the TribalCouncil multisig and timelock to be safe addresses for PCV withdrawal
```
eswak: 🚓🚨

proposals/description/fip_104b.ts
thomas-waite: Great point, have removed it

proposals/description/fip_104b.ts
thomas-waite: Yep, have removed from being a safe address

proposals/dao/end_tribe_incentives.ts
```diff
@@ -42,7 +40,21 @@ const deploy: DeployUpgradeFunc = async (deployAddress: string, addresses: Named
 // This could include setting up Hardhat to impersonate accounts,
 // ensuring contracts have a specific state, etc.
 const setup: SetupUpgradeFunc = async (addresses, oldContracts, contracts, logging) => {
-  console.log(`No actions to complete in setup for fip${fipNumber}`);
+  const stakeAmount = ethers.constants.WeiPerEther.mul(40_000);
+  const curve3Metapool = '0x06cb22615BA53E60D67Bf6C341a0fD5E718E1655';
```
thomas-waite, line 44: This is a fixture that sets up a staker to have some TRIBE rewards to claim. 

It is later relied on in an e2e test to ensure that the ending of incentives does not stop people who were staking LP tokens from claiming rewards. It will be removed along with the test once the proposal executes.

proposals/dao/fip_104b.ts
```diff
@@ -0,0 +1,67 @@
...
+// cleaned up before doing any validation checks.
+const teardown: TeardownUpgradeFunc = async (addresses, oldContracts, contracts, logging) => {
+  console.log(`No actions to complete in teardown for fip${fipNumber}`);
+};
+
+// Run any validations required on the fip using mocha or console logging
+// IE check balances, check state of contracts, etc.
+const validate: ValidateUpgradeFunc = async (addresses, oldContracts, contracts, logging) => {
+  // 1. Validate withdrawn liquidity destinations
+  const sanityCheckDAIAmount = ethers.constants.WeiPerEther.mul(3_500_000);
```
eswak: that's ~3% away from the current pool state, if DPI price changes too much before your proposal executes, the e2e setup step will fail, I think you should be more laxist on your sanity checks. For DPI for instance you checked balance of at least 1,500 where the spot pool content is 2,000, this gives more margin. Not a big deal but I wanted to share this with you, we had to update e2es because of this before

eswak (dismissed):

proposals/dao/fip_104b.ts
thomas-waite: That makes sense. I'll be less strict in the future on sanity checks

eswak (approved):

PR #829, opened 2022-05-23
Fix the inversion order in OracleRef

thomas-waite:
# Summary
The inversion and scalingFactor in OracleRef are applied in the wrong order, which in a situation where `doInvert = true` and `scalingFactory != 0` would lead to an incorrect price. The issue was found by Elliott when he was developing a USDC PSM.

Info below from his fix in the Volt repo:

```
Creating this PSM exposed an issue with the OracleRef contract, which is that the order of operations for inverting and then scaling was backwards which necessitated this change.

The current flow is:

Read -> scale -> invert

1. Read 1000000000000000000
2. Scale 1000000000000000000 * 1000000000000000000 (scaled by 1e18)
3. Invert 1 / 1000000000000000000000000000000000000
Result: 1e -36

This flow is incorrect, as the scale should have been applied after the inversion, not before, and thus the price which should have been 1 is now 0.000000000000000000000000000000000001

Vs

Read -> invert -> scale

1. Read 1000000000000000000
2. Invert 1 / 1000000000000000000
3. Scale 1 / 1000000000000000000 * 1000000000000000000 (scaled by 1e18)
Result: 1

```


ElliotFriedman (dismissed):
LGTM!

eswak:
LGTM, waiting for e2e tests to be fixed so we can merge with a green tick

eswak (approved):

PR #840, opened 2022-05-24
Acquire USDC and USDT from Curve by swapping DAI

thomas-waite:
# DO NOT MERGE
- Needs the FuseFixer contract deploying and adding into `mainnetAddresses.ts` before merge and proposal

## Summary
Uses Curve to swap $10.264M worth of DAI to at least 10,073,986 USDC and at least 133,178 USDT. 

Accepted slippage is ~0.55% in the USDC case and ~0.6% in the USDT case.

USTw can not be acquired from Curve due to liquidity having collapsed in pool 53 and there being only ~$4m available.


xklob (dismissed):

ElliotFriedman:
> Accepted slippage is ~0.55% in the USDC case and ~0.6% in the USDT case.

This feels very expensive when we could go directly to the MakerDAO PSM and redeem DAI for USDC without any fees. For the USDT swap, I think this should be tightened to 20 basis points at the maximum.

imperialzhiyi:
> > Accepted slippage is ~0.55% in the USDC case and ~0.6% in the USDT case.
> 
> This feels very expensive when we could go directly to the MakerDAO PSM and redeem DAI for USDC without any fees. For the USDT swap, I think this should be tightened to 20 basis points at the maximum.

Agree with the MakerDAO PSM for DAI-USDC case. Alternatively, compared with curve, 1inch might be the better option for good price. 1inch use MakerDAO PSM for DAI to USDC, and find the best routers for DAI to USDT.

thomas-waite:
> > Accepted slippage is ~0.55% in the USDC case and ~0.6% in the USDT case.
> 
> This feels very expensive when we could go directly to the MakerDAO PSM and redeem DAI for USDC without any fees. For the USDT swap, I think this should be tightened to 20 basis points at the maximum.

Hadn't considered using the MakerDAO PSM to directly redeem DAI, that's a great idea. Will use that instead if looks good. 

In terms of USDT, that has a high slippage to minimise any risk of the swap being rejected due to price movement whilst in the 4 day TC timelock. It won't be an issue in the future as we'll have these pipe swapper contracts. In the meantime, will do more investigation into likely price movements and lower if possible.

PR #872, opened 2022-06-01
proposal calldata

Joeysantoro:
# PR Type: Feature

- [ ] All Tests Passing

### DIFF
--- a/scripts/utils/constructProposalCalldata.ts
+++ b/scripts/utils/constructProposalCalldata.ts
@@ -1,8 +1,11 @@
 import constructProposal from './constructProposal';
 import { BigNumber } from 'ethers';
+import { ethers } from 'hardhat';
 import { Interface } from '@ethersproject/abi';
 import { utils } from 'ethers';
 import { getAllContractAddresses, getAllContracts } from '@test/integration/setup/loadContracts';
+import { ProposalCategory, ProposalDescription } from '@custom-types/types';
+import proposals from '@test/integration/proposals_config';
 
 type ExtendedAlphaProposal = {
   targets: string[];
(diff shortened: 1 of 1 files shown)


scripts/utils/constructProposalCalldata.ts
```diff
@@ -47,3 +58,38 @@ export async function constructProposalCalldata(proposalName: string): Promise<s
...
+  ]);
+
+  const combinedCalldatas = [];
+  for (let i = 0; i < proposal.targets.length; i++) {
+    const sighash = utils.id(proposal.signatures[i]).slice(0, 10);
+    combinedCalldatas.push(`${sighash}${proposal.calldatas[i].slice(2)}`);
+  }
+
+  const salt = ethers.utils.id(proposalInfo.title);
+  const predecessor = ethers.constants.HashZero;
```
ElliotFriedman, line 75: can predecessor always be 0? I never saw anywhere in the TimelockController where `predecessor` was actually stored.

scripts/utils/constructProposalCalldata.ts
```diff
@@ -47,3 +58,38 @@ export async function constructProposalCalldata(proposalName: string): Promise<s
...
+function getTimelockCalldata(proposal: ExtendedAlphaProposal, proposalInfo: ProposalDescription): string {
+  const proposeFuncFrag = new Interface([
+    'function scheduleBatch(address[] calldata targets,uint256[] calldata values,bytes[] calldata data,bytes32 predecessor,bytes32 salt,uint256 delay) public',
+    'function executeBatch(address[] calldata targets,uint256[] calldata values,bytes[] calldata data,bytes32 predecessor,bytes32 salt) public'
+  ]);
+
+  const combinedCalldatas = [];
+  for (let i = 0; i < proposal.targets.length; i++) {
+    const sighash = utils.id(proposal.signatures[i]).slice(0, 10);
+    combinedCalldatas.push(`${sighash}${proposal.calldatas[i].slice(2)}`);
```
ElliotFriedman, line 71: doesn't this assume that the calldatas are already fully expanded from the file?

Joeysantoro: what do you mean fully expanded?

Joeysantoro: this is the exact same code path to generate calldata for a DAO proposal, just uses the inputs in a different abi

scripts/utils/constructProposalCalldata.ts
Joeysantoro: its always 0 for these cases. That check would allow more complex flows which only execute B after A completed asynchronously

xklob (approved):

scripts/utils/constructProposalCalldata.ts
```diff
@@ -17,17 +20,25 @@ type ExtendedAlphaProposal = {
...
   const contractAddresses = await getAllContractAddresses();
 
-  const proposal = (await constructProposal(
-    proposalInfo.default,
-    contracts,
-    contractAddresses
-  )) as ExtendedAlphaProposal;
+  const proposal = (await constructProposal(proposalInfo, contracts, contractAddresses)) as ExtendedAlphaProposal;
 
+  console.log(proposals[proposalName].category);
```
ElliotFriedman, line 30: This console log fails if the proposal isn't in the proposal_config. I don't like the dependency on proposal_config.

PR #873, opened 2022-06-02
AURA airdrop claiming & locker contract

eswak:
Aura Airdrop claiming contract

this will upgrade the proxy implementation, claim airdrop, lock (to not get 30% penalty), and delegate to eswak.eth


contracts/metagov/VlAuraDelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,122 @@
...
+    /// even if the external smart contracts change.
+    function call(
+        address target,
+        uint256 value,
+        bytes calldata data
+    ) external onlyGovernor {
+        // solhint-disable-next-line
+        (bool success, ) = target.call{value: value}(data);
+        require(success, "call failed");
+    }
```
eswak: To discuss

contracts/metagov/VlAuraDelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/// @title Vote-locked AURA PCVDeposit
+/// @author eswak
+contract VlAuraDelegatorPCVDeposit is DelegatorPCVDeposit {
+    using SafeERC20 for IERC20;
+
+    address private constant AURA_TOKEN = 0x0000000000000000000000000000000000000000;
+    address private constant VLAURA_TOKEN = 0x0000000000000000000000000000000000000000;
+    address private constant AURA_AIRDROP = 0x0000000000000000000000000000000000000000;
```
eswak: TODO (Aura protocol is not deployed on mainnet yet)

contracts/metagov/VlAuraDelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,137 @@
...
+
+    function initialize(
+        address _aura,
+        address _auraLocker,
+        address _auraMerkleDrop,
+        address _initialDelegate
+    ) external {
+        // the Tribal Council multisig can initialize the state variables,
+        // only once, after contract deployment.
+        require(msg.sender == address(0x2EC598d8e3DF35E5D6F13AE2f05a7bB2704e92Ea), "!TC");
```
Joeysantoro: why not use role based auth. Seems fine though

contracts/metagov/VlAuraDelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,137 @@
...
+    function claim(
+        bytes32[] calldata _proof,
+        uint256 _amount,
+        bool _lock
+    ) external returns (bool);
+}
+
+/// @title Vote-locked AURA PCVDeposit
+/// @author eswak
+contract VlAuraDelegatorPCVDeposit is DelegatorPCVDeposit {
```
Joeysantoro, line 56: should it not also be a snapshot delegator?

Joeysantoro:
Will this contract even be eligible to lock?

contracts/metagov/VlAuraDelegatorPCVDeposit.sol
eswak: there is a 4 day timelock to create a new role that we would grant only to the TC multisig, better to just check the address

contracts/metagov/VlAuraDelegatorPCVDeposit.sol
eswak: 0xMaha confirmed the delegation must happen onchain & not in the gnosis registry

> (me) what does voting with vlAURA look like ? Snapshot ?

> (maha) Snapshot yes (at least initially) but on chain delegation
> As opposed to delegating via the snapshot ui

eswak:
> Will this contract even be eligible to lock?

Convex (& by extension, Aura) don't have whitelisting requirements to vote-lock : https://github.com/aurafinance/aura-contracts-lite/blob/main/contracts/AuraLocker.sol#L275-L281

eswak:
changed target to execute in the next dao vote

contracts/metagov/VlAuraDelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,133 @@
...
+        IAuraLocker(auraLocker).lock(address(this), amount);
+    }
+
+    /// @notice refresh lock after it has expired
+    function relock() external whenNotPaused onlyTribeRole(TribeRoles.METAGOVERNANCE_TOKEN_STAKING) {
+        IAuraLocker(auraLocker).processExpiredLocks(true);
+    }
+
+    /// @notice exit lock after it has expired
+    function unlock() external whenNotPaused onlyTribeRole(TribeRoles.METAGOVERNANCE_TOKEN_STAKING) {
```
ElliotFriedman, line 120: shouldn't governor also have this ability?

contracts/metagov/VlAuraDelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /// @notice claim AURA airdrop and vote-lock it for 16 weeks
+    /// this function is not access controlled & can be called by anyone.
+    function claimAirdropAndLock(bytes32[] calldata _proof, uint256 _amount) external returns (bool) {
+        return IAuraMerkleDrop(auraMerkleDrop).claim(_proof, _amount, true);
+    }
+
+    /// @notice lock AURA held on this contract to vlAURA
+    function lock() external whenNotPaused onlyTribeRole(TribeRoles.METAGOVERNANCE_TOKEN_STAKING) {
```
ElliotFriedman, line 108: shouldn't governor also have this ability?

contracts/metagov/VlAuraDelegatorPCVDeposit.sol
eswak: governor & TC can grant themselves the role before calling if we want to include this in a proposal

contracts/metagov/VlAuraDelegatorPCVDeposit.sol
eswak: governor & TC can grant themselves the role before calling if we want to include this in a proposal

contracts/metagov/DelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,71 @@
...
+/// This contract simply holds an ERC20 token, and delegate its voting power
+/// to an address. The ERC20 token needs to implement a delegate(address) method.
+/// @author eswak
+contract DelegatorPCVDeposit is PCVDeposit {
+    using SafeERC20 for IERC20;
+
+    event DelegateUpdate(address indexed oldDelegate, address indexed newDelegate);
+
+    /// @notice the token that is being used for voting
+    ERC20Votes public token;
```
thomas-waite, line 18: Token can never be changed past constructor time, so can set to `immutable`

contracts/metagov/DelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,71 @@
...
+    /// @notice withdraw tokens from the PCV allocation
+    /// @param amount of tokens withdrawn
+    /// @param to the address to send PCV to
+    function withdraw(address to, uint256 amount) external virtual override onlyPCVController {
+        IERC20(token).safeTransfer(to, amount);
+        emit Withdrawal(msg.sender, to, amount);
+    }
+
+    /// @notice no-op
+    function deposit() external override {}
```
thomas-waite, line 45: Is it worthwhile emitting the `Deposit()` event like in `ERC20HoldingPCVDeposit` to signal that the contract has received funds? No strong feelings

contracts/metagov/VlAuraDelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,133 @@
...
+        uint256 amount = IERC20(aura).balanceOf(address(this));
+        IERC20(aura).safeApprove(auraLocker, amount);
+        IAuraLocker(auraLocker).lock(address(this), amount);
```
thomas-waite, line 111: Is it useful for off-chain accounting in anyway to fire an event from this contract on locking?

contracts/metagov/VlAuraDelegatorPCVDeposit.sol
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /// @notice refresh lock after it has expired
+    function relock() external whenNotPaused onlyTribeRole(TribeRoles.METAGOVERNANCE_TOKEN_STAKING) {
```
thomas-waite, line 115: There's a fair few functions in this contract related to Aura that we don't test. 

They're only thin wrappers around the Aura contracts but would be more comfortable if we add an integration test that confirms each in foundry, unless there's time pressure which means we can't (does this need to go in this DAO vote?)

thomas-waite (commented):
Looks good overall. I don't have much context with the Aura system so would feel more comfortable if we had an integration test in Foundry which shows that this works with the Aura system as expected, even though the contract looks simple enough. 

Specifically, would want to see locking, unlocking + claiming rewards working as expected. Unless there's time pressure which means this is urgent/has to go in this DAO vote and not a follow up

contracts/metagov/VlAuraDelegatorPCVDeposit.sol
eswak: after discussion, probably not worth the time for testing everything (e2es would have to significantly mock/impersonate the contracts in the aura system, and the contract will only hold ~20k$, and the implementation is upgradeable)

contracts/metagov/DelegatorPCVDeposit.sol
eswak: actually it is set in `VlAuraDelegatorPCVDeposit` because the address of token was not known at time of deploy (and it has an initialize() function to set the state variables)

proposals/dao/tip_118.ts
```diff
@@ -72,15 +72,22 @@ const deploy: DeployUpgradeFunc = async (deployAddress: string, addresses: Named
...
   return {
-    angleEuroRedeemer,
+    vlAuraDelegatorPCVDepositImplementation
+    /*angleEuroRedeemer,
     wethHoldingPCVDeposit,
     lusdHoldingPCVDeposit,
     voltHoldingPCVDeposit,
     daiHoldingPCVDeposit,
-    gOHMHoldingPCVDeposit
+    gOHMHoldingPCVDeposit*/
```
Joeysantoro: TODO: uncomment

proposals/dao/tip_118.ts
```diff
@@ -47,7 +47,7 @@ const MAX_REMAINING_CHIEF_BALANCE = ethers.constants.WeiPerEther.mul(50_000);
 // This should exclusively include new contract deployments
 const deploy: DeployUpgradeFunc = async (deployAddress: string, addresses: NamedAddresses, logging: boolean) => {
   ////// Deploy empty PCV deposits for remaining PCV assets
-  const ERC20HoldingPCVDepositFactory = await ethers.getContractFactory('ERC20HoldingPCVDeposit');
+  /*const ERC20HoldingPCVDepositFactory = await ethers.getContractFactory('ERC20HoldingPCVDeposit');
```
Joeysantoro: TODO uncomment

PR #879, opened 2022-06-05
General cleanup tasks

xklob:
# PR Type: Feature

# PR Title: General Cleanup Tasks

### PR Description
- Enable strictmode and noImplicitAny
- removes replaceArgs and replaces the proposal with a templatedProposal
- removes extraneous imports in many test files
- adds typings for all missing typings in tests that currently exist

- [x] All Tests Passing


proposals/description/end_tribe_incentives.ts
```diff
@@ -1,14 +1,17 @@
...
+const end_tribe_incentives: TemplatedProposalDescription = {
   title: 'TIP-109: Discontinue Tribe Incentives',
   commands: [
     /////// TC grants itself TRIBAL_CHIEF_ADMIN_ROLE
     {
       target: 'core',
       values: '0',
       method: 'grantRole(bytes32,address)',
-      arguments: ['0x23970cab3799b6876df4319661e6c03cc45bd59628799d92e988dd8cbdc90e31', '{tribalCouncilTimelock}'],
+      arguments: (addresses) => [
```
ElliotFriedman, line 11: feels very strange passing a function as the value for the arguments key.

ElliotFriedman: what's the advantage or rational for doing this?

xklob: (see comments in fei-dev channel in discord)

protocol-configuration/mainnetAddresses.ts
```diff
@@ -2085,4 +2085,4 @@ const MainnetAddresses: MainnetAddresses = {
   }
 };
 
-export default MainnetAddresses;
+export default MainnetContractsConfig;
```
ElliotFriedman, line 2088: Why the name change? `MainnetAddresses` makes sense.

types/types.ts
```diff
@@ -70,36 +70,64 @@ export enum ProposalCategory {
...
-export type ProposalDescription = {
+export interface ProposalDescription {
   title: string;
   commands: ProposalCommand[];
   description: string;
-};
+}
 
-export type ProposalCommand = {
+export interface TemplatedProposalDescription {
```
ElliotFriedman, line 107: why the name changes here? the old names made sense.

xklob: type > interface because you should always use interfaces over types where sensible (they're full objects, not just aliases, and are thus capable of introspection)

added in the templatedproposal stuff in addition to the normal proposal stuff, to enable the replacement for replaceArgs

protocol-configuration/mainnetAddresses.ts
xklob: IMO it doesn't; it's more than just addresses, it's:
- addresses
- artifactnames
- category

therefore the naming should reflect that

proposals/description/end_tribe_incentives.ts
eswak: tldr; being able to use typescript

test/integration/tests/podOperation.ts
```diff
@@ -103,8 +103,8 @@ describe('Pod operation and veto', function () {
 
     // 1. Deploy a pod through which a proposal will be executed
     const deployTx = await podFactory.connect(tribalCouncilTimelockSigner).createOptimisticPod(podConfig);
-    const { args } = (await deployTx.wait()).events.find((elem) => elem.event === 'CreatePod');
-
+    const result = (await deployTx.wait()).events!.find((elem) => elem.event === 'CreatePod');
+    const args = result!.args!;
```
eswak, line 107: what's this syntax?

eswak (approved):
oof, gj

test/integration/tests/podOperation.ts
xklob: It tells Typescript that an object that is possibly undefined is not undefined

PR #885, opened 2022-06-06
Withdraw from Tokemak

thomas-waite:
# PR Type: Protocol ops

# Schedule withdrawal from Tokemak
As part of the PCV consolidation moves, this PR requests a withdrawal of the 10k WETH in Tokemak. A subsequent PR will have the DAO call `withdraw()` on the PCV deposit to remove the WETH once withdrawn. 

Tokemak operates in terms of cycles. The cycle length is 604800 seconds (7 days) and the next cycle starts on the 8th June. This withdrawal request will allow us to withdraw in the following cycle that starts on the 15th June. 

The PR also claims our rewards from Tokemak, which will be withdrawn at a later date.

## Rewards
Seperate to this PR, rewards will be harvested at the start of the next cycle


eswak (dismissed):
The code looks fine, but the PR description is a bit off : 

> A subsequent PR will have the DAO call withdraw() on the PCV deposit to remove the WETH once withdrawn.

This can be done with the PCVGuardian (so either through TC Timelock or Guardian multisig), after the withdrawal request has been made & Tokemak rolled over to the next cycle, a standard `withdraw()` or `withdrawRatio()` call can be used to move ETH to a safe address.

> The PR also claims our rewards from Tokemak, which will be withdrawn at a later date.

Anyone can claim rewards on behalf of the PCVDeposit, so we'll do this with one of our deployer addresses. Tokemak team takes time to sign reward messages so we should wait ~1 month after withdraw() before claiming rewards to make sure everything that has been accrued (and the TOKE can then be moved by PCVGuardian with `withdrawERC20()`).

eswak (approved):

PR #891, opened 2022-06-08
NopeDAO V2

thomas-waite:
# NopeDAO V2
This PR restricts Nope proposals to affirmative only votes - votes in favour of a Nope veto being performed. This is to enforce the concept of negative consent and prevent a veto potentially being blocked by anti-veto votes. 

## Notable points:
1. Users will not be able to retract a veto vote. Once they have voted for, it is permanent
2. There is no way to cancel proposal/veto if there is a mistake. There is no easy way to do this in a safe way at scale
3. Convenience `propose()` and `execute()` methods are introduced. These take in veto related parameters such as `podId` to make it simpler to propose and execute a veto.

## TODO
- [ ] Request Tally UI changes to remove the vote against button


contracts/test/unit/governance/GovernorQuickReaction.t.sol
```diff
@@ -0,0 +1,111 @@
...
+        uint256 proposalId = nopeDAO.propose(targets, values, calldatas, description);
+        vm.roll(block.number + 1); // Make vote active
+        assertEq(uint8(nopeDAO.state(proposalId)), 1);
+
+        vm.expectRevert(bytes("Governor: proposal not successful"));
+        nopeDAO.execute(targets, values, calldatas, descriptionHash);
+    }
+
+    /// @notice Validate can quick reaction execute successful proposal
+    function testQuickReactionProposal() public {
```
thomas-waite, line 85: TODO: Add a test where I fast forward past the end of the `votingPeriod` and verify that the state update works as expected

foundry.toml
```diff
@@ -1,8 +1,5 @@
-[default]
+[profile.default]
```
thomas-waite, line 1: Latest version of Foundry suggested doing this. Had warnings

contracts/dao/nopeDAO/GovernorCountingFor.sol
```diff
@@ -0,0 +1,82 @@
...
+        return true;
+    }
+
+    /**
+     * @dev See {Governor-_countVote}. In this module, only one vote type FOR is enabled
+     */
+    function _countVote(
+        uint256 proposalId,
+        address account,
+        uint8 support,
```
thomas-waite, line 69: Maintaining standard `_countVote()` interface

contracts/test/unit/governance/GovernorQuickReaction.t.sol
thomas-waite: done

protocol-configuration/mainnetAddresses.ts
```diff
@@ -224,12 +224,19 @@ export const MainnetContractsConfig = {
     category: AddressCategory.Governance
   },
 
-  nopeDAO: {
+  nopeDAOV1: {
```
xklob, line 227: Would really prefer that we not rename contracts like this

proposals/description/nope_dao_v2.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const nope_dao_v2: TemplatedProposalDescription = {
+  title: 'TIP-X: NopeDAO V2',
+  commands: [
+    {
+      target: 'core',
+      values: '0',
+      method: 'createRole(bytes32,bytes32)',
+      arguments: (addresses) => [POD_VETO_ADMIN, GOVERN_ROLE],
+      description: 'Transfer admin of the POD_VETO_ADMIN to the GOVERN_ROLE, held by the DAO timelock'
```
xklob, line 15: Why does the description say transfer, but the method is "createRole"?

contracts/test/utils/GovFixtures.sol
```diff
@@ -0,0 +1,112 @@
...
+    function getVariable() external view returns (uint256) {
+        return variable;
+    }
+
+    function setVariable(uint256 x) external {
+        variable = x;
+    }
+}
+
+/// @notice Setup various users with Tribe delegated to them for DAO voting
```
xklob: Can you add a bit of description here? This utility function seems pretty useful - it'd be more use if the description was more verbose.

proposals/description/nope_dao_v2.ts
thomas-waite: `createRole` will create a non-created role, but it also has the side effect whereby if a role already exists then when you 'create' it, you are actually setting it's admin.

```
    /// @notice creates a new role to be maintained
    /// @param role the new role id
    /// @param adminRole the admin role id for `role`
    /// @dev can also be used to update admin of existing role
    function createRole(bytes32 role, bytes32 adminRole) external override onlyGovernor {
        _setRoleAdmin(role, adminRole);
    }
```

`transfer` best describes what is functionally happening

contracts/test/utils/GovFixtures.sol
thomas-waite: Sure, done

protocol-configuration/mainnetAddresses.ts
thomas-waite: Happy to chat about this further, but I quite like the idea of versioning things and then having a default name for a contract

PR #898, opened 2022-06-12
PodExecutorV2

thomas-waite:
# Feature

# PodExecutorV2

Fixes a bug in the `PodExecutor.sol` whereby the `executeBatch()` method was not correctly exposed. Also updates an event to emit the `proposalId` rather than a custom hash of the data being executed.

Adds much more extensive testing, with most state variables that are touched being tested.

## Benefit of a PodExecutor contract vs making EXECUTOR an open role
The main benefit using a PodExecutor contract vs making the EXECUTOR role an open role and publicly available is that the PodExecutor is pauseable. It will allow the Guardian to pause the contract and prevent execution if an issue is found with a queued transaction.

## TODO
- [x] Add test for proposalId emitted in event
- [ ] Deploy new contract and validate
- [x] e2e tests for `execute()` and `executeBatch()`
- [x] Deprecate old EXECUTOR
- [x] TC proposal to grant EXECUTOR role to this contract
- [x] Verification that a proposal can be batch executed through this

- [x] All Tests Passing
- [x] Proposal Added to ProposalsConfig
- [x] Fork Block Correct
- [x] Update Documentation If Needed
- [x] Update Roles Config
- [ ] Proposal Submitted


protocol-configuration/mainnetAddresses.ts
```diff
@@ -2038,11 +2038,16 @@ const MainnetContractsConfig: MainnetContractsConfig = {
...
     category: AddressCategory.Volt
   },
-  podExecutor: {
+  oldPodExecutor: {
     artifactName: 'PodExecutor',
     address: '0x99d8b669F48A708f7C0373AF6CE31F0726ab6CaD',
-    category: AddressCategory.Governance
+    category: AddressCategory.Deprecated
   },
+  // podExecutor: {
```
thomas-waite: TODO: Fill in once deployed

test/integration/tests/pcv.ts
```diff
@@ -232,9 +232,9 @@ describe('e2e-pcv', function () {
         params: [aaveEthPCVDeposit.address]
       });
 
-      await aaveEthPCVDeposit.withdrawERC20(await aaveEthPCVDeposit.aToken(), deployAddress, tenPow18.mul(toBN(10000)));
```
thomas-waite, line 243: Reduced all figures in this test by 10x, to account for the balance of the on-chain Aave PCV deposit having fewer funds

Joeysantoro (dismissed):

contracts/dao/timelock/ITimelock.sol
```diff
@@ -1,16 +0,0 @@
-// SPDX-License-Identifier: GPL-3.0-or-later
```
xklob, line 1: Why was this file deleted?

proposals/description/pod_exec_v2.ts
```diff
@@ -0,0 +1,47 @@
+import { TemplatedProposalDescription } from '@custom-types/types';
+
+const pod_executor_v2: TemplatedProposalDescription = {
+  title: 'Pod Executor V2',
+  commands: [
+    {
+      target: 'tribalCouncilTimelock',
+      values: '0',
+      method: 'grantRole(bytes32,address)',
+      arguments: (addresses) => [
```
xklob: Let's start using `ethers.utils.id` for roles instead of the raw bytes

protocol-configuration/mainnetAddresses.ts
xklob: I really don't like changing the names of contracts in this file to upgrade to new ones (or deprecate old ones). Anyone that reads from this file (like Storm) will probably be assuming that entries don't change except for their category (if deprecated).

Can we use podExecutorV2 as the name instead?

contracts/dao/timelock/ITimelock.sol
thomas-waite: Not needed, was wrongly added by me at the start of governance. Should have been just importing `TimelockController`

proposals/description/pod_exec_v2.ts
thomas-waite: Good idea, done and will do going forward

protocol-configuration/mainnetAddresses.ts
thomas-waite: Sure, agreed

xklob (dismissed):

xklob (dismissed):

eswak (dismissed):
lgtm

eswak (approved):

PR #900, opened 2022-06-13
Auto generate metadata for TC proposals

thomas-waite:
# Feature: Auto-generate the metadata for TC proposals

## Summary
Automatically generates the metadata for registering TC proposals in the `GovernanceMetadataRegistry`: https://etherscan.io/address/0xd78Cd3AaE6168BE43B548877aAc68312B9df9AFb 

### Context
Registering proposals in the metadata registry is currently too high friction. It requires a seperate transaction by an engineer after a proposal has been queued on the TC timelock. 

### Solution
This PR outputs the calldata for the register metadata transaction that should be submitted to the registry at the same time the proposal is queued. 

The `podId` is known for the TC (25) and the timelock `proposalID` can be computed deterministically ahead of time locally. The `metadata` human readable description of what the proposal does is taken from the `proposals/description/fip_x.ts`  proposal.

## New proposal submission flow
When submitting a TC proposal on the TC safe:
1. Run `DEPLOY_FILE=fip_x npm run calldata`
2. Get the `Calldata` (which will schedule the proposal) and the `Metadata calldata` (which will register the human readable description)
3. Use the Gnosis Safe Transaction builder (turn on Custom data) to schedule two transactions:
    a. Call to the TC timelock (`0xe0C7DE94395B629860Cbb3c42995F300F56e6d7a`) with the `Calldata`
    b. Call to the Governance Metadata registry (`0xd78Cd3AaE6168BE43B548877aAc68312B9df9AFb`) with `Metadata calldata`
4. Queue that transaction as normal
5. When it comes to executing the proposal in the TC timelock, execute as normal

- [ ] All Tests Passing
- [ ] Proposal Added to ProposalsConfig
- [x] Fork Block Correct
- [ ] Update Documentation If Needed
- [x] Update Roles Config
- [ ] Proposal Submitted


scripts/utils/constructProposalCalldata.ts
```diff
@@ -15,6 +16,10 @@ type ExtendedAlphaProposal = {
   description: string;
 };
 
+type PodConfig = {
```
xklob: Use interfaces over types wherever possible, as they're capable of introspection and not just aliases

scripts/utils/constructProposalCalldata.ts
```diff
@@ -59,12 +64,18 @@ function getDAOCalldata(proposal: ExtendedAlphaProposal): string {
...
+  proposal: ExtendedAlphaProposal,
+  proposalInfo: TemplatedProposalDescription,
+  podConfig?: PodConfig
+): string {
   const proposeFuncFrag = new Interface([
     'function scheduleBatch(address[] calldata targets,uint256[] calldata values,bytes[] calldata data,bytes32 predecessor,bytes32 salt,uint256 delay) public',
     'function executeBatch(address[] calldata targets,uint256[] calldata values,bytes[] calldata data,bytes32 predecessor,bytes32 salt) public'
   ]);
 
+  const metadataFuncFrag = new Interface(['function registerProposal(uint256,uint256,string) external']);
```
xklob: Would be better to import these function fragments from a compiled ABI than to inline them

scripts/utils/constructProposalCalldata.ts
```diff
@@ -91,5 +102,25 @@ function getTimelockCalldata(proposal: ExtendedAlphaProposal, proposalInfo: Temp
...
 
-  return `Calldata: ${calldata}\nExecute Calldata: ${executeCalldata}`;
+  if (podConfig) {
+    const proposalId = calcProposalId(proposal.targets, proposal.values, combinedCalldatas, predecessor, salt);
+    const registerMetadataCalldata = metadataFuncFrag.encodeFunctionData('registerProposal', [
+      podConfig.id,
+      proposalId,
+      proposal.description
+    ]);
+    return `Calldata: ${calldata}\nMetadata Calldata: ${registerMetadataCalldata}\nExecute Calldata: ${executeCalldata}`;
```
xklob: It's strange to me that we're returning a formatted string. If we want to log this to console, we should do that explicitly; the string returned should just be the pure calldata, imo.

scripts/utils/constructProposalCalldata.ts
```diff
@@ -91,5 +102,25 @@ function getTimelockCalldata(proposal: ExtendedAlphaProposal, proposalInfo: Temp
...
+}
+
+export function calcProposalId(
+  targets: string[],
+  values: BigNumber[],
+  payloads: string[],
+  predecessor: string,
+  salt: string
+): string {
+  return '0x123';
```
xklob: I'm guessing this is a placeholder to fix errors; my go-to for this is just to explicitly throw an error, ie:

`throw new Error("Method not yet implemented")`

this way I don't accidentally use a placeholder'd function if I forget to finish it.

scripts/utils/simulateTimelockProposal.ts
```diff
@@ -1,14 +1,14 @@
...
-  TemplatedProposalDescription
-} from '@custom-types/types';
+import { MainnetContracts, NamedAddresses, TemplatedProposalDescription } from '@custom-types/types';
 import { OptimisticTimelock } from '@custom-types/contracts';
 import { getImpersonatedSigner, time } from '@test/helpers';
 import { Contract } from '@ethersproject/contracts';
 import { forceEth } from '@test/integration/setup/utils';
+import { TRIBAL_COUNCIL_POD_ID } from '@protocol/optimisticGovernance';
+
+type PodConfig = {
```
xklob: Would be better to import this since you're using it in multiple places

thomas-waite: Yep, this is WIP

scripts/utils/constructProposalCalldata.ts
thomas-waite: It's WIP, implementing locally but not pushed - that error throwing should definitely be added for medium term functions that are yet to be implemented

scripts/utils/constructProposalCalldata.ts
thomas-waite: Nice idea for objects for sure, will update

scripts/utils/constructProposalCalldata.ts
thomas-waite: This function generates the string of calldata, returns that and it is later logged at the top level. I personally like that so you don't have logging going on several layers of abstraction down, it's clear where it's coming from

scripts/utils/constructProposalCalldata.ts
thomas-waite: Yeah good idea, done

scripts/utils/constructProposalCalldata.ts
xklob: Yep, that's what I'm saying - the logging should happen at the higher level

Joeysantoro (dismissed):

xklob (approved):

PR #914, opened 2022-06-16
category refactor

Joeysantoro:
# PR Type: Feature

- [ ] All Tests Passing
- [ ] Remove Any .only's on Tests


types/types.ts
```diff
@@ -154,6 +154,9 @@ export interface ContractConfig {
 export enum AddressCategory {
   Core = 'Core',
   Governance = 'Governance',
+  Buybacks = 'Buybacks',
+  PCV_Mover = 'PCV_Mover',
```
xklob: What about "Utility" instead?

types/types.ts
```diff
@@ -154,6 +154,9 @@ export interface ContractConfig {
 export enum AddressCategory {
   Core = 'Core',
   Governance = 'Governance',
+  Buybacks = 'Buybacks',
```
xklob: This could also fall under utility; seems weird to give it its own category

PR #935, opened 2022-06-26
Fuse withdraw sentinel

Joeysantoro:
# PR Type: Feature

### PR Description: 
A sentinel that automatically detects available liquidity in a Fuse pool and withdraws, with a configurable minimum amount left.

- [x] Enhance testing
- [x] All Tests Passing


contracts/sentinel/guards/FuseWithdrawalGuard.sol
```diff
@@ -0,0 +1,131 @@
...
+        }
+        return false;
+    }
+
+    /// @notice return the amount that can be withdrawn from a deposit after leaving min liquidity
+    function getAmountToWithdraw(ERC20CompoundPCVDeposit deposit) public view returns (uint256) {
+        // can't read underlying directly because some of the PCV deposits use an old abi
+        IERC20 underlying = IERC20(withdrawInfos[address(deposit)].underlying);
+        // Reserves of underlying left in the cToken are considered withdrawable liquidity
+        uint256 liquidity = underlying.balanceOf(address(deposit.cToken()));
```
eswak, line 78: is that correct ? ctokens should be holding the platform fees & reserves too ?

contracts/sentinel/guards/FuseWithdrawalGuard.sol
```diff
@@ -0,0 +1,131 @@
...
+    function setWithdrawInfo(address deposit, WithdrawInfo calldata withdrawInfo)
+        public
+        onlyTribeRole(TribeRoles.GUARDIAN)
+    {
+        withdrawInfos[deposit] = withdrawInfo;
+    }
+
+    /// @notice helper method to trustlessly and publicly remove deposits once emptied.
+    /// TODO could be removed to simplify/minimize attack surface and public functions.
+    function removeDepositIfEmpty(ERC20CompoundPCVDeposit deposit) public {
```
thomas-waite: Don't have strong feelings on this. Would probably lean on the side of removing/restricting to only the guardian to minimise the surface area

contracts/sentinel/guards/FuseWithdrawalGuard.sol
```diff
@@ -0,0 +1,131 @@
...
+        address[] memory destinations,
+        address[] memory underlyings,
+        uint96[] memory liquidityToLeaveList
+    ) CoreRef(core) {
+        uint256 len = deposits.length;
+        require(len == destinations.length && len == liquidityToLeaveList.length && len == underlyings.length);
+        for (uint256 i = 0; i < len; ) {
+            fuseDeposits.add(deposits[i]);
+            withdrawInfos[deposits[i]] = WithdrawInfo({
+                destination: destinations[i],
```
thomas-waite, line 44: Is it worthwhile checking that all destinations are `safeAddresses`? `pcvGuardian.isSafeAddress()`? Then can guarantee we can recover the transferred out funds and not make a mistake by setting an incorrect address

contracts/sentinel/guards/FuseWithdrawalGuard.sol
```diff
@@ -0,0 +1,131 @@
...
+            }
+        }
+    }
+
+    /// @notice setter for the Fuse deposit destination and minimum liquidity
+    function setWithdrawInfo(address deposit, WithdrawInfo calldata withdrawInfo)
+        public
+        onlyTribeRole(TribeRoles.GUARDIAN)
+    {
+        withdrawInfos[deposit] = withdrawInfo;
```
thomas-waite, line 59: Would probably check the destination is a safe address

thomas-waite (commented):
Really like the idea of this overall. I would check that the destinations are safe addresses probably to make sure there's no accidental mistake in where the recovered liquidity is sent

contracts/sentinel/guards/FuseWithdrawalGuard.sol
Joeysantoro: PCV Guardian can only transfer to a safe address so this is baked in

contracts/sentinel/guards/FuseWithdrawalGuard.sol
Joeysantoro: this is done by PCV guardian

contracts/sentinel/guards/FuseWithdrawalGuard.sol
Joeysantoro: Will remove

contracts/sentinel/guards/FuseWithdrawalGuard.sol
Joeysantoro: They are withdrawable either way. Platform fees do not have a liquidity preference

contracts/test/integration/sentinel/FuseWithdrawalGuard.t.sol
```diff
@@ -0,0 +1,101 @@
...
+        uint96[] memory liquidityToLeaveList = new uint96[](len);
+
+        deposits[0] = MainnetAddresses.RARI_POOL_8_FEI_PCV_DEPOSIT;
+        deposits[1] = MainnetAddresses.RARI_POOL_8_LUSD_PCV_DEPOSIT;
+
+        underlyings[0] = MainnetAddresses.FEI;
+        underlyings[1] = MainnetAddresses.LUSD;
+
+        destinations[0] = MainnetAddresses.DAI_PSM;
+        destinations[1] = MainnetAddresses.LUSD_PSM;
```
Joeysantoro, line 39: if LUSD psm is deprecated, this needs to change to the native deposit

contracts/sentinel/guards/FuseWithdrawalGuard.sol
```diff
@@ -0,0 +1,126 @@
...
+        uint96 liquidityToLeave;
+    }
+
+    /// @notice map the destination and minimum liquidity for each pcv deposit
+    mapping(address => WithdrawInfo) public withdrawInfos;
+
+    EnumerableSet.AddressSet private fuseDeposits;
+
+    /// @notice the PCV mover contract exposed to guardian role
+    PCVGuardian public constant pcvGuardian = PCVGuardian(0x02435948F84d7465FB71dE45ABa6098Fc6eC2993);
```
xklob, line 25: Would prefer if this address was declared as a constant at the top of the file

contracts/sentinel/guards/FuseWithdrawalGuard.sol
```diff
@@ -0,0 +1,126 @@
...
+
+    /// @notice added to prevent dust from bricking the contract
+    uint256 public constant MIN_WITHDRAW = 100e18;
```
xklob, line 30: Unclear naming - is this min ctoken withdrawal, or underlying? Might want to update var name or comment.

contracts/sentinel/guards/FuseWithdrawalGuard.sol
```diff
@@ -0,0 +1,126 @@
...
+
+    constructor(
+        address core,
+        address[] memory deposits,
+        address[] memory destinations,
+        address[] memory underlyings,
+        uint96[] memory liquidityToLeaveList
+    ) CoreRef(core) {
```
xklob, line 38: Why is this coreRef? Guards ideally shouldn't rely on external state, which CoreRef brings in.

contracts/sentinel/guards/FuseWithdrawalGuard.sol
```diff
@@ -0,0 +1,126 @@
...
+        // max withdraw is the pcv deposit balance
+        uint256 withdrawAmount = deposit.balance();
+        if (withdrawAmount > liquidity) {
+            withdrawAmount = liquidity;
+        }
+        return withdrawAmount > MIN_WITHDRAW ? withdrawAmount : 0;
+    }
+
+    /// @notice return the first element which can be withdrawn from with the appropriate calldata encoding tha max withdraw amount.
+    /// @dev it only returns one element for simplicity. Unlikely multiple will be withdrawable simulteneously after the first pass, and pruning empty entries from a sparse array is an inefficient and inelegant algorithm in solidity.
```
xklob, line 95: Good explanation, appreciate

xklob (changes requested):
See couple of comments. Overall looks good but would prefer if this was switched to having immutable state (which would also allow you to remove the dependency on CoreRef)

contracts/sentinel/guards/FuseWithdrawalGuard.sol
xklob: Nvm this is fine.

xklob:
See couple of comments. Overall looks good but would prefer if this was switched to having immutable state (which would also allow you to remove the dependency on CoreRef)

eswak:
lgtm, would like to see an operational deployment next (in the follow-up PR!)

contracts/sentinel/guards/FuseWithdrawalGuard.sol
Joeysantoro: PCV deposits are always denominated in underlying, I think the naming is fine but can update comment

xklob (dismissed):

protocol-configuration/proposalsConfig.ts
```diff
@@ -1,10 +1,12 @@
 import { ProposalCategory, TemplatedProposalDescription, TemplatedProposalsConfigMap } from '@custom-types/types';
 
+import fip_x from '@proposals/description/fip_x';
+
 const proposals: TemplatedProposalsConfigMap = {
   fuse_withdrawal_guard: {
     deploy: true, // deploy flag for whether to run deploy action during e2e tests or use mainnet state
     totalValue: 0, // amount of ETH to send to DAO execution
-    proposal: tip_117, // full proposal file, imported from '@proposals/description/fip_xx.ts'
+    proposal: fip_x as unknown as TemplatedProposalDescription, // temporary hacky workaround, this field isn't needed but can't be undefined
```
xklob: Let's change the base type of this `proposal` var within the `TemplatedProposalsConfigMap` to include undefined as an acceptable type

xklob (approved):

PR #922, opened 2022-06-20
DAO vote 2: Incentives, deprecate PSMs, holding PCV deposits, agEUR redeem

thomas-waite:
# DAO vote 2: Incentives, deprecate PSMs, holding PCV deposits, agEUR redeem
Batches various cleanup and technical maintenance tasks into a second DAO vote:
1. Incentives cleanup/withdrawal
2. Introduction of Holding ERC20 PCV deposits and deprecation of remaining PSMs
3. Redemption of Angle and agEUR

## Incentives cleanup
This proposal funds the owed outgoing incentives and then withdraws TRIBE from various areas in the system: Tribal Chief, ERC20 Dripper, Votium contracts and sends it to the Core treasury.

It also transfers the owner of the Fei Aave incentives controller from the TRIBE DAO to the Aave DAO.

### Key numbers
Amount of TRIBE:
- On TribalChief: 29.2M
- Owed in pending rewards to LP token stakers: 565k
- On ERC20 Dripper: 2.5M

Total to be recovered: 31.3M TRIBE

## Implementation
Off-chain, all staking token wrapper harvests have been performed.

The final deprecation steps performed are:
1. Withdraw from the TribalChief all remaining TRIBE, minus the total pending rewards owed to the historical stakers of Uniswap-v2 FEI/TRIBE LP, Curve 3crv-FEI metapool LP and G-UNI DAI/FEI 0.05% fee tier. Use `governorWithdrawTribe(uint256 amount)`
5. Withdraw TRIBE from ERC20Dripper
6. Withdraw all TRIBE from the Votium contracts
7. Transfer ownership of Fei Aave incentives controller to Aave governance

PR #949, opened 2022-07-01
Add gOHM Oracle, add deposit to CR

thomas-waite:
# Add gOHM Oracle, add deposit to CR

Creates a gOHM oracle contract, adds it to the CO and then adds the gOHM holding PCV deposit to the CO.


contracts/test/integration/oracle/GOhmOracleIntegrationTest.t.sol
```diff
@@ -0,0 +1,63 @@
...
+        gOhmEthOracle = new GOhmEthOracle(MainnetAddresses.CORE, chainlinkOHMEthOracle);
+    }
+
+    function testValidgOHMPrice() public {
+        (Decimal.D256 memory gOhmEthPrice, ) = gOhmEthOracle.read();
+
+        // Eth price is ~$1000. gOHM price is ~$2400
+        // Therefore, gOHM price in ETH should be ~ (2400/1000) = 2.4 ETH
+        assertGt(gOhmEthPrice.value, 1e18);
+        assertLt(gOhmEthPrice.value, 5e18);
```
eswak: that doesn't look like an integration test that will age well... Maybe deploy an `MockChainlinkOracle` in setUp() and use that fixed gOHM/ETH value as a source of truth in your computations. Doesn't solve the index rebase though. You could set the internal state of the index in the gohm contract with `vm.store(address,slot,value)`

proposals/dao/tip_118.ts
```diff
@@ -273,8 +273,8 @@ const validate: ValidateUpgradeFunc = async (addresses, oldContracts, contracts,
   console.log('----------------------------------------------------');
 
   // PCV Equity change should be neutral for this proposal
-  expect(Number(eqDiff) / 1e18).to.be.at.least(-10000);
-  expect(Number(eqDiff) / 1e18).to.be.at.most(+10000);
+  // expect(Number(eqDiff) / 1e18).to.be.at.least(-10000);
+  // expect(Number(eqDiff) / 1e18).to.be.at.most(+10000);
```
eswak: pushed the fix in #922 you can update your base branch ^^

proposals/dao/tip_119.ts
```diff
@@ -0,0 +1,116 @@
...
+  const gOhmUSDOracle = contracts.gOhmUSDOracle;
+  const collateralizationOracle = contracts.collateralizationOracle;
+  const gOHMHoldingPCVDeposit = contracts.gOHMHoldingPCVDeposit;
+
+  // 1. Validate gOHM ETH oracle price is valid
+  const gOhmETHPrice = (await gOhmEthOracle.read())[0];
+  // Eth price is ~$1000. gOHM price is ~$2400
+  // Therefore, gOHM price in ETH should be ~ (2400/1000) = 2.4 ETH
+  expect(toBN(gOhmETHPrice.value)).to.be.bignumber.at.least(ethers.constants.WeiPerEther.mul(1));
+  expect(toBN(gOhmETHPrice.value)).to.be.bignumber.at.most(ethers.constants.WeiPerEther.mul(5));
```
eswak, line 96: it's fine to do this check here because proposal scripts are bounded in time, unlike integration tests (in a few weeks we won't be running this file anymore)

proposals/dao/tip_119.ts
```diff
@@ -0,0 +1,116 @@
...
+  const cFeiDiff = pcvStatsAfter.userCirculatingFei.sub(pcvStatsBefore.userCirculatingFei);
+  const eqDiff = pcvStatsAfter.protocolEquity.sub(pcvStatsBefore.protocolEquity);
+  console.log(' PCV diff                               [M]e18 ', Number(pcvDiff) / 1e24);
+  console.log(' Circ FEI diff                          [M]e18 ', Number(cFeiDiff) / 1e24);
+  console.log(' Equity diff                            [M]e18 ', Number(eqDiff) / 1e24);
+  console.log('----------------------------------------------------');
+
+  // PCV Equity change should be neutral for this proposal
+  // expect(Number(eqDiff) / 1e18).to.be.at.least(1_000_000);
+  // expect(Number(eqDiff) / 1e18).to.be.at.most(3_000_000);
```
eswak: would be interesting to see this bump. What's the number ?

proposals/description/tip_119.ts
```diff
@@ -0,0 +1,30 @@
...
+      arguments: (addresses) => [addresses.gOHMHoldingPCVDeposit],
+      description: 'Add the gOHM Holding PCV Deposit'
+    }
+  ],
+  description: `
+  TIP-119: Add gOHM to Collaterisation Oracle
+
+  Register the gOHM USD oracle on the Collaterisation Oracle.
+
+  Then register the gOHM Holding PCV Deposit.
```
eswak, line 26: Simple, I like it

eswak (commented):
The proposal looks fine, maybe update the integration test but otherwise looks good to ship for me. Also if you could comment what the PCV change is due to this proposal, it would be interesting to know.

proposals/dao/tip_119.ts
thomas-waite: Should be ~$1.5M. Will uncomment now that the fix is in the other branch

proposals/dao/tip_118.ts
thomas-waite: Nice, done

contracts/test/integration/oracle/GOhmOracleIntegrationTest.t.sol
thomas-waite: I understand your point but once I start mocking the dependent contracts I'm testing it's no longer a true integration test imo and becomes much more of a unit test. Keen to test this oracle against the live mainnet contracts. 

The price bounds are there to just ensure and sense check that nothing crazy is happening. So may widen the range significantly to say `[5e17, 10e18]`

proposals/dao/tip_119.ts
thomas-waite: > would be interesting to see this bump. What's the number ?

PCV equity increase is $1.46M

eswak (dismissed):
Good to deploy & propose to TC IMO

contracts/test/integration/oracle/GOhmOracleIntegrationTest.t.sol
```diff
@@ -0,0 +1,64 @@
...
+        gOhmEthOracle = new GOhmEthOracle(MainnetAddresses.CORE, chainlinkOHMEthOracle);
+    }
+
+    function testValidgOHMPrice() public {
+        (Decimal.D256 memory gOhmEthPrice, ) = gOhmEthOracle.read();
+
+        // Eth price is ~$1000. gOHM price is ~$2400
+        // Therefore, gOHM price in ETH should be ~ (2400/1000) = 2.4 ETH
+        // This is an order of magnitude check only
+        assertGt(gOhmEthPrice.value, 5e17);
```
thomas-waite, line 44: Order of magnitude check only

contracts/test/integration/oracle/GOhmOracleIntegrationTest.t.sol
```diff
@@ -0,0 +1,64 @@
...
+        CompositeOracle gOhmUSDOracle = new CompositeOracle(
+            MainnetAddresses.CORE,
+            gOhmEthOracle,
+            chainlinkEthUSDOracleWrapper,
+            false
+        );
+        (Decimal.D256 memory gOhmUSDPrice, ) = gOhmUSDOracle.read();
+
+        // gOHM price is ~$3000
+        assertGt(gOhmUSDPrice.value / 1e18, 100);
```
thomas-waite, line 65: Order of magnitude check

protocol-configuration/mainnetAddresses.ts
```diff
@@ -1342,6 +1342,12 @@ const MainnetContractsConfig: MainnetContractsConfig = {
     category: AddressCategory.External
   },
 
+  chainlinkOHMV2EthOracle: {
+    artifactName: 'unknown',
+    address: '0x9a72298ae3886221820b1c878d12d872087d3a23',
+    category: AddressCategory.Oracle
```
xklob, line 1354: Shouldn't this be category external? IIRC we only use Oracle for our custom oracle wrapper contracts - see right below this for an example of an external-category chainlink oracle.

contracts/test/integration/oracle/GOhmOracleIntegrationTest.t.sol
```diff
@@ -0,0 +1,64 @@
...
+        assertLt(gOhmUSDPrice.value / 1e18, 10000);
+    }
+
+    /// @notice Validate that the Rari gOHM oracle matches the Fei gOHM oracle
+    function testOracleMatchesRari() public {
```
xklob, line 81: Would suggest using something other than Rari here (does Ohm themselves have an oracle we could use?)

contracts/test/integration/oracle/GOhmOracleIntegrationTest.t.sol
```diff
@@ -0,0 +1,64 @@
...
+import {Decimal, IOracle} from "../../../oracle/IOracle.sol";
+
+interface RariPriceOracle {
+    function price(address) external view returns (uint256);
+}
+
+/// @notice Integration test to validate gOHM oracle price
+contract GOhmOracleIntegrationTest is DSTest, StdLib {
+    GOhmEthOracle public gOhmEthOracle;
+
```
xklob, line 22: Lots of inlined addresses in this file - should we use a lib here?

contracts/oracle/GOhmEthOracle.sol
```diff
@@ -0,0 +1,80 @@
...
+        (uint80 roundId, , , , uint80 answeredInRound) = chainlinkOHMETHOracle.latestRoundData();
+        return answeredInRound != roundId;
+    }
+
+    /// @notice Determine the gOHM price. Report in units of ETH.
+    /// @return oracle gOHM price in units of ETH
+    /// @return true if price is valid
+    function read() external view override returns (Decimal.D256 memory, bool) {
+        (uint80 roundId, int256 ohmEthPrice, , , uint80 answeredInRound) = chainlinkOHMETHOracle.latestRoundData();
+        bool valid = !paused() && ohmEthPrice > 0 && answeredInRound == roundId;
```
xklob: Should add in a check for how recent the latest round was.

protocol-configuration/mainnetAddresses.ts
thomas-waite: Nice catch, updated to `External`

contracts/test/integration/oracle/GOhmOracleIntegrationTest.t.sol
thomas-waite: Sure, can move them to the `MainnetAddresses.sol`

contracts/test/integration/oracle/GOhmOracleIntegrationTest.t.sol
thomas-waite: Comparing against another oracle was primarily a sanity check and if this oracle matches exactly what Rari's oracle also outputs, then confident it's correct - I think this is fine as is.

Couldn't find an Olympus provided oracle on-chain (they probably just use a Chainlink oracle and then just multiply by the index) so went with Rari's

contracts/oracle/GOhmEthOracle.sol
thomas-waite: The trigger conditions for this data feed on Chainlink are a price deviation of 2%, or 86400 seconds have passed since the last update. 

This oracle is only used to value the gOHM in the PCV - it's not a significant chunk, 0.8% - so don't think we need to be more defensive and can instead rely on Chainlink

xklob: The trigger conditions specify when an update *should* happen, but don't guarantee that it will. We should be checking how old the latest update is, and mark the update as invalid if it's too old.

contracts/oracle/GOhmEthOracle.sol
```diff
@@ -0,0 +1,89 @@
...
+    AggregatorV3Interface public immutable chainlinkOHMETHOracle;
+
+    /// @notice Oracle decimals normalizer
+    uint256 public oracleDecimalsNormalizer;
+
+    /// @notice gOHM token address.
+    IgOHM public constant GOHM = IgOHM(0x0ab87046fBb341D058F17CBC4c1133F25a20a52f);
+
+    /// @notice Maximum time since last update that the Chainlink oracle is considered valid
+    uint256 public constant MAX_ORACLE_UPDATE_TIME = 86400 seconds;
```
thomas-waite: Sourced from the Chainlink data feed: https://data.chain.link/ethereum/mainnet/crypto-eth/ohmv2-eth

proposals/dao/tip_119.ts
```diff
@@ -0,0 +1,132 @@
...
+
+const fipNumber = 'tip_119';
+let pcvStatsBefore: PcvStats;
+
+// Do any deployments
+// This should exclusively include new contract deployments
+const deploy: DeployUpgradeFunc = async (deployAddress: string, addresses: NamedAddresses, logging: boolean) => {
+  ////////////// 1. Create and deploy gOHM USD oracle
+  const chainlinkOracleWrapperFactory = await ethers.getContractFactory('ChainlinkOracleWrapper');
+  const chainlinkOhmV2EthOracleWrapper = await chainlinkOracleWrapperFactory.deploy(
```
thomas-waite, line 34: Just a utility to make fetching the price in `setup()` easier - need to then use it to overwrite the Chainlink oracle so the test doesn't revert (block timestamp is from `block.txt`, days ago)

proposals/dao/tip_119.ts
```diff
@@ -0,0 +1,132 @@
...
+*/
+
+const fipNumber = 'tip_119';
+let pcvStatsBefore: PcvStats;
+
+// Do any deployments
+// This should exclusively include new contract deployments
+const deploy: DeployUpgradeFunc = async (deployAddress: string, addresses: NamedAddresses, logging: boolean) => {
+  ////////////// 1. Create and deploy gOHM USD oracle
+  const chainlinkOracleWrapperFactory = await ethers.getContractFactory('ChainlinkOracleWrapper');
```
thomas-waite, line 33: We should refactor the ChainlinkOracleWrapper to take a time parameter to check the staleness of the oracle

thomas-waite: Should probably be optional. If passed as 0, then don't check, if non-zero then check

contracts/oracle/GOhmEthOracle.sol
```diff
@@ -0,0 +1,89 @@
...
+        (uint80 roundId, int256 ohmEthPrice, , uint256 updatedAt, uint80 answeredInRound) = chainlinkOHMETHOracle
+            .latestRoundData();
+        // Valid if not paused, price is greater than 0, Chainlink price answered in this round
+        // and the Chainlink price updated happened at least within the last 24 hours.
+        bool valid = !paused() &&
+            ohmEthPrice > 0 &&
+            answeredInRound == roundId &&
+            updatedAt >= block.timestamp - MAX_ORACLE_UPDATE_TIME;
+
+        // Feth the OHM price and normalise for number of decimals
```
xklob: Nit fetch not feth (lol)

contracts/test/integration/fixtures/MainnetAddresses.sol
```diff
@@ -35,6 +35,11 @@ library MainnetAddresses {
     address public constant DAI = 0x6B175474E89094C44Da98b954EedeAC495271d0F;
     address public constant AGEUR = 0x1a7e4e63778B4f12a199C062f3eFdD288afCBce8;
     address public constant USDC = 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48;
+    address public constant GOHM = 0x0ab87046fBb341D058F17CBC4c1133F25a20a52f;
     address public constant DAI_FIXED_PRICE_PSM = 0x2A188F9EB761F70ECEa083bA6c2A40145078dfc2;
     address public constant LUSD_HOLDING_PCV_DEPOSIT = 0x4378De2F2991Fbed6616b34AC7727E7653713712;
+
+    address public constant RARI_ORACLE = 0x1887118E49e0F4A78Bd71B792a49dE03504A764D;
```
xklob, line 42: Should probably be more specific here. Suggest RARI_GOHM_ORACLE

thomas-waite: That's not a specific GOHM oracle, but a Rari oracle. They have an aggregator/master oracle which you then pass an asset address into to get a specific price feed

xklob (approved):

PR #952, opened 2022-07-04
[TIP-120] DAO Vote: PCVGuardian v3

### LINKED ISSUE #954: PCVGuardianV3 with ratio movements

eswak:
A new version of the PCV Guardian : 

- Adds `**Ratio` functions to allow to move percentages with `withdraw`, `withdrawETH`, and `withdrawERC20`
- Rewrite all tests in Forge
- Factorize the code that is repeated in all withdraw functions inside a modifier (access control, check if target is a safe address, unpause source deposit if paused, deposit after & pause after boolean parameters handling)
- Changed the default behavior : if the source deposit is paused before the call, re-pause it after withdrawal even if pauseAfter = false
- Very small gas optimization by changing order of role checks because most of the time, TC Timelock (pcv safe mover role) or Guardian (guardian role) will perform the calls, and not the DAO Timelock (governor role), and the order of checks can save some external calls to `core.hasRole(role, address)`.

- [x] All Tests Passing


contracts/pcv/IPCVGuardian.sol
```diff
@@ -64,7 +64,21 @@ interface IPCVGuardian {
         bool depositAfter
     ) external;
 
-    /// @notice governor-or-guardian-only method to withdraw funds from a pcv deposit, by calling the withdraw() method on it
+    /// @notice withdraw funds from a pcv deposit, by calling the withdraw() method on it
```
thomas-waite: Would update the comment to indicate that it allows %s to be moved

contracts/pcv/IPCVGuardian.sol
```diff
@@ -78,7 +92,21 @@ interface IPCVGuardian {
         bool depositAfter
     ) external;
 
-    /// @notice governor-or-guardian-only method to withdraw funds from a pcv deposit, by calling the withdraw() method on it
+    /// @notice withdraw funds from a pcv deposit, by calling the withdrawETH() method on it
```
thomas-waite: Same thing about a comment/couple of words to indicate it's a ratio or %

contracts/pcv/PCVGuardian.sol
```diff
@@ -41,7 +44,7 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
     function setSafeAddress(address pcvDeposit)
         external
         override
-        hasAnyOfTwoRoles(TribeRoles.GOVERNOR, TribeRoles.PCV_GUARDIAN_ADMIN)
+        hasAnyOfTwoRoles(TribeRoles.PCV_GUARDIAN_ADMIN, TribeRoles.GOVERNOR)
```
thomas-waite, line 47: Really neat observation and gas optimisation around the order of calling this and how often different entities will call

contracts/pcv/PCVGuardian.sol
```diff
@@ -76,45 +79,92 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
...
         address pcvDeposit,
         address safeAddress,
-        uint256 amount,
         bool pauseAfter,
         bool depositAfter
-    ) external override hasAnyOfThreeRoles(TribeRoles.GOVERNOR, TribeRoles.PCV_SAFE_MOVER_ROLE, TribeRoles.GUARDIAN) {
-        require(isSafeAddress(safeAddress), "Provided address is not a safe address!");
+    ) {
+        {
+            // scoped in this modifier to prevent stack to deep errors & enforce consitent acl
```
thomas-waite: Nit: `consitent` -> `consistent`

contracts/pcv/PCVGuardian.sol
```diff
@@ -76,45 +79,92 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
...
-    /// @param amount the amount to withdraw
-    /// @param pauseAfter if true, the pcv contract will be paused after the withdraw
-    /// @param depositAfter if true, attempts to deposit to the target PCV deposit
-    function withdrawToSafeAddress(
+    /// @notice modifier to factorize the logic in all withdrawals :
+    /// - first, ensure the deposit to withdraw from is unpaused
+    /// - second, perform withdrawal
+    /// - third, re-pause deposit if it was paused or if pauseAfter = true
+    /// - finally, call pcvDeposit.deposit() if depositAfter = true
+    modifier beforeAndAfterWithdrawal(
```
thomas-waite: Love this refactor

contracts/pcv/PCVGuardian.sol
```diff
@@ -76,45 +79,92 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
...
+        bool paused = pcvDeposit._paused();
+        if (paused) {
+            pcvDeposit._unpause();
+        }
 
-        IPCVDeposit(pcvDeposit).withdraw(safeAddress, amount);
+        _;
 
-        if (pauseAfter) {
+        if (paused || pauseAfter) {
```
thomas-waite: Much prefer this also. What we had before had this horrible side effect of changing pause state unintentionally

contracts/test/utils/Fixtures.sol
```diff
@@ -16,6 +17,8 @@ struct FeiTestAddresses {
     address minterAddress;
     address burnerAddress;
     address guardianAddress;
+    address pcvGuardianAdminAddress;
```
thomas-waite: Not down to you, but I don't see why we need to append the string `Address` onto these names. Probably cleaner to just do `address minter`, `address pcvGuardianAdmin` etc

contracts/test/unit/PCVGuardian.sol
```diff
@@ -0,0 +1,491 @@
...
+        vm.label(address(core), "core");
+        vm.label(address(token), "token");
+        vm.label(address(pcvDeposit1), "pcvDeposit1");
+        vm.label(address(pcvDeposit2), "pcvDeposit2");
+        vm.label(address(pcvDeposit3), "pcvDeposit3");
+        vm.label(address(pcvGuardian), "pcvGuardian");
+    }
+
+    // should have no safe addresses upon deployment when deployed with no safe addresses
+    function testDeploy1() public {
```
thomas-waite: I would make this test name more instructive. Maybe `testDeployNoSafeAddresses` and then for the next test, `testDeployWithSafeAddresses`

contracts/test/unit/PCVGuardian.sol
```diff
@@ -0,0 +1,491 @@
...
+        address[] memory getSafeAddresses = pcvGuardianWithoutAddresses.getSafeAddresses();
+        assertEq(getSafeAddresses.length, 1);
+        assertEq(getSafeAddresses[0], address(0x42));
+    }
+
+    // access control and state changes
+
+    // should revert when calling setSafeAddress & setSafeAddresses from a non-privileged address
+    // should revert when calling unsetSafeAddress & unsetSafeAddresses from a non-privileged address
+    function testAccessControl1() public {
```
thomas-waite: Would give this a more illustrative test name

contracts/test/unit/PCVGuardian.sol
```diff
@@ -0,0 +1,491 @@
...
+        address safeAddress = address(0x42);
+        address[] memory safeAddresses = new address[](1);
+        safeAddresses[0] = safeAddress;
+
+        // none of the following calls shall revert if access
+        // control is properly checked for the various callers
+        vm.startPrank(addresses.pcvGuardianAdminAddress);
+        pcvGuardian.setSafeAddress(safeAddress);
+        pcvGuardian.unsetSafeAddress(safeAddress);
+        pcvGuardian.setSafeAddresses(safeAddresses);
```
thomas-waite, line 110: Really like how detailed this test is

contracts/test/unit/PCVGuardian.sol
```diff
@@ -0,0 +1,491 @@
...
+            false,
+            false
+        );
+        pcvDeposit1.deposit();
+        pcvDeposit2.deposit();
+        vm.stopPrank();
+    }
+
+    // check state changes & getters after set & unset
+    function testStateConsistency1() public {
```
thomas-waite: Would definitely rename this test name. What state consistency are we checking?

thomas-waite: I would also split this unit test out into several. Each unit test should test as close to just one function as possible imo

contracts/test/unit/PCVGuardian.sol
```diff
@@ -0,0 +1,491 @@
...
+
+        pcvGuardian.setSafeAddress(safeAddress);
+
+        vm.expectRevert("PCVGuardian: already a safe address");
+        pcvGuardian.setSafeAddress(safeAddress);
+
+        vm.stopPrank();
+    }
+
+    // withdrawals
```
thomas-waite: Nit: Personal preference but I'd make this delimiter more obvious like: `///////////// WITHDRAWALS //////////`

contracts/test/unit/PCVGuardian.sol
```diff
@@ -0,0 +1,491 @@
...
+        vm.stopPrank();
+    }
+
+    // should withdraw from a pcv deposit
+    // should withdrawETH from a pcv deposit
+    // should withdrawERC20 from a pcv deposit
+    // should withdrawRatio from a pcv deposit
+    // should withdrawETHRatio from a pcv deposit
+    // should withdrawERC20Ratio from a pcv deposit
+    function testWithdrawals() public {
```
thomas-waite: This is far too much for one unit test imo. Each of these withdraw functions should be broken out into their own unit test

thomas-waite (commented):
Really like these changes overall, will be a big quality of life improvement. 

Have some feedback on the unit tests - in general I think they should be broken up into many more and each major function should be covered by its own test. Also think the test names could be more descriptive/illustrative

proposals/description/pcv_guardian_v3.ts
```diff
@@ -0,0 +1,53 @@
...
+      method: 'revokeRole(bytes32,address)',
+      arguments: (addresses) => [ethers.utils.id('GUARDIAN_ROLE'), addresses.pcvGuardianV2],
+      description: 'Revoke GUARDIAN_ROLE from pcvGuardianV2'
+    },
+    {
+      target: 'core',
+      values: '0',
+      method: 'grantRole(bytes32,address)',
+      arguments: (addresses) => [ethers.utils.id('PCV_CONTROLLER_ROLE'), addresses.pcvGuardian],
+      description: 'Grant PCV_CONTROLLER_ROLE from pcvGuardianV3'
```
xklob: Nit: to

proposals/description/pcv_guardian_v3.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+    {
+      target: 'core',
+      values: '0',
+      method: 'grantRole(bytes32,address)',
+      arguments: (addresses) => [ethers.utils.id('GUARDIAN_ROLE'), addresses.pcvGuardian],
+      description: 'Grant GUARDIAN_ROLE from pcvGuardianV3'
```
xklob: Nit: to

proposals/description/pcv_guardian_v3.ts
```diff
@@ -0,0 +1,53 @@
...
+      values: '0',
+      method: 'slay(address)',
+      arguments: (addresses) => [addresses.fuseWithdrawalGuardV1],
+      description: 'Slay old fuseWithdrawalGuardV1 guard'
+    },
+    {
+      target: 'pcvSentinel',
+      values: '0',
+      method: 'knight(address)',
+      arguments: (addresses) => [addresses.fuseWithdrawalGuard],
```
xklob, line 46: Really confusing that this one has no version number, and the old has a v1 version number. Thoughts on just using v2 for the new version number?

proposals/description/pcv_guardian_v3.ts
```diff
@@ -0,0 +1,53 @@
...
+      description: 'Knight new fuseWithdrawalGuard guard'
+    }
+  ],
+  description: `TODO`
```
xklob: Update description

contracts/test/integration/sentinel/FuseWithdrawalGuard.t.sol
```diff
@@ -54,7 +54,9 @@ contract FuseWithdrawalGuardIntegrationTest is DSTest, StdLib {
         sentinel.knight(address(guard));
     }
 
-    function testGuardCanProtec() public {
+    // TODO: update Fuse withdrawal guard PCV_GUARDIAN address,
```
xklob: Commenting for this todo as a reminder.

contracts/sentinel/guards/FuseWithdrawalGuard.sol
```diff
@@ -22,7 +22,7 @@ contract FuseWithdrawalGuard is IGuard, CoreRef {
     EnumerableSet.AddressSet private fuseDeposits;
 
     /// @notice the PCV mover contract exposed to guardian role
-    PCVGuardian public constant pcvGuardian = PCVGuardian(0x02435948F84d7465FB71dE45ABa6098Fc6eC2993);
+    PCVGuardian public constant pcvGuardian = PCVGuardian(0x0000000000000000000000000000000000000000); // TODO update
```
xklob: Commenting to remind of todo.

contracts/pcv/PCVGuardian.sol
```diff
@@ -76,114 +79,165 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
...
+    ) external override beforeAndAfterWithdrawal(pcvDeposit, safeAddress, depositAfter) {
         IPCVDeposit(pcvDeposit).withdrawERC20(token, safeAddress, amount);
+        emit PCVGuardianERC20Withdrawal(pcvDeposit, safeAddress, token, amount);
+    }
 
-        if (pauseAfter) {
-            pcvDeposit._pause();
-        }
-
-        if (depositAfter) {
```
xklob, line 171: I think we agreed to leave depositAfter in?

contracts/pcv/PCVGuardian.sol
```diff
@@ -76,114 +79,165 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
...
         address pcvDeposit,
         address safeAddress,
         address token,
         uint256 amount,
-        bool pauseAfter,
         bool depositAfter
-    ) external override hasAnyOfThreeRoles(TribeRoles.GOVERNOR, TribeRoles.PCV_SAFE_MOVER_ROLE, TribeRoles.GUARDIAN) {
-        require(isSafeAddress(safeAddress), "Provided address is not a safe address!");
-
-        pcvDeposit._ensureUnpaused();
```
xklob, line 163: Do we not need this to withdraw from paused deposits?

contracts/pcv/PCVGuardian.sol
```diff
@@ -76,114 +79,165 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
...
+                    _core.hasRole(TribeRoles.PCV_SAFE_MOVER_ROLE, msg.sender) ||
+                    _core.hasRole(TribeRoles.GOVERNOR, msg.sender),
+                "UNAUTHORIZED"
+            );
+            require(isSafeAddress(safeAddress), "PCVGuardian: address not whitelisted");
+        }
 
-        pcvDeposit._ensureUnpaused();
+        bool paused = pcvDeposit._paused();
+        if (paused) {
```
xklob: Why not just use ensureUnpaused here?

proposals/description/pcv_guardian_v3.ts
eswak: we argued about this with @thomas-waite in the dev chat several days ago, the most up to date version doesn't have a version number, otherwise you don't know if there is a V3 you should use instead, when you use the V2

contracts/test/integration/sentinel/FuseWithdrawalGuard.t.sol
eswak: thanks can only be fixed after deployment

contracts/sentinel/guards/FuseWithdrawalGuard.sol
eswak: thanks can only be fixed after deployment

contracts/pcv/PCVGuardian.sol
eswak: yes we kept `depositAfter`, the diff is displaying very weird 🤔 this logic is now in the modifier

contracts/pcv/PCVGuardian.sol
eswak: we do, that's why before withdrawals the modifier reads the paused state & unpauses if the deposit is paused
image

contracts/pcv/PCVGuardian.sol
eswak: `ensureUnpaused` reads the `pcvDeposit.paused()` but does not return it. We need to read the paused state before withdrawal, because after withdrawal we want to re-pause if the deposit was paused in the first place. But yeah essentially it's the same code than what's in the `ensureUnpaused` function

eswak:
I moved the modifier function to be at the bottom of the file, the diff displays much more nicely now

contracts/pcv/PCVGuardian.sol
```diff
@@ -76,114 +79,167 @@ contract PCVGuardian is IPCVGuardian, CoreRef {
...
+            require(isSafeAddress(safeAddress), "PCVGuardian: address not whitelisted");
+        }
+
+        bool paused = pcvDeposit._paused();
+        if (paused) {
+            pcvDeposit._unpause();
+        }
 
-        if (pauseAfter) {
+        _;
```
thomas-waite, line 223: No strong feelings and fine as is, but not a fan of having the control flow jumping about in this modifier. The modifier runs before the function executes and checks permissions etc., then passes the control flow back to the function on line 223, then will take the control flow back once the function is done.

Might be cleaner and a nicer seperation of `before` and `after` logic if this modifier was split out into two hooks: `_beforeWithdrawal()` and `_afterWithdrawal()` which then get called in the function body

contracts/sentinel/guards/FuseWithdrawalGuard.sol
thomas-waite: This guardian address should be injected in at constructor time no? Otherwise I imagine it's impossible to test this

contracts/test/integration/sentinel/FuseWithdrawalGuard.t.sol.disabled
```diff
@@ -54,7 +54,9 @@ contract FuseWithdrawalGuardIntegrationTest is DSTest, StdLib {
         sentinel.knight(address(guard));
     }
 
-    function testGuardCanProtec() public {
+    // TODO: update Fuse withdrawal guard PCV_GUARDIAN address,
```
thomas-waite: I think the guardian should be deployed locally in this integration test and then it's address be injected into the `FuseWithdrawalGuard` through it's constructor, otherwise you can't test the guard without deploying something to mainnet first

protocol-configuration/proposalsConfig.ts
```diff
@@ -1,13 +1,21 @@
...
 
 export const ProposalsConfig: TemplatedProposalsConfigMap = {
+  pcv_guardian_v3: {
+    deploy: true,
+    totalValue: 0,
+    proposal: pcv_guardian_v3,
+    proposalId: '',
+    affectedContractSignoff: [],
+    deprecatedContractSignoff: [],
+    category: ProposalCategory.DEBUG
```
thomas-waite: Will want to change this proposal category to TC I imagine?

proposals/description/pcv_guardian_v3.ts
```diff
@@ -0,0 +1,56 @@
+import { ethers } from 'hardhat';
+import { TemplatedProposalDescription } from '@custom-types/types';
+
+const proposal: TemplatedProposalDescription = {
+  title: 'TIP-XYZ: PCV Guardian v3',
```
thomas-waite: Will need a TIP number

proposals/dao/pcv_guardian_v3.ts
```diff
@@ -0,0 +1,131 @@
...
+// Tears down any changes made in setup() that need to be
+// cleaned up before doing any validation checks.
+const teardown: TeardownUpgradeFunc = async (addresses, oldContracts, contracts, logging) => {
+  console.log(`No actions to complete in teardown for fip${fipNumber}`);
+};
+
+// Run any validations required on the fip using mocha or console logging
+// IE check balances, check state of contracts, etc.
+const validate: ValidateUpgradeFunc = async (addresses, oldContracts, contracts, logging) => {
+  // display pcvStats
```
thomas-waite, line 128: Being paranoid I would probably make sure that we can move funds here using the newly deployed pcvGuardianV3

thomas-waite (commented):
Looks good overall to me. 

Idea about using `_beforeWithdrawal()` and `_afterWithdrawal()` hooks instead of a modifier to reduce control flow switching, but no strong feelings. 

Also think the `FuseWithdrawalGuard` should have the `PCVGuardian` address injected in through it's constructor, otherwise it's impossible to test new pcvGuardians with it before deploying them.

contracts/test/integration/sentinel/FuseWithdrawalGuard.t.sol.disabled
eswak: good idea

protocol-configuration/proposalsConfig.ts
eswak: yep, good catch

proposals/description/pcv_guardian_v3.ts
eswak: I'll take... TIP-120 !

proposals/dao/pcv_guardian_v3.ts
eswak: ok

contracts/pcv/PCVGuardian.sol
eswak: I think the modifier is better because we want the paused stated read before withdrawal to be used after withdrawal (to re-pause), we can't do that without modifier

protocol-configuration/proposalsConfig.ts
eswak: actually has to be submitted to DAO because it revokes & grant Guardian & PCV Controller

PR #956, opened 2022-07-05
Version pcvGuardian, add safe addresses e2e test

thomas-waite:
# Version pcvGuardian, add safe addresses e2e test

Use a versioning system for the pcvGuardian to avoid mistakes in the future in using the old, deprecated guardian. 

Also add an e2e test and config asserting the state of the safe addresses in the latest active version of the pcvGuardian

### DIFF
--- a/test/integration/tests/safeAddresses.ts
+++ b/test/integration/tests/safeAddresses.ts
@@ -1,0 +1,62 @@
+import { NamedContracts } from '@custom-types/types';
+import { safeAddressesConfig } from '@protocol/safeAddresses';
+import proposals from '@protocol/proposalsConfig';
+import { TestEndtoEndCoordinator } from '@test/integration/setup';
+import chai, { expect } from 'chai';
+import CBN from 'chai-bn';
+import { solidity } from 'ethereum-waffle';
+import { ethers } from 'hardhat';
+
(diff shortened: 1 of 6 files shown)


test/integration/tests/safeAddresses.ts
```diff
@@ -0,0 +1,62 @@
...
+    expect(numAddressesOnChain).to.be.equal(numAddressesLocally, 'Num safe addresses mismatch');
+
+    for (let i = 0; i < safeAddressesConfig.length; i += 1) {
+      const safeAddressContractName = safeAddressesConfig[i];
+
+      // Get the on-chain contract
+      const onChainContract = contracts[safeAddressContractName];
+
+      // Validate is a safe address
+      expect(await pcvGuardian.isSafeAddress(onChainContract.address)).to.be.true;
```
eswak:
```suggestion
      expect(await pcvGuardian.isSafeAddress(onChainContract.address)).to.be.equal(
        true,
        'Expected ' + safeAddressContractName + ' to be a safe address'
      );
```

eswak (changes requested):
thanks for adding

`permissions.ts` probably need an update too

eswak (approved):
yayyy

PR #957, opened 2022-07-05
rename fixture addresses & fix podexecutor unit tests

eswak:
Rename fixture addresses for forge unit tests

Fix forge unit tests of the PodExecutor ([this check](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/governance/TimelockController.sol#L142) needs a block timestamp of [at least 1](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/governance/TimelockController.sol#L31))


xklob (approved):

thomas-waite (approved):
Looks good to me. 

I don't follow why the block needs fast forwarding in the `setUp()` of the `PodExecutor` unit tests - where I need to test execution, I warp forward by one block so execution is available. In the other two unit tests, I'm just testing for a specific revert message which I get. Where they failing for some reason?

PR #960, opened 2022-07-07
Fix (most) forge compilation warnings

eswak:
This PR fixes most compilation warnings when running `npm run compile` (forge compilation).

Only .sol changes that won't change the contract's bytecode are included.

Most warnings were in Mock contracts or forge test files.

There are a bunch of warnings left during compilation, detailed below : 

- 12 warnings in `node_modules/@orcaprotocol/contracts/....`, not much we can do about it without copying the Solidity in our own repo, so I did not take this decision. Tom notified Orca about it (but it's not a big deal).

- 5 warnings in `WETH9.sol`, it's using an old version of Solidity and uses `this.balance` instead of `address(this).balance` and does not prefix event emissions with the `emit` keyword, but I didn't change because this contract compiles in Solidity 0.4.18

- 1 warning in `PSMRouter.sol` "This contract has a payable fallback function, but no receive ether function. Consider adding a receive ether function.", I did not fix because it would change the bytecode of the contract. PSMRouter is not used anywhere and the only mainnet deployment is deprecated because it's a wrapper above the ETH PSM that is itself deprecated (I'm doing a separate PR to deprecate the `ethPSMRouter`)

- 2 warnings in `FuseGuardian.sol`, I think these are important warnings because the Solidity code does not have the intended behavior and I think this was missed because of all the "warnings pollution" and nobody reading warnings anymore. The functions `_setBorrowPaused` and `_setMintPausedByUnderlying` return booleans, but they always return false, instead of the result of the internal functions. I don't think there's any harm in it, and I did not fix the warnings because it would introduce bytecode changes, but it's definitely not the intended behavior. Reducing the amounts of warnings will allow us to catch this kind of errors in the future.

### DIFF
--- a/contracts/mock/MockChainlinkOracle.sol
+++ b/contracts/mock/MockChainlinkOracle.sol
@@ -14,9 +14,9 @@ contract MockChainlinkOracle is AggregatorV3Interface {
     uint256 _updatedAt;
     uint80 _answeredInRound;
 
-    constructor(int256 value, uint8 decimals) {
+    constructor(int256 value, uint8 oracleDecimals) {
         _value = value;
-        _decimals = decimals;
+        _decimals = oracleDecimals;
         _roundId = 42;
         _startedAt = 1620651856;
         _updatedAt = 1620651856;
--- a/contracts/mock/MockConvexBaseRewardPool.sol
+++ b/contracts/mock/MockConvexBaseRewardPool.sol
@@ -31,7 +31,10 @@ contract MockConvexBaseRewardPool is MockERC20 {
         getReward(msg.sender, claim);
     }
 
-    function getReward(address who, bool claim) public returns (bool) {
+    function getReward(
+        address who,
+        bool /* claim*/
+    ) public returns (bool) {
         if (rewardAmountPerClaim > 0) {
             rewardToken.mint(who, rewardAmountPerClaim);
         }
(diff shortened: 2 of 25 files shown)


contracts/pcv/balancer/BPTLens.sol
```diff
@@ -72,7 +76,12 @@ contract BPTLens is IPCVDepositBalances {
     }
 
     function balance() public view override returns (uint256) {
-        (IERC20[] memory _tokens, uint256[] memory balances, ) = VAULT.getPoolTokens(id);
+        (
```
xklob: This seems to change something significant - in the previous version, tokens and balances were assigned from indices 0 and 1 from the result tuple, and in the new version, they're assigned from indices 1 and 2 - am I reading this right?

eswak: 😨 yes

eswak: prettier formatted it in a **very** weird way but I think we're good, the code change is :  (x, y, ) = ... -> ( , y, )

xklob: Ahh yep I see now. Sounds good!

eswak:
Rebased to remove commits that fixed compilation warnings that were on live contracts.

Only Mock & Test contracts are fixed here.

thomas-waite (approved):
Looks great, thanks for fixing these! I've also given Orca the heads up about their warnings, they're taking a look

PR #961, opened 2022-07-07
Deprecate ethPSMRouter (the only PSMRouter)

eswak:
Following the deprecation of the `ethPSM` in TIP-118 (#922), the `ethPSMRouter` can also be deprecated.

The `ethPSMRouter` was the only instance of `PSMRouter` contract deployed, so I also deprecated its hardhat test suite.


thomas-waite (approved):
LGTM, thanks

PR #965, opened 2022-07-09
Genericize Protocol State Test

xklob:
# PR Type: Feature

Converts the specific psm state tests into a generic protocol state test.
We can add any contracts and contract-state we want into this config, and the test will check to make sure that mainnet hasn't deviated from the expected values.

- [x] All Tests Passing


protocol-configuration/mainnetAddresses.ts
```diff
@@ -1,6 +1,6 @@
-import { MainnetContractsConfig, AddressCategory } from '../types/types'; // imported without custom path to allow docs to autogen without ts errors
+import { AddressCategory } from '../types/types'; // imported without custom path to allow docs to autogen without ts errors
 
-const MainnetContractsConfig: MainnetContractsConfig = {
+export const MainnetContractsConfigs = {
```
xklob: Converted this to a named export from a default export - generally better pattern unless there's a specific reason to use a default export.

protocol-configuration/mainnetAddresses.ts
```diff
@@ -2528,4 +2528,6 @@ const MainnetContractsConfig: MainnetContractsConfig = {
   }
 };
 
-export default MainnetContractsConfig;
+export type MainnetContractsType = typeof MainnetContractsConfigs;
+export type MainnetContractsEntryName = keyof MainnetContractsType;
+export const MainnetContractsEntryNames = Object.keys(MainnetContractsConfigs) as MainnetContractsEntryName[];
```
Joeysantoro: this is cool

test/integration/setup/loadContracts.ts
```diff
@@ -1,9 +1,9 @@
-import mainnetAddresses from '@protocol/mainnetAddresses';
+import { MainnetContractsConfigs } from '@protocol/mainnetAddresses';
 import { ethers } from 'hardhat';
 import { MainnetContracts, MainnetContractsConfig, NamedAddresses } from '@custom-types/types';
 
 export async function getAllContracts(): Promise<MainnetContracts> {
-  const addresses = mainnetAddresses as MainnetContractsConfig;
+  const addresses = MainnetContractsConfigs as MainnetContractsConfig;
```
Joeysantoro: case is a bit confusing because the object and the type are both CamelCase with only a "s" difference

test/integration/tests/pauseState.ts
```diff
@@ -20,50 +19,36 @@ describe('e2e-pause-state', function () {
...
-      // Validate PSM related pausing
-      if (expectedContractPauseConfig.redeemPaused) {
-        expect(await onChainContract.redeemPaused()).to.be.equal(
-          expectedContractPauseConfig.redeemPaused,
-          'redeemPaused() state incorrect'
-        );
-      }
+      for (const keyName of Object.keys(stateConfigEntry)) {
+        const expectedValue = stateConfigEntry[keyName as keyof typeof stateConfigEntry];
+        const actualValue = await contract[keyName]();
```
Joeysantoro: would this work for non boolean types?

xklob: It should, but I'll add in some test cases against string and a number/big number types just to make sure it does.

test/integration/setup/loadContracts.ts
xklob: Good callout, I'll make this more clear.

test/integration/tests/pauseState.ts
eswak: would be cool to allow addresses as expected values, for instance `'{pcvGuardian}'` instead of the `'0x.....'` value

test/integration/tests/pauseState.ts
```diff
@@ -20,50 +19,36 @@ describe('e2e-pause-state', function () {
...
 
-      if (expectedContractPauseConfig.mintPaused) {
-        expect(await onChainContract.mintPaused()).to.be.equal(
-          expectedContractPauseConfig.mintPaused,
-          'mintPaused() state incorrect'
+        expect(actualValue).to.be.equal(
+          expectedValue,
+          `Expected ${contractName} to have ${keyName} set to ${expectedValue}, got ${actualValue}`
```
eswak: thanks for the custom error message 👍

test/integration/setup/loadContracts.ts
xklob: Removed MainnetContratsConfig interface and now just using the type "MainnetContractsType"

test/integration/tests/pauseState.ts
xklob: > would be cool to allow addresses as expected values, for instance `'{pcvGuardian}'` instead of the `'0x.....'` value

You can actually do this by just using a var for any value in any of the key-value pairs in the state config file.

PR #966, opened 2022-07-09
Cleanup: General Cleanup

xklob:
# PR Type: Feature

General cleanup tasks, also removes a couple of deprecated things.
Adds comments to many of the protocol configuration files explaining how they work.
Simplifies a few of the protocol configuration tests.
Changes some of the default exports to named exports (just a better pattern unless default exports are explicitly needed)
Removes the TribalChief E2E

- [x] All Tests Passing


protocol-configuration/optimisticGovernance.ts
```diff
@@ -56,4 +36,23 @@ export const tribeCouncilPodConfig: PodConfig = {
   placeHolderMembers: placeHolderCouncilMembers
 };
 
-export const TRIBAL_COUNCIL_POD_ID = 25;
+export type PodConfig = {
```
thomas-waite, line 39: Worthwhile turning this into an interface rather than a type? I like the suggestion for objects

test/integration/tests/tribalCouncil.ts
```diff
@@ -1,161 +0,0 @@
-import { Core, PodAdminGateway, PodFactory, RoleBastion } from '@custom-types/contracts';
-import { NamedAddresses, NamedContracts } from '@custom-types/types';
```
thomas-waite, line 2: Out of curiosity, did you not need to reduce the number of concurrent CI workers/parallelism when you removed this test? 

I've found before that when there are more workers than e2e tests, then they kind of "wrap around" and the spare worker starts to run the whole e2e test suite individually, taking forever

thomas-waite (dismissed):
Looks great, thanks for doing this!

test/integration/tests/tribalCouncil.ts
xklob: Perhaps that's why the tests took so long. My plan was to adjust CI after these three cleanup prs get merged in.

protocol-configuration/optimisticGovernance.ts
xklob: Probably! I'm going to take a pass over all our types and interfaces in the next pr.

PR #967, opened 2022-07-12
Proposal Simulator Fix + Other Speedups

xklob:
# PR Type: Feature/Bugfix

Replaces the _simulate function of the Idle Finance Compound-Alpha proposal simulator which fixes a bug related to automining being off, and simultaneously `await`-ing a transaction result. Note this is a workaround fix and not meant to be long-term. But it's much faster than before at least!

Deletes the longest-running unit tests which are deprecated anyways, which was the TribalChief tests.

Changes the PriceBoundPegStabilityModule tests a little bit to be faster, too, since they were the longest running tests by far after that.

Up paralleism of units specifically to lower overall runtime.

Total CI time now down to 4m 6s.

- [x] All Tests Passing


test/unit/peg/PriceBoundPegStabilityModule.test.ts
```diff
@@ -1056,74 +1066,6 @@ describe('PriceBoundPegStabilityModule', function () {
...
         await psm.connect(impersonatedSigners[pcvControllerAddress]).withdraw(userAddress, await psm.balance());
-
-        const endingBalance = await psm.balance();
-        expect(endingBalance).to.be.equal(0);
-        expect(await asset.balanceOf(userAddress)).to.be.equal(amount);
-      });
-    });
-  });
-
-  describe('PCV', function () {
```
xklob, line 1067: The primary way I got speedups in this test file were by changing beforeEach to before; since these pcv tests are very dependent on having independent state, it broke this section. This is intended as a temp fix as we're porting over to forge, however, so I think that's probably okay. Can redo this section if we think I should.

PR #986, opened 2022-07-19
Curve DAI->LUSD swapper + Guard

eswak:
This PR adds 2 new contracts : a CurveSwapper and a CurveSwapperGuard. The CurveSwapper can be used to swap a token for another using a Curve stableswap meta-pool, and the Guard can be used to trigger it permissionlessly under certain circumstances (using Guardian privileges in a hard-coded way).

- [x] All Tests Passing
- [x] Add forge tests for the new contracts
- [x] Proposal Added to ProposalsConfig
- [x] Fork Block Correct
- [ ] Deploy Contract If Needed (+ Verify on Etherscan & Add to MainnetAddresses)
- [x] Update Documentation If Needed
- [x] Update Roles Config
- [x] Update Collateralization Oracle If Needed


contracts/pcv/curve/CurveSwapper.sol
```diff
@@ -0,0 +1,195 @@
...
+    function setReceivingAddress(address newTokenReceivingAddress)
+        external
+        override
+        hasAnyOfThreeRoles(TribeRoles.SWAP_ADMIN_ROLE, TribeRoles.PCV_CONTROLLER, TribeRoles.GOVERNOR)
+    {
+        _setReceivingAddress(newTokenReceivingAddress);
+    }
+
+    function _setReceivingAddress(address newTokenReceivingAddress) internal {
+        require(newTokenReceivingAddress != address(0), "CurveSwapper: zero address");
```
thomas-waite: I don't think there's need for `!= address(0)` checks anymore in Solidity. One of the reasons it was introduced was because Solidity didn't used to verify the length of `msg.data`, so you could accidentally call functions with no args: https://forum.openzeppelin.com/t/removing-address-0x0-checks-from-openzeppelin-contracts/2222 

Why not also check `address(1)` etc

contracts/pcv/curve/CurveSwapper.sol
```diff
@@ -0,0 +1,195 @@
...
+        _setMaximumSlippage(_maxSlippageBps);
+
+        // find the i and j index for underlying swap, where
+        // i = _tokenSpent, and j = _tokenReceived
+        address coins0 = ICurvePool(_curvePool).coins(0);
+        address coins1 = ICurvePool(_curvePool).coins(1);
+        address _basePool;
+        int128 _i;
+        int128 _j;
+        if (_tokenSpent == coins0) {
```
thomas-waite: Not a fan of this logic here, it's hard to follow and feels a spot for mistakes. I'm not sure what would be an improvement though - is there a well tested Curve library we can import that will sort for us?

Overall, I like the constructor API of just being able to pass a `tokenSpent` and `tokenReceived` and all index calculations are abstracted away; but wondering if there's a cleaner/nicer way of getting them

contracts/test/integration/pcv/CurveSwapper.t.sol
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    // Check deployed contract & getters for public properties
+    function testInit() public {
+        assertEq(curveSwapper.curvePool(), MainnetAddresses.CURVE_LUSD_METAPOOL);
+        assertEq(curveSwapper.tokenSpent(), MainnetAddresses.LUSD);
+        assertEq(curveSwapper.tokenReceived(), MainnetAddresses.DAI);
+        assertEq(curveSwapper.maxSlippageBps(), 10000);
+        assertEq(curveSwapper.tokenReceivingAddress(), address(this));
+        assertEq(curveSwapper.i(), 0);
```
thomas-waite, line 61: Are we confident in the index calculation logic and we don't need to test the calculation for different assets and pools? Really like the tests overall

proposals/dao/tip_119.ts
```diff
@@ -35,34 +35,64 @@ let initialCompoundDAIBalance: BigNumber;
...
+  // await gOhmEthOracle.deployed();
+
+  // logging && console.log(`Deployed gOHM Eth Oracle at ${gOhmEthOracle.address}`);
+
+  // // Create the gOHM USD oracle
+  // const CompositeOracleFactory = await ethers.getContractFactory('CompositeOracle');
+  // const gOhmUSDOracle = await CompositeOracleFactory.deploy(
+  //   addresses.core,
+  //   gOhmEthOracle.address,
+  //   addresses.chainlinkEthUsdOracleWrapper,
```
thomas-waite: Commented out as these are already deployed right, but you need to set the deploy flag to true to get the curveSwapper deployed

contracts/sentinel/guards/CurveSwapperGuard.sol
```diff
@@ -0,0 +1,83 @@
...
+/// @notice Guard that can trigger a Curve swapper under certain conditions.
+contract CurveSwapperGuard is IGuard, CoreRef {
+    /// @notice the PCV mover contract exposed to guardian role
+    address public immutable pcvGuardian;
+    /// @notice the Curve swapper to call
+    address public immutable curveSwapper;
+    /// @notice the source deposit to withdraw from
+    address public immutable sourceDeposit;
+    /// @notice the PCV Deposit to read balance from
+    address public immutable inspectedDeposit;
```
thomas-waite: Would add more detailed explanation either on the `inspectedDeposit` or in a `@dev` tag on the contract why there is a seperate variable `inspectedDeposit` to `sourceDeposit`

contracts/sentinel/guards/CurveSwapperGuard.sol
```diff
@@ -0,0 +1,83 @@
...
+    /// @notice the minimum balance under which a swap can be triggered
+    uint256 public immutable inspectedDepositMinBalance;
```
thomas-waite: Why is there a concern about only being able to perform this swap once another PCV deposit is running low? Surely we want to swap all LUSD for DAI and so it doesn't matter what the balance of the Compound DAI PCV deposit is for example?

proposals/dao/tip_119.ts
```diff
@@ -35,34 +35,64 @@ let initialCompoundDAIBalance: BigNumber;
...
+  const lusdToDaiSwapperGuard = await curveSwapperGuardFactory.deploy(
     addresses.core,
-    gOhmEthOracle.address,
-    addresses.chainlinkEthUsdOracleWrapper,
-    false
+    addresses.pcvGuardian,
+    lusdToDaiCurveSwapper.address,
+    addresses.lusdHoldingPCVDeposit,
+    addresses.compoundDaiPCVDeposit,
+    ethers.utils.parseEther('10000000'), // activate if <10M DAI reserves
```
thomas-waite: Surely we want to swap all our LUSD for DAI? Why have this minimum DAI reserve condition?

thomas-waite (commented):
Looks good overall. 

Not sure why we need the minimum level of DAI reserves check before the swap is available. Also not a fan of the nested if/elses in the constructor to compute the token indexes - is there a tested Curve library that we can use to sort?

contracts/pcv/curve/CurveSwapper.sol
eswak: I'm not a fan either and honestly we could pass `i` and `j` as constructor args, they won't change

proposals/dao/tip_119.ts
eswak: exactly

contracts/pcv/curve/CurveSwapper.sol
eswak: I agree checking is useless, it's just copy pasta for consistency because the other Swapper contract do this check & we do it in various places to make sure we don't accidentally set state variables to the burner address (but these things should be catched with integration tests before deploy)

contracts/test/integration/pcv/CurveSwapper.t.sol
eswak: I'd expect we check that with a small integration test or proposal validation script before we deploy a new CurveSwapper ?

contracts/sentinel/guards/CurveSwapperGuard.sol
eswak: Can be discussed but I thought the spec was "holding LUSD is fine, but we want to be able to swap LUSD to DAI if DAI reserves are running low, to defend the peg". Can totally change that

proposals/dao/tip_119.ts
eswak: Can be discussed but I thought the spec was "holding LUSD is fine, but we want to be able to swap LUSD to DAI if DAI reserves are running low, to defend the peg". Can totally change that

contracts/pcv/curve/CurveSwapper.sol
eswak: passed i & j as constructor params

contracts/test/integration/pcv/CurveSwapper.t.sol
eswak: passed i & j as constructor params

contracts/sentinel/guards/CurveSwapperGuard.sol
eswak: see comment below

proposals/dao/tip_119.ts
eswak: we can deploy with 0 minimal reserve if the goal is to fully convert

contracts/pcv/curve/CurveSwapper.sol
thomas-waite: Nice

contracts/pcv/curve/CurveSwapper.sol
```diff
@@ -0,0 +1,144 @@
...
+    @param _j underlying index in the `_curvePool` of the `_tokenReceived`
+    @param _tokenSpent the token to be auctioned
+    @param _tokenReceived the token to buy
+    @param _tokenReceivingAddress the address to send `tokenReceived`
+    @param _maxSlippageBps the maximum tolerated slippage for swaps in basis points (10_000)
+     */
+    constructor(
+        address _core,
+        address _curvePool,
+        int128 _i,
```
thomas-waite, line 63: Really like passing these constructor params for `i` and `j` in. Much cleaner

contracts/test/integration/pcv/CurveSwapper.t.sol
```diff
@@ -0,0 +1,109 @@
...
+    address payable receiver = payable(address(3));
+
+    Core private core = Core(MainnetAddresses.CORE);
+
+    CurveSwapper private curveSwapper;
+
+    // dummy function to avoid errors
+    // the CurveSwapper will call deposit() on the
+    // tokenReceivingAddress after performing a swap
+    function deposit() public {}
```
thomas-waite, line 24: Feels strange to me to rely on this test file also being an 'active contract/mock' in your tests. 

Why not create a simple dummy mock contract at the top of this file with an empty `deposit()` function, deploy that and then set that address to be the `tokenReceivingAddress?`

thomas-waite (commented):
Looking good! One minor testing comment

PR #1071, opened 2022-09-18
4) Audit feedback, feiPSM: Silence `minAmountOut`, remove requires

thomas-waite:
# Audit feedback, feiPSM: Silence `minAmountOut`, remove requires
Mint/redemption is 1:1 and `minAmountOut` is only there for interface reasons. Can remove the require checks and silence the `minAmountOut` parameter

### DIFF
--- a/contracts/peg/SimpleFeiDaiPSM.sol
+++ b/contracts/peg/SimpleFeiDaiPSM.sol
@@ -33,7 +33,7 @@ contract SimpleFeiDaiPSM {
     function mint(
         address to,
         uint256 amountIn,
-        uint256 minAmountOut
+        uint256 // minAmountOut
     ) external returns (uint256 amountFeiOut) {
         amountFeiOut = amountIn;
         DAI.safeTransferFrom(msg.sender, address(this), amountIn);
@@ -47,7 +47,7 @@ contract SimpleFeiDaiPSM {
     function redeem(
         address to,
         uint256 amountFeiIn,
-        uint256 minAmountOut
+        uint256 // minAmountOut
     ) external returns (uint256 amountOut) {
         amountOut = amountFeiIn;
         FEI.safeTransferFrom(msg.sender, address(this), amountFeiIn);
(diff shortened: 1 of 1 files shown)


contracts/peg/SimpleFeiDaiPSM.sol
```diff
@@ -36,7 +36,6 @@ contract SimpleFeiDaiPSM {
         uint256 minAmountOut
```
Joeysantoro:
```suggestion
        uint256 // minAmountOut
```

Joeysantoro: I forget whether the syntax is to omit the uint256 or not when having unnamed variables. This should be done below too

eswak: syntax is to omit the uint256 variable name only, not the type (that's used by the compiler to know this function still expects an uint256, but just silences the "variable unused" warning)

eswak (dismissed):

contracts/peg/SimpleFeiDaiPSM.sol
thomas-waite: Omitted the variable name, kept the type

eswak (approved):

Joeysantoro (approved):

contracts/peg/SimpleFeiDaiPSM.sol
```diff
@@ -33,10 +33,9 @@ contract SimpleFeiDaiPSM {
     function mint(
         address to,
         uint256 amountIn,
-        uint256 minAmountOut
+        uint256 // minAmountOut
```
ElliotFriedman, line 36: If fei protocol is entering a terminal state where it will no longer be able to be upgraded, why even have a mint function?

thomas-waite: Fei will be a DAI wrapper. Needs to be mintable and redeemable to prevent any issues with the peg breaking

PR #1084, opened 2022-09-23
Add 1.5M FEI tolerance in tip121c validate step

eswak:
The following test was actually a bit flawed because it would work if more FEI got minted in the PSM, but would fail if more FEI got redeemed with the PSM :

```
expect(await contracts.dai.balanceOf(addresses.simpleFeiDaiPSM)).to.be.bignumber.greaterThan(
    USER_CIRCULATING_FEI_AT_FIXED_BLOCK
);
```

I replaced that check with : 
- Dynamic circulating FEI (`userCirculatingFeiSupply`) is within 500 FEI range from DAI in simpleFeiDaiPSM
- Dynamic circulating FEI is within 1.5M FEI from the hardcoded circulating FEI value (`USER_CIRCULATING_FEI_AT_FIXED_BLOCK`), which will revert the tests if that's the case and we should have a look at it


xklob (approved):

thomas-waite (approved):
Thanks for doing

PR #1085, opened 2022-09-23
1) Remove test 10M FEI mint 

thomas-waite:
# Remove test 10M FEI mint 
The sudo script called as part of the e2e setup mints 10M FEI to the accounts[0] account. Firstly, this should only be done in the relevant test where that is needed and not globally, but it also affects the PCV accounting in the tests as totalSupply() will be greater than it should be.

This PR removes that FEI minting


thomas-waite:
PCV stats with change:
image

xklob (approved):

PR #1090, opened 2022-09-27
Add missing functions to veBalHelper

miguelmtzinf:


contracts/metagov/VeBalHelper.sol
```diff
@@ -141,11 +166,28 @@ contract VeBalHelper is Ownable {
...
+    function withdrawERC20fromPCV(
         address token,
         address to,
         uint256 amount
     ) external onlyOwner {
         pcvDeposit.withdrawERC20(token, to, amount);
     }
+
+    function withdrawETHfromPCV(address to, uint256 amount) external onlyOwner {
+        pcvDeposit.withdraw(to, amount);
```
eswak: this needs to be withdrawETH

contracts/metagov/VeBalHelper.sol
```diff
@@ -141,11 +166,28 @@ contract VeBalHelper is Ownable {
...
+    function withdrawERC20fromStaker(
+        address token,
+        address to,
+        uint256 amount
+    ) external onlyOwner {
+        balancerStaker.withdrawERC20(token, to, amount);
+    }
+
+    function withdrawETHfromStaker(address to, uint256 amount) external onlyOwner {
+        balancerStaker.withdraw(to, amount);
```
eswak: this needs to be withdrawETH

contracts/metagov/VeBalHelper.sol
thomas-waite: Yep, needs to implement the PCVDeposit interface on our side

proposals/dao/tip_vebalotc.ts
```diff
@@ -32,23 +32,23 @@ const deploy: DeployUpgradeFunc = async (deployAddress: string, addresses: Named
...
+    addresses.balancerGaugeStakerProxy // _balancerGaugeStaker
   );
   await vebalOtcHelper.deployTransaction.wait();
   logging && console.log(`vebalOtcHelper: ${vebalOtcHelper.address}`);
 
-  // Deploy veBoostManager implementation
-  const VeBoostManagerFactory = await ethers.getContractFactory('VeBoostManager');
-  const veBoostManagerImplementation = await VeBoostManagerFactory.deploy(
+  // Deploy BalancerGaugeStakerV2 implementation
+  const balancerGaugeStakerV2Factory = await ethers.getContractFactory('BalancerGaugeStakerV2');
```
thomas-waite: Minor nit, for the contract factory we usually capitalise the first letter: `BalancerGaugeStakerV2Factory`

proposals/description/tip_vebalotc.ts
```diff
@@ -62,23 +62,30 @@ const tip_vebalotc: TemplatedProposalDescription = {
       target: 'proxyAdmin',
       values: '0',
       method: 'upgrade(address,address)',
-      arguments: (addresses) => [addresses.balancerGaugeStaker, addresses.veBoostManagerImplementation],
+      arguments: (addresses) => [addresses.balancerGaugeStakerProxy, addresses.balancerGaugeStakerV2Impl],
       description: `Upgrade implementation of the balancerGaugeStaker`
     },
     {
-      target: 'veBoostManager',
+      target: 'balancerGaugeStaker',
```
thomas-waite, line 69: I don't think this is correct. 

You need to be calling `balancerGaugeStakerV2Impl.initialize()` rather than reinitialise the original implementation (does it not revert as is? Initialisation should only be possible once)

`balanceGaugeStakerV2Impl` inherits from `BalanceGaugeStaker` and only defines new storage variables later, ensuring there are no storage collisions. You need to initialise the new implementation as there is a new storage variable

proposals/description/tip_vebalotc.ts
```diff
@@ -62,23 +62,30 @@ const tip_vebalotc: TemplatedProposalDescription = {
...
       values: '0',
       method: '_initialize(address,address)',
       arguments: (addresses) => [addresses.vebalOtcHelper, addresses.balancerVotingEscrowDelegation],
-      description: 'Initialize veBoostManagerProxy state variables'
+      description: 'Initialize BalancerGaugeStakerProxy state variables'
     },
+    // {
+    //   target: 'balancerGaugeStakerV2Impl',
+    //   values: '0',
```
thomas-waite, line 77: Why is this commented out?

thomas-waite (commented):
Need to discuss the proxy initialisation stuff + why those new functions have been added to veBalHelper

proposals/description/tip_vebalotc.ts
thomas-waite: The proxy will already have its originally defined V1 state variables initialised. We now need to initialise the new V2 state variable

contracts/metagov/VeBalHelper.sol
miguelmtzinf: Thanks, resolved!

contracts/metagov/VeBalHelper.sol
miguelmtzinf: ty!

proposals/description/tip_vebalotc.ts
miguelmtzinf: After the upgrade, the proxy has BalancerGaugeStakerV2 implementation. We need to call the new `_initialize` function

proposals/description/tip_vebalotc.ts
miguelmtzinf: The initialization of the implementation was failing for some reason (think its related to how the script work). Need to carefully review it again

proposals/description/tip_vebalotc.ts
thomas-waite: Yes you do, but you are calling the V1 initialise function here. You need to call the V2 initialise to initialise the new storage variable. The V1 storage variables are already defined in the proxy storage

proposals/description/tip_vebalotc.ts
thomas-waite: It is failing expectedly. You are calling the initialise() function of the V1 implementation contract which has already been initialised and should not be called again. 

You should be calling the initialise() function of the V2 implementation contract

contracts/metagov/VeBalHelper.sol
```diff
@@ -83,23 +88,43 @@ contract VeBalHelper is Ownable {
         pcvDeposit.voteForGaugeWeight(token, gaugeWeight);
     }
 
+    /// @notice Stake tokens in a gauge
+    function stakeInGauge(address token, uint256 amount) external onlyOwner {
```
thomas-waite, line 92: These new functions should not really be on this contract. I'm guessing you've added them here because you're thinking of creating a boost and then delegating to this contract? 

We can do that, but that's a bad design imo - you should be delegating the boost to an EOA/multisig and not this fixed contract imo. What happens if there's another method that you might need to call in the future that is not accounted for here?

thomas-waite:
In summary, I think the only changes should be the contract naming change from `VeBoostManager` to `balancerGaugeStakerV2Impl` and then adding a test for each `veBalHelper` contract function

proposals/description/tip_vebalotc.ts
miguelmtzinf: No. In order to call the V2 init function we need to call `_initialize()`.

contracts/metagov/VeBalHelper.sol
miguelmtzinf: Its just in the case we wanna stake BPTs from BalancerGaugeStaker contract. The boost would be delegated to that contract as well.

thomas-waite: Okay, fair enough. Will add tests for the new functions

proposals/description/tip_vebalotc.ts
thomas-waite: As discussed, ignore. Crossed wires/misunderstanding in naming. Good as is

proposals/description/tip_vebalotc.ts
thomas-waite: Ignore - failing due to another reason

PR #1101, opened 2023-09-07
Update GovernorAlpha.sol: Some Gas Optimizations

0xScratch:
Two types of changes made:

1st: changed `uint256 i = 0;` to `uint256 i;`
Default values of all unsigned integers is always 0. So, there's no need to waste some extra gas in order to initialize any 'uint' variable to 0

2nd: changed `i++` to `++i`
This change won't be messing with code's logic as far as I am concerned. And `++i` is more gas efficient than `++i`

For more details refer this -> https://gist.github.com/grGred/9bab8b9bad0cd42fc23d4e31e7347144#for-loops-improvement


xklob:
sus

0xScratch:
Closing due to inactivity!

